PR #3, opened 2019-08-26
Locations

tsibley:
Supporting code for (the start of) our location catalog. Depends on seattleflu/id3c#56.


tsibley:
Yep, should tag the sqitch schema. Good call!

tsibley:
I also changed the sqitch change from `views` to `warehouse/views` since `shipping/views` was used by @kairstenfay and is now on master. I was thinking we'd collect all custom views into one change, but splitting by schema is fine too. (If we have inter-schema view dependencies later, we can merge the files then.)

kairstenfay:
Oh, I didn't mean to start my own schema-dependent views organization. I mistakenly thought I was following what we did in `id3c`. Sorry about that.

tsibley:
Really no worries. :-) Either way is good right now.

PR #4, opened 2019-09-04
Update reportable notifications view

kairstenfay:
Update `shipping.reportable_notification_v1` to read the reporting log
from the new 'reporting_log' key in `presence_absence.details`.

Per https://github.com/seattleflu/id3c/pull/47 

Does this need a new sqitch tag?

### DIFF
--- a/deploy/shipping/views.sql
+++ b/deploy/shipping/views.sql
@@ -27,8 +27,7 @@ create or replace view shipping.reportable_condition_v1 as
         sample.identifier as sample,
         barcode,
         site.identifier as site,
-        presence_absence.details->>'reporting_status' as reporting_status,
-        presence_absence.details->>'timestamp' as reporting_timestamp
+        presence_absence.details->>'reporting_log' as reporting_log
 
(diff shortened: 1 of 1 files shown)


tsibley:
Because we haven't actually deployed the recent sqitch changes on master yet, we could actually get away with modifying the view definition in place and leaving everything else alone (i.e. not doing a `sqitch rework`).

kairstenfay:
> Because we haven't actually deployed the recent sqitch changes on master yet, we could actually get away with modifying the view definition in place and leaving everything else alone (i.e. not doing a `sqitch rework`).

Excellent point! I re-pushed with the simpler version of this commit.

tsibley:
:raised_hands:

tsibley:
Let's merge and deploy this? Deploy should be running this from your laptop:

```
sqitch status production
# check that the plan looks good
PGUSER=postgres sqitch deploy production
```

kairstenfay:
Merged to master and deployed to production

PR #5, opened 2019-09-06
Augur build data

joverlee521:
Shipping views for exporting metadata and consensus genome sequences to SFS augur-build.

SFS augur-build expect the metadata and genomic sequences to be separate, so it's easier for now to just supply them through separate views. 

Includes the creation of `augur-build-exporter` role.


joverlee521:
Will add a schema tag before merging into master

deploy/shipping/views.sql
```diff
@@ -43,4 +43,35 @@ create or replace view shipping.reportable_condition_v1 as
...
+            residence_census_tract as location,
+            'Seattle Flu Study' as authors,
+            case
+                when age < 18 then 'child'
+                else 'adult'
+            end as age_category,
+            case
+                when site_type in ('childrensHospital', 'childrensClinic', 'childrensHospital', 'clinic', 'hospital', 'retrospective') then 'clinical'
+                when site_type in ('childcare' , 'collegeCampus' , 'homelessShelter' , 'port', 'publicSpace', 'workplace') then 'community'
+            end as site_category,
```
kairstenfay: Just curious: what's the motivation for creating new categories of age and site here?

joverlee521: It's what Trevor had requested. I believe it's used as broad filters on the tree.

tsibley: It might be nice to eventually move these into `site.details` as `category` or something.

deploy/shipping/views.sql
```diff
@@ -43,4 +43,54 @@ create or replace view shipping.reportable_condition_v1 as
...
+
+			from shipping.incidence_model_observation_v2;
+
+comment on view shipping.metadata_for_augur_build_v1 is
+		'View of metadata necessary for SFS augur build';
+
+
+create or replace view shipping.genomic_sequences_for_augur_build_v1 as
+
+    select distinct on (sample.identifier, organism.lineage, segment)
```
tsibley: Nice use of `distinct on`. :-)

deploy/roles/augur-build-exporter/grants.sql
```diff
@@ -0,0 +1,15 @@
+-- Deploy seattleflu/id3c-customizations:roles/augur-build-exporter/grants to pg
+
+begin;
+
```
tsibley: I got the role/grants reworkable pattern a little wrong when I introduced it in id3c. I fixed it for the roles I touched in the `locations` branch, but the gist is that the deploy change should be revoking everything and then regranting only specific stuff. This then works well when the change is reworked later and the deploy becomes revert as it doesn't require updating.

So for example, before the first grant in this file you'd write:

```sql
revoke all on database :"DBNAME" from "augur-build-exporter";
revoke all on schema receiving, warehouse, shipping from "augur-build-exporter";
revoke all on all tables in schema receiving, warehouse, shipping from "augur-build-exporter";
```

tsibley:
More broadly, I wonder about in the future pushing these queries into the augur build itself instead of using views. That would let it determine the values of the static fields and special handling of things without requiring sqitch deploys or commits to two repos (id3c-customizations and augur-build).

joverlee521:
> More broadly, I wonder about in the future pushing these queries into the augur build itself instead of using views. That would let it determine the values of the static fields and special handling of things without requiring sqitch deploys or commits to two repos (id3c-customizations and augur-build).

Yeah, I can see it getting a bit annoying having to work with 2 repos to make changes. We can talk about this in person

PR #8, opened 2019-09-30
Link more targets to organisms

kairstenfay:
Update the target and organism tables according to the new Thermo Assay
names.

I have a few questions here:
~* How do I know if one of these new pathogens is reportable? ~
~* Should I add anything to the `identifiers` column of `warehouse.organism`? Having seen only the new Thermo Assay names data dictionary, I don't currently understand how that column is populated. I see the comment in the `organism.sql` file of ID3C that says :~
~    >    'Key-value pairs of external identifiers (i.e. dbxrefs) for this organism; keys should use well-known database or ontology prefix (e.g. NCBITaxon)';~

   ~But don't really know what a dbxref is or where to find it, even after Googling.~

~* Should I track controls (e.g. RnaseP Control)? I can add an extra column to the `with target_lineage (identifier, lineage) as (...` clause. Then should I update all targets to have a boolean `control` value?~
~* Why does only Bordetella (in `target_lineage`) use `::ltree`?~
~* Am I only adding the new Thermo Assay definition? I see that there are still quite a few targets in `warehouse.target` without linked a `organism_id` that exist as old Thermo Assay definitions.~

### DIFF
--- a/deploy/warehouse/target/data.sql
+++ b/deploy/warehouse/target/data.sql
@@ -5,7 +5,26 @@ begin;
 
 insert into warehouse.organism (lineage, identifiers, details)
     values
+        ('Adenovirus',              'NCBITAXON => 10508',       null),
         ('Bordetella_pertussis',    'NCBITAXON => 520',         '{"report_to_public_health": true}'),
+        ('Chlamydophila_pneumoniae','NCBITAXON => 83558',       null),
+        ('Enterovirus',             'NCBITAXON => 12059',       null),
+        ('Enterovirus.D',           'NCBITAXON => 138951',      null),
+        ('Enterovirus.D.68',        'NCBITAXON => 42789',       null),
+        ('Rhinovirus',              null,                       null),
+        ('Human_bocavirus',         'NCBITAXON => 329641',      null),
+        ('Human_coronavirus',       null,                       null),
+        ('Human_coronavirus.HKU1',  'NCBITAXON => 290028',      null),
+        ('Human_coronavirus.NL63',  'NCBITAXON => 277944',      null),
+        ('Human_coronavirus.229E',  'NCBITAXON => 11137',       null),
+        ('Human_coronavirus.OC43',  'NCBITAXON => 31631',       null),
+        ('Human_metapneumovirus',   'NCBITAXON => 162145',      null),
+        ('Human_parainfluenza',     null,                       null),
+        ('Human_parainfluenza.1',   'NCBITAXON => 12730',       null),
+        ('Human_parainfluenza.2',   'NCBITAXON => 1979160',     null),
+        ('Human_parainfluenza.3',   'NCBITAXON => 11216',       null),
+        ('Human_parainfluenza.4',   'NCBITAXON => 1979161',     null),
+        ('Human_parechovirus',      'NCBITAXON => 1803956',     null),
         ('Influenza',               null,                       null),
         ('Influenza.A',             'NCBITAXON => 11320',       null),
         ('Influenza.A.H1N1',        'NCBITAXON => 114727',      null),
@@ -16,9 +35,12 @@ insert into warehouse.organism (lineage, identifiers, details)
         ('Influenza.C',             'NCBITAXON => 11552',       null),
         ('Measles',                 'NCBITAXON => 11234',       '{"report_to_public_health": true}'),
         ('Mumps',                   'NCBITAXON => 1979165',     '{"report_to_public_health": true}'),
+        ('Mycoplasma_pneumoniae',   'NCBITAXON => 2104',        null),
         ('RSV',                     'NCBITAXON => 11250',       null),
         ('RSV.A',                   'NCBITAXON => 208893',      null),
-        ('RSV.B',                   'NCBITAXON => 208895',      null)
+        ('RSV.B',                   'NCBITAXON => 208895',      null),
+        ('Streptococcus_pneumoniae','NCBITAXON => 1313',        null)
+
 
     on conflict (lineage) do update
         set identifiers = coalesce(organism.identifiers, '') || EXCLUDED.identifiers,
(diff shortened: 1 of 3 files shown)


deploy/warehouse/target/data.sql
```diff
@@ -5,7 +5,17 @@ begin;
...
 insert into warehouse.organism (lineage, identifiers, details)
     values
+        ('Adenovirus.1',            null,                       null),
+        ('Adenovirus.2',            null,                       null),
         ('Bordetella_pertussis',    'NCBITAXON => 520',         '{"report_to_public_health": true}'),
+        ('Chlamydophila_pneumoniae',null,                       null),
+        ('Coronavirus.229E_OC43',   null,                       null),
+        ('Coronavirus.HKU1_NL63',   null,                       null),
+        ('Enterovirus',             null,                       null),
+        ('Enterovirus.D68',         null,                       null),
```
joverlee521: Should be `Enterovirus.D.68` since enterovirus D is the species and 68 is the subtype

deploy/warehouse/target/data.sql
```diff
@@ -16,9 +26,16 @@ insert into warehouse.organism (lineage, identifiers, details)
         ('Influenza.C',             'NCBITAXON => 11552',       null),
         ('Measles',                 'NCBITAXON => 11234',       '{"report_to_public_health": true}'),
         ('Mumps',                   'NCBITAXON => 1979165',     '{"report_to_public_health": true}'),
+        ('Mycoplasma_pneumoniae',   null,                       null),
+        ('Parainfluenza.1_2',       null,                       null),
+        ('Parainfluenza.3_4',       null,                       null),
```
joverlee521: Ditto different species for parainfluenza 1, 2, 3, and 4.

deploy/warehouse/target/data.sql
```diff
@@ -16,9 +26,16 @@ insert into warehouse.organism (lineage, identifiers, details)
...
+        ('Rhinovirus.1',            null,                       null),
+        ('Rhinovirus.2',            null,                       null),
```
joverlee521: Looks like Rhinovirus 1 and 2 are both pan Rhinovirus, so should only need one organism `Rhinovirus`

tsibley:
> How do I know if one of these new pathogens is reportable?

The existing pathogens/organisms were partially added so that I could mark the reportable ones as such. I don't think we have any new pathogens, just new target names for the same set. You could double-check in #lab.

> > Should I add anything to the identifiers column of warehouse.organism?
> 
> I think we are just sticking with `NCBITAXON` for now? These can be found at https://www.ncbi.nlm.nih.gov/taxonomy

Sticking with NCBITAXON would be good. The goal here is to anchor/document our organism definitions using the much much broader NCBI Taxonomy.

> But don't really know what a dbxref is or where to find it, even after Googling.

A dbxref is a commonly-used term in computational biology for a "database cross-reference". What that means is that one database (e.g. ID3C) refers to (cross-references) a record in another database. The broad convention for this is to use well-known short names for the referred-to database + that database's record identifier. Did you find [this page](https://www.ncbi.nlm.nih.gov/genbank/collab/db_xref/) describing GenBank's use of dbxrefs?

> > Should I track controls (e.g. RnaseP Control)?
> 
> I don't think so? They're not specific to any organism so I don't think we need to add them

Correct.

> > Am I only adding the new Thermo Assay definition? I see that there are still quite a few targets in warehouse.target without linked a organism_id that exist as old Thermo Assay definitions.
> 
> I think the goal is to have all non-control targets linked to an organism so that downstream analysis can be based on organism rather than target.

Yep, or most non-control targets. This means we should link both the old and new assay/target names to an organism, e.g. both `M_pneumoninae` and `M. pneumoninae_AI5IRK5` to `Mycoplasma_pneumoniae`.

> Why does only Bordetella (in `target_lineage`) use `::ltree`?

Since the string literals in that `values` statement happen outside of a typed row/table context, they will default to textual types, which won't work in the subsequent comparison to the lineage column. By typing one of the string literals in the `target_lineage.lineage` column, the SQL engine can infer the type of the whole column. This is less noisy than explicitly typing every value, and I used the first since it also clues in the reader early.

deploy/warehouse/target/data.sql
```diff
@@ -5,7 +5,17 @@ begin;
 
 insert into warehouse.organism (lineage, identifiers, details)
     values
+        ('Adenovirus.1',            null,                       null),
+        ('Adenovirus.2',            null,                       null),
```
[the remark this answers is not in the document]
tsibley: I concur, but that's pretty confusing given that "Human adenovirus 1" and "2" are a thing!

deploy/warehouse/target/data.sql
```diff
@@ -5,7 +5,17 @@ begin;
...
         ('Bordetella_pertussis',    'NCBITAXON => 520',         '{"report_to_public_health": true}'),
+        ('Chlamydophila_pneumoniae',null,                       null),
+        ('Coronavirus.229E_OC43',   null,                       null),
```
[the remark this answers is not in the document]
tsibley: Ugh, yeah. Looks like we'll need a `target_organism` link/join table. Since it'll require more work, for now I think we can punt on the cases like this where a target is linked to multiple specific species/subtypes. We can address it later if necessary.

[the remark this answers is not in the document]
tsibley: Ugh, yeah. Looks like we'll need a `target_organism` link/join table. Since it'll require more work, for now I think we can punt on the cases like this where a target is linked to multiple specific species/subtypes. We can address it later if necessary.

kairstenfay:
[Link more targets to organisms](https://trello.com/c/TJjOHnR6/214-link-more-targets-to-organisms)

deploy/warehouse/target/data.sql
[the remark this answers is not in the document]
kairstenfay: Do you mean that I should ignore Coronavirus for now?

deploy/warehouse/target/data.sql
kairstenfay: So, skip these along with Coronavirus for now?

deploy/warehouse/target/data.sql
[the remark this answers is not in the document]
tsibley: Yep. If you would create a card in Trello so we remember this needs following up, that'd be great.

deploy/warehouse/target/data.sql
tsibley: Yep.

kairstenfay:
I've incorporated the requested changes. 
I added old thermo assay names, and now every non-`BLANK`, non-doubly linked organism (e.g. `CoV_HKU1_CoV_NL63`), non-control target in the database now has an associated organism_id.

deploy/warehouse/target/data.sql
```diff
@@ -48,4 +112,15 @@ insert into warehouse.target (identifier, organism_id, control)
...
 
+
+insert into warehouse.target (identifier, control)
+    values
+        ('Hs04930436_g1',   't'),
+        ('Ac00010014_a1',   't')
+
+    on conflict (identifier) do update
+        set identifier  = coalesce(target.identifier, '')   || EXCLUDED.identifier,
+            control     = EXCLUDED.control
```
kairstenfay: does this need a `coalesce()`? My guess is no since `f` is the default value.

deploy/warehouse/target/data.sql
```diff
@@ -5,8 +5,19 @@ begin;
...
+        ('Coronavirus.HKU1',        null,                       null),
+        ('Coronavirus.NL63',        'NCBITAXON => 277944',      null),
+        ('Coronavirus.229E',        'NCBITAXON => 11137',       null),
+        ('Coronavirus.OC43',        null,                       null),
+        ('Enterovirus',             'NCBITAXON => 12059',       null),
+        ('Enterovirus.D.68',        'NCBITAXON => 42789',       null),
+        ('Human_bocavirus',         'NCBITAXON => 329641',      null),
+        ('Human_metapneumovirus',   'NCBITAXON => 162145',      null),
+        ('Human_parechovirus',      null,                       null),
+        ('Influenza',               'NCBITAXON => 11309',       null),
```
joverlee521: I'm not sure this is the correct NCBI Taxon ID, [11309](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=11309&lvl=3&lin=f&keep=1&srchmode=1&unlock) might be for a specific unidentified _strain_ of Influenza. I don't think `Influenza` exists as a NCBI Taxon entry.

kairstenfay:
>     * other than `Influenza` and `Rhinovirus`, I believe everything else should have a NCBI Taxon ID

I see [human coronavirus hku1](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=290028&lvl=3&lin=f&keep=1&srchmode=1&unlock). Should I set this as the NCBI Taxon ID for `Coronavirus.HKU1`?

>     * `Rhinovirus` is technically under the genus `Enterovirus`, should the lineage be `Enterovirus.Rhinovirus` or are we not following the entire taxonomy?

No idea here

>     * should we keep `Human` in the organism lineages? I'm assuming all pathogens we are dealing with are human

Also no idea here and touches on why I'm hesitant to say Coronavirus HKU1 == Human coronavirus HKU1. 

>     * We should probably have an organism entry for every label in a lineage.

Do you mean for example `Coronavirus` should have its own stand-alone entry? If not, please provide an example here.

kairstenfay:
@joverlee521 
The only Adenivorus is `unidentified adenovirus`. Should I skip that one, too?

joverlee521:
> The only Adenovirus is `unidentified adenovirus`. Same for Human coronavirus. Should I skip these?

Not sure if this should count for [adenovirus](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=10508&lvl=3&lin=f&keep=1&srchmode=1&unlock) and if this should count for [coronavirus](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=11118&lvl=3&lin=f&keep=1&srchmode=1&unlock), but we can also just skip them for now and add them later.

kairstenfay:
> > The only Adenovirus is `unidentified adenovirus`. Same for Human coronavirus. Should I skip these?
> 
> Not sure if this should count for [adenovirus](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=10508&lvl=3&lin=f&keep=1&srchmode=1&unlock) and if this should count for [coronavirus](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=11118&lvl=3&lin=f&keep=1&srchmode=1&unlock), but we can also just skip them for now and add them later.

I'll add the one for adenovirus but hold on Coronavirus (since we may want to specifically look for the taxon ID for human coronavirus). If that doesn't make sense, let me know and I'll use the taxon ID for Coronavirus you found.

deploy/warehouse/target/data.sql
tsibley: There is no need for a coalesce here. We want to set control to the value we hardcode here regardless of what its previous value was.

deploy/warehouse/target/data.sql
```diff
@@ -48,4 +120,15 @@ insert into warehouse.target (identifier, organism_id, control)
...
 ;
 
+
+insert into warehouse.target (identifier, control)
+    values
+        ('Hs04930436_g1',   't'),
+        ('Ac00010014_a1',   't')
+
+    on conflict (identifier) do update
+        set identifier  = coalesce(target.identifier, '')   || EXCLUDED.identifier,
```
tsibley: This line doesn't make much sense. The coalesce() is unnecessary because target.identifier is required to be not null by a column constraint and is also the conflicting value named in the `on conflict (identifier)` clause. Additionally, `||` is the concatenation operator (not a logical or), so this line would update the identifier to be doubled. The setting of `identifier` can be entirely removed from this upsert.

deploy/warehouse/target/data.sql
```diff
@@ -5,7 +5,26 @@ begin;
 
 insert into warehouse.organism (lineage, identifiers, details)
     values
+        ('Adenovirus',              'NCBITAXON => 10508',       null),
         ('Bordetella_pertussis',    'NCBITAXON => 520',         '{"report_to_public_health": true}'),
+        ('Chlamydophila_pneumoniae','NCBITAXON => 83558',       null),
+        ('Enterovirus',             'NCBITAXON => 12059',       null),
+        ('Enterovirus.D',           'NCBITAXON => 138951',      null),
+        ('Enterovirus.D.68',        'NCBITAXON => 42789',       null),
+        ('Enterovirus.Rhinovirus',  null,                       null),
```
tsibley: This strikes me as a little weird, but indeed it's how NCBI Taxonomy structures it and how Wikipedia notes it is defined cladistically. I think the question is if we expect to include rhinoviruses when querying for all `Enterovirus`. @trvrb may have thoughts/opinions here.

deploy/warehouse/target/data.sql
```diff
@@ -27,16 +49,66 @@ insert into warehouse.organism (lineage, identifiers, details)
...
+        ('AI1RW2H',                     'Chlamydophila_pneumoniae'),
+        ('C_pneumoniae',                'Chlamydophila_pneumoniae'),
+        ('C. pneumoniae',               'Chlamydophila_pneumoniae'),
+        ('C. pneumoniae_AI1RW2H',       'Chlamydophila_pneumoniae'),
+        ('C.pneumoniae',                'Chlamydophila_pneumoniae'),
+        ('AP7DPVF',                     'Enterovirus'),
+        ('EnterovirusA_B 1_AP7DPVF',    'Enterovirus'),
+        ('Enterovirus_pan',             'Enterovirus'),
+        ('Ev_pan',                      'Enterovirus'),
+        ('EV_pan',                      'Enterovirus'),
```
tsibley: Based on the target name `EnterovirusA_B 1_AP7DPVF` and the target equivalency spreadsheet on the Trello card, I think these all actually target only Enterovirus A and B. This gives them the same problem as the Coronavirus combos.

tsibley:
@kairstenfay @joverlee521 Thanks both for dealing with this. This is more a can of worms than I expected!

deploy/warehouse/target/data.sql
kairstenfay: Agreed. Jover and I discussed this and were unsure whether we are more concerned with representing true taxonomies or with reporting organisms colloquially.

deploy/warehouse/target/data.sql
tsibley: I think this applies not just to `EnterovirusA_B 1_AP7DPVF` but also the three variations on the pan Enterovirus targets. The old target name was `Enterovirus_pan` which is now `EnterovirusA_B 1_AP7DPVF`, but I don't think the actual target changed, just the name...

tsibley: I've asked Peter for confirmation: https://seattle-flu-study.slack.com/archives/CCAEWSFTK/p1570490686074300

kairstenfay: Oh, I see now. Thanks for checking with Peter.

deploy/warehouse/target/data.sql
tsibley: I've pinged Slack about this: https://seattle-flu-study.slack.com/archives/CCAA9RBFS/p1570550480065400

tsibley: Let's call this just `Rhinovirus` for now.

kairstenfay: It is now just `Rhinovirus`.

deploy/warehouse/target/data.sql
tsibley: Ah, looks like the A and B assumption was wrong:

> It does not just pick up the A and B subtypes.  That assay ID AP7DPVF includes all enterovirus species A, B, C, D, G

So we can leave this as-is!

tsibley:
Merged and deployed.

PR #9, opened 2019-10-14
Hook up custom command entrypoint for newly-extensible ID3C

tsibley:
The other half of https://github.com/seattleflu/id3c/pull/72.


lib/seattleflu/id3c/__version__.py
```diff
@@ -0,0 +1,7 @@
+"""
+Defines the version of the Seattle Flu ID3C customizations as ``__version__``.
+
+The version scheme is the same as ID3C for the same reasons.  See
+``lib/id3c/__version__.py``.
+"""
+__version__ = '2019.1'
```
kairstenfay: Does this mean every time we update the version of id3c we also have to update the id3c-customizations version (and vice-versa)?

setup.py
```diff
@@ -0,0 +1,65 @@
...
+    entry_points = {
+        "id3c.cli.commands": [
+            "seattleflu = seattleflu.id3c.cli.command"
+        ],
+    },
+
+    python_requires = ">=3.6",
+
+    install_requires = [
+        "id3c >=2019.1",
```
kairstenfay: Will this need to be manually updated as well?

joverlee521:
Currently all extensions are loaded as `id3c.cli.commands`, does this mean we would have to create another `entry_point` for `id3c.cli.commands.etl` to register extended `etl` subcommands or is there another way of doing this?

tsibley:
> Currently all extensions are loaded as `id3c.cli.commands`, does this mean we would have to create another `entry_point` for `id3c.cli.commands.etl` to register extended `etl` subcommands or is there another way of doing this?

Talked about this in person, but for posterity, the short answer is no. Extensions can register subcommands of `etl` (sub-subcommands) by importing that `click.Group` object directly and decorating the new command with `@etl.command`.

lib/seattleflu/id3c/__version__.py
tsibley: Good question! The answer is no. These two projects are setup to use the same versioning scheme, but the versions are independent. We'll bump 2019.1 to 2019.2 when make a new release/deploy. This will be part of the release/deploy automation yet to come.

setup.py
tsibley: This only needs to be updated if our customizations depend on specific code we add in a later release of ID3C, the same way we would for other deps, e.g. pandas, if there was a bugfix/feature we needed in a newer version.

PR #10, opened 2019-10-17
WIP: Create new routine, etl redcap-det swab-n-send

kairstenfay:
Here is the "bare bones" ETL for REDCap Swab n Send records. 

The outstanding TODOs are related to:
* moving all the REDCap ETLs to a shared file with subcommands (see the stub)
* geocoding
* deidentification (personal and household) 
* leftovers from the draft-mode "Single Swab and Send" that need to be updated with the new "Swab and Send" survey (that currently has no data points) 

I wanted to get some eyes on this to see if we have a reasonable approach in parsing REDCap records to FHIR format. 

See seattleflu/id3c/pull/73 for minor helper function changes that accompany this PR.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
@@ -515,10 +515,10 @@ def create_resource_questionnaire_response(record: dict, patient: dict,
         "item": items,
     }
 
-    if not items:
-        del questionnaire_response['item'] # TODO I get a typing error if I add items after declaring questionnaire_response
+    if items:
+        return questionnaire_response
 
-    return questionnaire_response
+    return None
 
 
 def questionnaire_item(record: dict, question_id: str, response_type: str) -> Optional[dict]:
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
@@ -96,18 +96,20 @@ def redcap_det_swab_n_send(*, log_output: bool, db: DatabaseSession):
             specimen        = create_resource_specimen(redcap_record, patient)
             immunization    = create_resource_immunization(redcap_record, patient)
 
+            other_resources = [
+                patient,
+                encounter,
+                questionnaire,
+                specimen,
+                immunization,
+            ]
+
             bundle = {
                 "resourceType": "Bundle",
                 "id": str(uuid4()),
                 "type": "collection",
                 "timestamp": datetime.now().astimezone().isoformat(),
-                "entry": [
-                    resource(patient),
-                    resource(encounter),
-                    resource(questionnaire),
-                    resource(specimen),
-                    resource(immunization)
-                ]
+                "entry": [ resource(r) for r in other_resources if r is not None ]
             }
 
             for location in location_resources:
@@ -245,11 +247,15 @@ def generate_patient_hash(redcap_record: dict) -> dict:
     return generate_hash(str(sorted(personal_information.items())))
 
 
-def create_resource_immunization(redcap_record: dict, patient: dict) -> dict:
+def create_resource_immunization(redcap_record: dict, patient: dict) -> Optional[dict]:
     """ Returns a FHIR Immunization resource. """
+    vaccine_status = vaccine(redcap_record)
+    if not vaccine_status:
+        return None
+
     immunization = {
         "resourceType": "Immunization",
-        "status": vaccine(redcap_record),
+        "status": vaccine_status,
         "vaccineCode": {
             "coding": [
                 {
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
@@ -1,0 +1,668 @@
+"""
+Process REDCap DET documents into the relational warehouse.
+
+Contains some hard-coded logic for which project ID correlates to which project
+(e.g. 17421 is the PID for Shelters)
+"""
+import os
+import re
+import click
+import json
+import hashlib
+import logging
+import requests
+from uuid import uuid4
+from typing import Any, Collection, Dict, List, Match, Optional, Union
+from datetime import datetime, timezone
+from id3c.db.session import DatabaseSession
+from id3c.db.datatypes import Json
+from id3c.cli.command import with_database_session
+from id3c.cli.command.etl import etl, race
+
+from id3c.cli.command.etl.redcap_det import (
+    is_complete,
+    redcap_det,
+    insert_fhir_bundle,
+    mark_skipped,
+    mark_processed,
+    get_redcap_record
+)
+
+
+LOG = logging.getLogger(__name__)
+
+
+REVISION = 1
+ETL_NAME = 'redcap-det swab-n-send'
+etl_id = { 'revision': REVISION, "etl_name": ETL_NAME }
+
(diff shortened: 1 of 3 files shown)


kairstenfay:
[Custom ETL for swab and send records from REDCap](https://trello.com/c/oMrD7z7Y/239-custom-etl-for-swab-and-send-records-from-redcap)

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
```diff
@@ -0,0 +1,575 @@
...
+import requests
+from typing import Any, List, Dict, Optional, Match, Collection
+from datetime import datetime, timezone
+from id3c.db.session import DatabaseSession
+from id3c.db.datatypes import Json
+from id3c.cli.command.etl import etl, race
+
+from id3c.cli.command.etl.redcap_det import (
+    COMPLETE_AND_VERIFIED,
+    LOG,
```
joverlee521: I don't think we want to share the `LOG`. We will probably want to differentiate the log output between the different redcap-det etl processes so we can debug things more easily if something goes wrong.

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
```diff
@@ -0,0 +1,575 @@
...
+    HEADERS,
+    REVISION,
```
joverlee521: Do we want to share the `REVISION` number for all of the projects?

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
```diff
@@ -0,0 +1,575 @@
...
+            with db.savepoint(f"redcap_det {det.id}"):
+                LOG.info(f"Processing REDCap DET {det.id}")
+
+                redcap_record = get_redcap_record(PROJECT_ID, det.document['record'], TOKEN_NAME)
+
+                for instrument in INSTRUMENTS_REQUIRED_TO_BE_COMPLETE:
+                    if redcap_record[f"{instrument}_complete"] != COMPLETE_AND_VERIFIED:
+                        LOG.debug(f"Skipping incomplete or unverified REDCap DET {det.id}")
+                        mark_skipped(db, det.id)
+                        continue
```
joverlee521: This `continue` is within a nested for loop, so the process doesn't move on to the next `det`

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
```diff
@@ -0,0 +1,575 @@
...
+    mark_skipped,
+    mark_processed,
+    get_redcap_record
+)
+
+REDCAP_URL = 'redcap.iths.org'
+INTERNAL_SYSTEM = "https://seattleflu.org"
+UW_CENSUS_TRACT = '53033005302'
+
+@etl.command("redcap-det-swab-n-send", help = __doc__)
```
joverlee521: Since we have the `redcap-det` group setup in https://github.com/seattleflu/id3c/pull/73, this can be changed to `@redcap-det.command('swab-n-send')`

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
kairstenfay: Oh, right. I had originally written that temporary import thinking the contents of this file would go in the same file as `redcap-det`. Of course this is no longer the case, so I have to change that.

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
kairstenfay: Same thoughts as above comment.

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
kairstenfay: Yep, that was the plan for it and no reason I can't make it a subcommand now.

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
kairstenfay: Actually, I don't know the answer here. `mark_processed` requires `REVISION` so this is a trickier question.

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
kairstenfay: Good catch

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
joverlee521: Same! An option would be to pass the `REVISION` number to `mark_processed`. Just not sure if we want to do that or just have the same number for all redcap-det etls

kairstenfay:
This PR now inserts created documents into receiving and incorporates the `action` arguments (dry-run, etc.)

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
tsibley: I think each ETL should have it's own revision, else when only one of them changes, all of them will re-run on the entire DET log.

kairstenfay: Should we pass `REVISION` as an argument to `mark_processed` or create a new function within each ETL process?

kairstenfay:
Heads up, y'all: I'm going to be making a lot of `fixup!` commits now that I have found a way to validate FHIR documents. I didn't get it 100% right on the first try.

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
joverlee521: If we are using separate REVISION numbers then we also need to include ETL_NAME in the processing log

tsibley: > Should we pass `REVISION` as an argument to `mark_processed` or create a new function within each ETL process?

I don't think I have a strong preference.

> If we are using separate REVISION numbers then we also need to include ETL_NAME in the processing log

Yep!

tsibley:
Sounds good! Let us know when you've made all your fixups and rebased them into the commit.

kairstenfay:
Fixed up and rebased! Ready for re-review. There are still a lot of TODOs including some where I could use some feedback.

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
```diff
@@ -0,0 +1,668 @@
...
+    ]
+
+    question_categories = {
+        'valueCoding': coding_questions,
+        'valueBoolean': boolean_questions,
+        'valueInteger': integer_questions,
+        'valueString': string_questions,
+    }
+
+    redcap_record['race'] = create_custom_race_key(redcap_record)
```
kairstenfay: on second thought I'll probably rename the function invoked here to `create_custom_race_value`

tsibley:
I'm rebasing the work I've done this week so far onto the repushed branches from you and Jover. My plan is to squash these commits together since they're all fixups of the first. Also, I'm going to drop the changes to `Pipfile.lock` which look spurious (deps aren't used in the code and aren't declared in setup.py or Pipfile).

tsibley:
Closing in favor of #15

PR #16, opened 2019-10-31
Return results view

joverlee521:
Creates shipping view for return of results on the SFS website. 

This view includes `barcode`, `status`, `organisms_present`, and `organisms_absent`. 
This does not include sequencing status/info yet.

### DIFF
--- a/schema/sqitch.plan
+++ b/schema/sqitch.plan
@@ -24,4 +24,7 @@ warehouse/target/data [warehouse/target/data@2019-09-27b] 2019-09-30T22:08:52Z K
 @2019-09-30 2019-09-30T23:59:55Z Kairsten Fay <kfay@fredhutch.org> # Link more targets to organisms
 shipping/views [shipping/views@2019-09-30] 2019-10-18T16:23:26Z Jover Lee <joverlee@fredhutch.org> # Update reportable-conditions to include sample_origin and swab_site
 @2019-10-18 2019-10-18T16:32:08Z Jover Lee <joverlee@fredhutch.org> # Changes for 18 October 2019
+
 shipping/views [shipping/views@2019-10-18] 2019-10-29T23:58:26Z Jover Lee <joverlee@fredhutch.org> # Add return of results view
+roles/return-results-exporter/create [shipping/views] 2019-11-06T18:12:21Z Jover Lee <joverlee@fredhutch.org> # Role for exporting return results
+roles/return-results-exporter/grants [shipping/views] 2019-11-06T18:14:00Z Jover Lee <joverlee@fredhutch.org> # Grants to return-results-exporter
--- a/schema/sqitch.plan
+++ b/schema/sqitch.plan
@@ -24,3 +24,4 @@ warehouse/target/data [warehouse/target/data@2019-09-27b] 2019-09-30T22:08:52Z K
 @2019-09-30 2019-09-30T23:59:55Z Kairsten Fay <kfay@fredhutch.org> # Link more targets to organisms
 shipping/views [shipping/views@2019-09-30] 2019-10-18T16:23:26Z Jover Lee <joverlee@fredhutch.org> # Update reportable-conditions to include sample_origin and swab_site
 @2019-10-18 2019-10-18T16:32:08Z Jover Lee <joverlee@fredhutch.org> # Changes for 18 October 2019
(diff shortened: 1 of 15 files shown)


joverlee521:
[Connect return of results to ID3C](https://trello.com/c/m9uMWQYM/220-connect-return-of-results-to-id3c)

schema/sqitch.plan
```diff
@@ -24,3 +24,8 @@ warehouse/target/data [warehouse/target/data@2019-09-27b] 2019-09-30T22:08:52Z K
 @2019-09-30 2019-09-30T23:59:55Z Kairsten Fay <kfay@fredhutch.org> # Link more targets to organisms
 shipping/views [shipping/views@2019-09-30] 2019-10-18T16:23:26Z Jover Lee <joverlee@fredhutch.org> # Update reportable-conditions to include sample_origin and swab_site
 @2019-10-18 2019-10-18T16:32:08Z Jover Lee <joverlee@fredhutch.org> # Changes for 18 October 2019
+
+shipping/views [shipping/views@2019-10-18] 2019-10-29T23:58:26Z Jover Lee <joverlee@fredhutch.org> # Add return of results view
+roles/return-results-exporter/create 2019-10-30T19:13:57Z Jover Lee <joverlee@fredhutch.org> # Role for exporting return results
+roles/return-results-exporter/grants 2019-10-30T19:16:44Z Jover Lee <joverlee@fredhutch.org> # Grants to return-results-exporter
```
tsibley: I think both of these `roles/…` changes should declare their dependency on the latest `shipping/views`. It is a technicality, but good practice.

schema/deploy/shipping/views.sql
```diff
@@ -131,4 +131,40 @@ create or replace view shipping.flu_assembly_jobs_v1 as
...
     'View of flu jobs that still need to be run through the assembly pipeline';
 
+
+create or replace view shipping.return_results_v1 as
+
+    select barcode,
+           case
+             when sample_id is null then 'notReceived'
+             when sample_id is not null and string_agg(present::text, ',') is null then 'processing'
+             when string_agg(present::text, ',') is not null then 'complete'
```
tsibley: `string_agg(present::text, ',') is null` is a little weird to me. What about `count(present) = 0` and `count(present) > 0`?

schema/deploy/shipping/views.sql
```diff
@@ -131,4 +131,40 @@ create or replace view shipping.flu_assembly_jobs_v1 as
...
+             distinct
+             case
+                 -- We only return the top level organisms for results so we want to omit subtypes
+                 when present then subpath(organism, 0, 1)::text
+             end), null)  as organisms_present,
+           array_remove(array_agg(
+             distinct
+             case
+                 when not present then subpath(organism, 0, 1)::text
+             end), null) as organisms_absent
```
tsibley: Something like the below reads a bit more clearly to me by making use of aggregate `filter` clause to accomplish the same thing as the `case` and `array_remove`.

```sql
array_agg(distinct subpath(organism, 0, 1)::text)
    filter (where present and organism is not null)
    as organisms_present,

array_agg(distinct subpath(organism, 0, 1)::text)
    filter (where not present and organism is not null)
    as organisms_absent
```

schema/deploy/shipping/views.sql
```diff
@@ -131,4 +131,40 @@ create or replace view shipping.flu_assembly_jobs_v1 as
...
+
+      from warehouse.identifier
+      left join warehouse.sample on uuid::text = sample.collection_identifier
+      left join shipping.presence_absence_result_v1 on sample.identifier = presence_absence_result_v1.sample
+
+    --These are all past and current collection identifier sets not including self-test
+    where identifier_set_id in (4,7,9,10,11,12,13) and
```
tsibley: It would be clearer to name these instead of using their opaque numbers. :-)

joverlee521:
This PR now includes a custom API endpoint for return of results. It depends on PR https://github.com/seattleflu/id3c/pull/85

tsibley:
Merged! Would you do the deploy?

tsibley:
Ah, I merged slightly too soon. The `origins=["http://localhost:8080"]` value should also include `"https://seattleflu.org"` and `"https://dev.seattleflu.org"`.

joverlee521:
> Ah, I merged slightly too soon. The `origins=["http://localhost:8080"]` value should also include `"https://seattleflu.org"` and `"https://dev.seattleflu.org"`.

I'll add those and push to master, then deploy

PR #17, opened 2019-11-19
Update views for data privacy

joverlee521:
This PR is dependent on https://github.com/seattleflu/id3c/pull/87
Update downstream views to use `shipping.incidence_model_observation_v3`

The need to rework the grants for `augur-build-exporter` made it necessary to do 2 reworks of `shipping/views`. `shipping.metadata_for_augur_build_v1` needs to be dropped after the rework of the role grants so that the grants could be reverted without error. 

Also moved augur-export routes to be a custom API endpoint.


joverlee521:
[Remove unnecessary variables from export for modeling team](https://trello.com/c/p4Yzflso/263-remove-unnecessary-variables-from-export-for-modeling-team)

lib/seattleflu/id3c/api/routes.py
```diff
@@ -17,3 +17,32 @@ def get_barcode_results(barcode, session):
     LOG.debug(f"Exporting presence/absence results for <{barcode}>")
     results = datastore.fetch_barcode_results(session, barcode)
     return jsonify(results)
+
+
+@api_v1.route("/shipping/augur-build-metadata", methods = ['GET'])
```
tsibley: Seems like this should be a `/v2/…` route since the format changed? Thoughts? We're the only consumer, so we can break the rules if it's easier/better.

schema/deploy/shipping/views.sql
```diff
@@ -79,6 +79,33 @@ comment on view shipping.metadata_for_augur_build_v1 is
...
+create or replace view shipping.metadata_for_augur_build_v2 as
+
+    select  sample as strain,
+            encountered as date,
+            'seattle' as region,
+            -- XXX TODO: Change to PUMA and neighborhoods
+            residence_census_tract as location,
+            'Seattle Flu Study' as authors,
+            case
+                when age_range_coarse <@ '[0 mon, 18 years)'::intervalrange then 'child'
```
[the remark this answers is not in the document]
joverlee521: Oh! That's good to know!

lib/seattleflu/id3c/api/routes.py
joverlee521: Oh, we did talk about this at one point. We can definitely change to `/v2/...`, the only thing is we would have to change the url in [augur-build](https://github.com/seattleflu/augur-build/blob/master/scripts/download_sfs_metadata.py)

kairstenfay: I agree that we probably want a `v2`

tsibley:
Looks ready to merge. Thanks! I pushed a commit adding a handler for the old v1 endpoint that responds 410 Gone instead of falling back to 404 Not Found. If that looks good to you, would you merge and deploy this and seattleflu/id3c#87?

PR #18, opened 2019-11-23
Sch year 1 clinical data

kairstenfay:
This relies on changes in seattleflu/id3c/pull/88.

Move the clinical, etl clinical, longitudinal, and etl longitudinal commands (and related helper functions) from ID3C to ID3C-customizations.

Make necessary changes to `clinical` and `etl clinical` to ingest new SCH Year 1 data.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/__init__.py
+++ b/lib/seattleflu/id3c/cli/command/__init__.py
@@ -11,8 +11,10 @@ import logging
 import pandas as pd
 from typing import List
 
+
 # Load all ETL subcommands.
 __all__ = [
+    "etl",
     "clinical",
     "longitudinal",
     "reportable_conditions"
--- a/lib/seattleflu/id3c/cli/command/__init__.py
+++ b/lib/seattleflu/id3c/cli/command/__init__.py
@@ -7,6 +7,117 @@ causes the core id3c.cli module to load this file at CLI runtime.
 By in turn loading our own individual commands here, we allow each command
 module to register itself via Click's decorators.
 """
-from . import (
-    reportable_conditions
-)
+import logging
+import pandas as pd
+from typing import List
+
+# Load all ETL subcommands.
+__all__ = [
+    "clinical",
+    "longitudinal",
+    "reportable_conditions"
+]
+
+
+LOG = logging.getLogger(__name__)
+
+
+def add_metadata(df: pd.DataFrame, filename: str) -> pd.DataFrame:
+    """ Adds a metadata column to a given DataFrame *df* for reporting """
+    df['_metadata'] = list(map(lambda index: {
+        'filename': filename,
+        'row': index + 2}, df.index))
+    return df
+
+
+def trim_whitespace(df: pd.DataFrame) -> pd.DataFrame:
+    """ Trims leading and trailing whitespace from strings in *df* """
+    str_columns = df.columns[every_value_is_str_or_na(df)]
+
+    # Guard against AttributeErrors from entirely empty non-object dtype columns
+    str_columns = list(df[str_columns].select_dtypes(include='object'))
+
+    df[str_columns] = df[str_columns].apply(lambda column: column.str.strip())
+
+    return df
+
+
+def every_value_is_str_or_na(df):
+    """
+    Evaluates whether every value in the columns of a given DataFrame *df* is
+    either a string or NA.
+    """
+    return df.applymap(lambda col: isinstance(col, str) or pd.isna(col)).all()
+
+
+def barcode_quality_control(clinical_records: pd.DataFrame, output: str) -> None:
+    """ Perform quality control on barcodes """
+    missing_barcodes = missing_barcode(clinical_records)
+    duplicated_barcodes = duplicated_barcode(clinical_records)
+
+    print_problem_barcodes(pd.concat([missing_barcodes, duplicated_barcodes],
+                                 ignore_index=True), output)
+
+    assert len(duplicated_barcodes) == 0, "You have duplicated barcodes!"
+
+
+def missing_barcode(df: pd.DataFrame) -> pd.DataFrame:
+    """
+    Given a pandas DataFrame *df*, returns a DataFrame with missing barcodes and
+    a description of the problem.
+    """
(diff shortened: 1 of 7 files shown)


lib/seattleflu/id3c/cli/command/__init__.py
```diff
@@ -7,6 +7,113 @@
...
+import pandas as pd
+from typing import List
+
+# Load all ETL subcommands.
+__all__ = [
+    "clinical",
+    "longitudinal",
+    "reportable_conditions"
+]
+
```
kairstenfay: Here, I went with the "old style" of `__init__` subcommand loading because
1. I don't fully understand how the "new style" works, and 
2. I had a boatload of other helper functions to import.

Thoughts? Suggestions?

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -301,6 +299,14 @@ def drop_unnecessary_columns(df: pd.DataFrame, columns_to_keep = List[str]) -> p
     return df[columns_to_keep]
 
 
+def insert_static_value_columns(df: pd.DataFrame, column_values: Dict[str, Any]) -> pd.DataFrame:
+    """ TODO """
```
kairstenfay: I'll fix this TODO

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -264,31 +265,47 @@ def parse_sch(sch_filename, output):
...
-
     clinical_records = insert_static_value_columns(clinical_records, {'site': 'SCH'})
     clinical_records = create_encounter_identifier(clinical_records)
     clinical_records = remove_pii(clinical_records)
 
     dump_ndjson(clinical_records)
 
 
+def add_insurance(df: pd.DataFrame) -> pd.DataFrame:
+    """ """ # TODO
```
kairstenfay: will handle this TODO

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -264,31 +265,47 @@ def parse_sch(sch_filename, output):
...
+    df['MedicalInsurance'] = df.apply(insurance, axis=1)
+    return df
+
+
+def insurance(series: pd.Series) -> pd.Series:
+    """ """ # TODO
```
kairstenfay: will handle this TODO

kairstenfay:
[Ingest new SCH clinical data](https://trello.com/c/nAUlByxY/266-ingest-new-sch-clinical-data)

tsibley:
Rebased this onto master to pick up new changes and brought in new changes from https://github.com/seattleflu/id3c/pull/88

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -291,6 +292,15 @@ def parse_sch(sch_filename, output):
 
 
 
+
+def drop_unnecessary_columns(df: pd.DataFrame, columns_to_keep = List[str]) -> pd.DataFrame:
+    """
+    Drops all columns from a given *df* except those explicitly listed in *columns_to_keep*. Returns
+    the new DataFrame.
+    """
+    return df[columns_to_keep]
```
tsibley: On the face of it, this seems like refactoring gone too far. What's the motivation here? (The commit message doesn't say why, just what. :-)

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -359,9 +368,7 @@ def parse_kp(kp_filename, kp_specimen_manifest_filename, output):
     # Convert dtypes
     clinical_records["encountered"] = pd.to_datetime(clinical_records["encountered"])
 
-    # Insert static value columns
-    clinical_records["site"] = "KP"
-
+    clinical_records = insert_static_value_columns(clinical_records, {"site": "KP"})
```
tsibley: Ditto with this refactoring. It's not clear to me why this change is an improvement.

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -264,31 +265,49 @@ def parse_sch(sch_filename, output):
...
+    Adds a new column for insurance type to a given *df*. Returns the new
+    DataFrame.
+    """
+    def insurance(series: pd.Series) -> pd.Series:
+        """ Returns an array of unique insurance types from a given *series*. """
+        insurance_columns = ['insurance_1', 'insurance_2']
+        insurances = [ series[i] for i in insurance_columns if not pd.isna(series[i])]
+        return list(set(insurances))
+
+    df['MedicalInsurance'] = df.apply(insurance, axis=1)
```
tsibley: Using `axis='columns'` is clearer than `axis=1`, at least to me. :-)

lib/seattleflu/id3c/cli/command/clinical.py
kairstenfay: That's fair. I thought it made the code more readable. By refactoring it and naming the function after what it does, I removed a comment that said "Drop unnecessary columns".
If you think that's maybe overkill then I understand.

lib/seattleflu/id3c/cli/command/clinical.py
kairstenfay: Yeah.. this made more sense when we were adding multiple static values to a DataFrame across multiple lines, but the motivation is no longer clear to me for this refactor. I'm ok with dropping this commit.

lib/seattleflu/id3c/cli/command/clinical.py
tsibley: Nod. I think it's skewing a little too far. Indexing/subsetting operations like `df[columns]` are a pretty fundamental syntax. Wrapping that up into a function means more cognitive overhead (for me) because I have to recall if the function is just `df[columns]` or something more involved (with the latter being my assumed default since it's a function). It also expands a fundamental op from 2 lines (comment + code) to 7 (function def + call).

lib/seattleflu/id3c/cli/command/__init__.py
tsibley: I'm not sure what you mean by new vs. old style…? I think the code looks fine, though.

PR #19, opened 2019-11-23
Allow augur-build-exporter to read warehouse.tract

tsibley:
This is necessary for the science map export's joining of the augur
build metadata with tracts to produce GeoJSON.  It seems appropriate to
use this role instead of making a new one since the augur build will
also start using the tract table (and other location tables) with due
time.


tsibley:
Going to merge this to stop the errors instead of waiting for review, but I'm happy to make any changes arising from review afterwards.

PR #22, opened 2019-12-02
Fix etl clinical sex bug

kairstenfay:
We were previously converting unknown sex names to 'other' without raising a warning or error. 
Now, raise an Exception if an unknown sex name is provided to the `sex()` mapping function.


kairstenfay:
[Fix SCH data, clinical etl `sex` bug](https://trello.com/c/E5wR0Kdo/275-fix-sch-data-clinical-etl-sex-bug)

PR #24, opened 2019-12-04
Fix sch clinical bug

kairstenfay:
This PR reflects changes recently merged into master that make mapping sex and insurance names more strict. 

No longer will unknown names automatically be coded to "other", but rather they will raise an Exception.

This modified clinical ETL with a bump in revision number has been tested with a full run of `etl clinical` on a new clone of the production database.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/etl/clinical.py
@@ -40,7 +40,7 @@ LOG = logging.getLogger(__name__)
 # clinical records lacking this revision number in their log.  If a
 # change to the ETL routine necessitates re-processing all clinical records,
 # this revision number should be incremented.
-REVISION = 3
+REVISION = 4
 
 
 @etl.command("clinical", help = __doc__)
--- a/lib/seattleflu/id3c/cli/command/etl/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/etl/clinical.py
@@ -181,7 +181,9 @@ def sex(sex_name) -> str:
 
     def standardize_sex(sex):
         try:
-            sex = sex.lower()
+            if isinstance(sex, str):
+                sex = sex.lower()
+
             return sex if sex in sex_map.values() else sex_map[sex]
         except KeyError:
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Update bad SCH sex data](https://trello.com/c/cFJePpHJ/277-update-bad-sch-sex-data)

lib/seattleflu/id3c/cli/command/etl/clinical.py
```diff
@@ -275,6 +277,13 @@ def insurance(insurance_response: Optional[Any]) -> list:
         "medicare": "government",
         "tricare": "government",
         "other": "other",
+        "care": "other",
+        "caid": "other",
+        "comm": "other",
+        "self": "other",
+        "tce": "other",
+        "case rate": "other",
+        "wc": "other",
```
tsibley: Is the idea here that this lets the ETL complete on all existing data but in the future we'll update these new mappings to non-other values and bump the revision again?

tsibley: Izzy thinks (and I agree) safe to assume that CARE is Medicare (government), CAID is Medicaid (government), COMM is Commercial (private). Can we map those in a new commit for the benefit of the UW data?

kairstenfay: > Izzy thinks (and I agree) safe to assume that CARE is Medicare (government), CAID is Medicaid (government), COMM is Commercial (private). Can we map those in a new commit for the benefit of the UW data?

Yes, absolutely

kairstenfay: > Is the idea here that this lets the ETL complete on all existing data but in the future we'll update these new mappings to non-other values and bump the revision again?

Yes and no. The ETL does complete on all existing data, but in the future, if an unknown value is given, it will break.
I don't think the revision number would need to be bumped again if we're adding more values that code to "other" or "None". However if we are to change existing mapped values, then yes we would need to bump the revision number.

tsibley:
Thanks, this looks good to merge to me.

PR #15, opened 2019-10-25
REDCap ETL

tsibley:
@joverlee521 @kairstenfay I combined your work so far from #10 and #14 on this branch (`redcap-etl`) and did some refactoring into core ID3C which is on seattleflu/id3c#78 (also the `redcap-etl` branch).

I have not yet reviewed in detail the actual construction of the FHIR documents or the resulting documents themselves, but I'd like to and will get eyes on it when I get back from my trip.

Let's consolidate our future work onto this single branch for simplicity. This means that we shouldn't force push it without coordinating with each other. Feel free to edit this issue text to keep track of the current state of things while I'm away.

Unblocked todo items:

- [x] Refactor and standardize on commonly needed functions for constructing FHIR data structures into shared functions in a module in this repo. We will likely want to move this into core ID3C eventually, but it'll be easier to do a first pass here on this branch.

- [x] Test on a fresh set of DETs (produced with `id3c redcap-det generate`) for each project after clearing the existing ones locally.

- [ ] Check that our output bundles validate. It might be nice to do this automatically within the ETL themselves. This might be a good thing to put in core ID3C as part of `redcap_det.command_for_project()`?

Blocked todo items:

- [x] Geocoding (blocked on integration of geocoding functions into ID3C)
- [x] Deidentification via individual and address hashing (blocked on integration of this and geocoding functions into ID3C)
- [x] Update swab and send ETL for new REDCap project (blocked on the new project being finalized... when will this be?)
- [ ] ETLs for both household studies, which are still in flux in REDCap. I can still take on these, but want to wait until they stabilize.

PR #25, opened 2019-12-07
Redcap etl changes

kairstenfay:
I made a lot of changes to the `redcap-etl` branch, so I thought it might be preferable to create a new PR with those changes since our last, group-review on #15.

I'm happy to rebase this onto the `redcap-etl` branch (or vice-versa) if that's preferable.

The only outstanding task is to run the redcap-etl produced FHIR documents through validation. We're discussing in the #id3c slack channel what this looks like.


tsibley:
Pushed a bunch of changes just now. May push a few more tomorrow before deploy.

joverlee521:
@kairstenfay The reason why the Travis checks are failing is because it uses the master branch of id3c (see [.travis.yml](https://github.com/seattleflu/id3c-customizations/blob/master/.travis.yml#L19)), where `generate_hash` requires [two parameters](https://github.com/seattleflu/id3c/blob/master/lib/id3c/cli/command/de_identify.py#L94). 

We've changed this on the `etl-fhir` branch, which is why it doesn't error out when we run pytest locally.

PR #26, opened 2020-01-10
Add view shipping.fhir_encounter_details_v1

joverlee521:
This creates custom view of `encounter.details` that deconstructs the FHIR resources "Condition" and "QuestionnaireResponse".

This will make updating the modeling views _much_ easier!


joverlee521:
I wasn't sure if this should live in `id3c` or `id3c-customizations`. Ultimately decided should be in customizations due to the `linkId` being specific to SFS.

joverlee521:
[Add shipping view to deconstruct FHIR resources within encounter details](https://trello.com/c/pb4dPUZF/310-add-shipping-view-to-deconstruct-fhir-resources-within-encounter-details)

tsibley:
> I wasn't sure if this should live in `id3c` or `id3c-customizations`. Ultimately decided should be in customizations due to the `linkId` being specific to SFS.

I agree.

> Edit: Although this does mean we need to move all the `incidence_model_observation` views over to `id3c-customization` as well. Not sure if it's worth doing that now, also not quite sure how to do this properly. Do we drop the view within ID3C core and then add the views to ID3C-customization?

I agree those views (and several others in core) should ideally live in our customizations. That they're in core is a historical holdover. The sqitch changes should drop + (re-)create, but if we wanted, we could avoid actually dropping them by deploying both the core change and customizations change with sqitch's `--log-only` option. I don't know if it's definitely worth doing this now or not, but we will be forced to at some point by inter-view dependencies. So maybe best to do now? Do you have a sense of how gnarly it'll be?

schema/deploy/shipping/views.sql
```diff
@@ -164,4 +164,80 @@ create or replace view shipping.return_results_v1 as
...
     'View of barcodes and presence/absence results for return of results on website';
 
+
+create or replace view shipping.fhir_encounter_details_v1 as
+
+    with
+        fhir as (
+          select encounter_id,
+                 array_agg(distinct condition."id") as symptoms,
+                 jsonb_object_agg(response."linkId", response.answer) as responses
```
tsibley: This aggregate and others below may want an explicit ordering applied for consistency, particularly since subsequent answers for the same `linkId` will overwrite previous ones. The docs for [_Aggregate Expressions_](https://www.postgresql.org/docs/10/sql-expressions.html#SYNTAX-AGGREGATES) include examples of aggregate ordering.

schema/deploy/shipping/views.sql
```diff
@@ -164,4 +164,80 @@ create or replace view shipping.return_results_v1 as
...
+            from warehouse.encounter,
+                 jsonb_to_recordset(details -> 'Condition') as condition("id" text),
+                 jsonb_to_recordset(details -> 'QuestionnaireResponse') as q("item" jsonb),
+                 jsonb_to_recordset(q."item") as response("linkId" text, "answer" jsonb)
```
tsibley: Joins to `jsonb_to_recordset` (and other set-returning functions) are implicitly inner lateral joins when used like this. That means if there are no output rows returned by the function (e.g. input is null or empty), then the left hand side input rows will not be included in the final result set. The consequence of this in this view is that a missing piece of FHIR details or questionnaire response will end up removing either responses for related columns or even the entire row for an encounter.

I think this is easiest demonstrated by example:

```sql
with x(a,b) as (
    values
    ('b is null', null::jsonb),
    ('b has an object', '[{"baz": 13}]'),
    ('b has empty object', '[{}]'),
    ('b is empty array', '[]')
)
select * from x, jsonb_to_recordset(b) as b(baz integer);
```
```
         a          │       b       │ baz 
────────────────────┼───────────────┼─────
 b has an object    │ [{"baz": 13}] │  13
 b has empty object │ [{}]          │    
(2 rows)
```

compared to

```sql
select * from x left join jsonb_to_recordset(b) as b(baz integer) on true;
```
```
         a          │       b       │ baz 
────────────────────┼───────────────┼─────
 b is null          │               │    
 b has an object    │ [{"baz": 13}] │  13
 b has empty object │ [{}]          │    
 b is empty array   │ []            │    
(4 rows)
```

joverlee521:
> I agree those views (and several others in core) should ideally live in our customizations. That they're in core is a historical holdover. The sqitch changes should drop + (re-)create, but if we wanted, we could avoid actually dropping them by deploying both the core change and customizations change with sqitch's `--log-only` option. I don't know if it's definitely worth doing this now or not, but we will be forced to at some point by inter-view dependencies. So maybe best to do now? Do you have a sense of how gnarly it'll be?

I don't think it will be _that_ bad? My only thought is if we do drop the views in core, we need to remember to remove the `drop view...` statements in `shipping/views` on our next rework of shipping views so that the views that were ported over do not get dropped again. (Although I can also include this second rework as part of the porting process)

tsibley:
Nod. I think including the second rework during the porting process makes sense. Sounds like we might as well do the port now?

joverlee521:
> Nod. I think including the second rework during the porting process makes sense. Sounds like we might as well do the port now?

Sounds good to me. I can work on that after fixing/merging this PR.

joverlee521:
@tsibley I've added explicit ordering to all aggregates and used `left join` where I thought was necessary. 

Using the left join required some additional conditionals to not aggregate NULL values. Not entirely sure if this was the correct way to doing it.

joverlee521:
[Update modeling views to account for new encounter details (from FHIR ETL)](https://trello.com/c/9ZuXsq9D/285-update-modeling-views-to-account-for-new-encounter-details-from-fhir-etl)

kairstenfay:
I just noticed I get a permission denied error when running this query locally with these changes:
```
seattleflu=# select * from shipping.incidence_model_observation_v1 where encountered_date >= '2019-10-01';
ERROR:  permission denied for view fhir_encounter_details_v1

```

joverlee521:
> I just noticed I get a permission denied error when running this query locally with these changes:
> 
> ```
> seattleflu=# select * from shipping.incidence_model_observation_v1 where encountered_date >= '2019-10-01';
> ERROR:  permission denied for view fhir_encounter_details_v1
> ```

That's weird, shouldn't you be a Superuser locally?

Although this does bring up the question of who should have access to this view. I'm thinking at least "incidence-modeler" since the view is joined with `incidence_model_observation`

tsibley:
> Although this does bring up the question of who should have access to this view. I'm thinking at least "incidence-modeler" since the view is joined with `incidence_model_observation`

I concur. The `dumper` and `reporter` roles will pick it up automatically due to default privileges.

schema/deploy/shipping/views.sql
```diff
@@ -164,4 +164,100 @@ create or replace view shipping.return_results_v1 as
...
+          group by encounter_id
+        ),
+
+        vaccine as (
+            --Builds vaccine jsonb object without null values that spans multiple rows.
+            --Expand the rows into key/value pairs with jsonb_each and
+            --aggregate that back into a single jsonb value.
+            --  -Jover 10 Jan 2020
+            select encounter_id,
+                   jsonb_object_agg(t.k, t.v order by t.k) filter (where t.k is not null) as vaccine
```
tsibley: Similarly here and elsewhere with ordering by `t.k`.

PR #28, opened 2020-01-15
Add sites with details

joverlee521:
Adds sites with the types and categories described here: https://docs.google.com/document/d/1SqAlXuZttDXDdrRqdeSS910ecrM737MNRacBXHd6ePg/edit

The only site that is not on here is `Unknown` which I'm not entirely sure what to do with.
Also, `KaiserPermanente` is labeled as `clinic` but we have it as `retrospective`

### DIFF
--- a/schema/deploy/warehouse/site/data.sql
+++ b/schema/deploy/warehouse/site/data.sql
@@ -1,0 +1,45 @@
+-- Deploy seattleflu/id3c-customizations:warehouse/site/data to pg
+-- requires: seattleflu/schema:warehouse/site
+
+begin;
+
+insert into warehouse.site(identifier, details)
+    values
+        ('ChildrensHospitalBellevue',               '{"category": "clinic",     "type": "clinic"}'),
+        ('ChildrensHospitalSeattle',                '{"category": "clinic",     "type": "clinic"}'),
+        ('ChildrensSeaMar',                         '{"category": "clinic",     "type": "clinic"}'),
+        ('ColumbiaCenter',                          '{"category": "community",  "type": "workplace"}'),
+        ('Costco',                                  '{"category": "community",  "type": "workplace"}'),
+        ('DESC',                                    '{"category": "community",  "type": "shelter"}'),
+        ('FredHutchLobby',                          '{"category": "community",  "type": "workplace"}'),
+        ('Harborview',                              '{"category": "hospital",   "type": "longitudinalInpatient"}'),
+        ('HarborviewLobby',                         '{"category": "community",  "type": "workplace"}'),
+        ('HealthSciencesLobby',                     '{"category": "community",  "type": "collegeCampus"}'),
+        ('HealthSciencesRotunda',                   '{"category": "community",  "type": "collegeCampus"}'),
+        ('HUB',                                     '{"category": "community",  "type": "collegeCampus"}'),
+        ('HutchKids',                               '{"category": "community",  "type": "childcare"}'),
+        ('KaiserPermanente',                        '{"category": "clinic",     "type": "clinic"}'),
+        ('KingStreetStation',                       '{"category": "community",  "type": "publicSpace"}'),
+        ('PioneerSquare',                           '{"category": "clinic",     "type": "clinic"}'),
+        ('RetrospectiveChildrensHospitalSeattle',   '{"category": "hospital",   "type": "retrospective"}'),
+        ('RetrospectiveHarborview',                 '{"category": "hospital",   "type": "retrospective"}'),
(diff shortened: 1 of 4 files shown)


joverlee521:
[Revise site types/labels](https://trello.com/c/IzuwKzXh/265-revise-site-types-labels)

kairstenfay:
> The only site that is not on here is `Unknown` which I'm not entirely sure what to do with.
> Also, `KaiserPermanente` is labeled as `clinic` but we have it as `retrospective`

This makes me wonder if there's both Kaiser Permanente retrospective data (from Mike Jackson) and a new Kaiser Permanente kiosk site (with data from REDCap). We should verify this before merging, perhaps with Robin in #informatics?

joverlee521:
Putting this on hold until we get clarification about KaiserPermanente(see [slack](https://seattle-flu-study.slack.com/archives/CCAA9RBFS/p1579290122003900))

kairstenfay:
> I find that Google doc pretty confusing and not very considered. For example, it labels `RetrospectiveHarborview` as a "workplace", which can't be right. Glad to see you've kept that "retrospective"!
> 
> Minor question: Why is `Harborview` labeled `longitudinalInpatient`?

I'll wager that the labels like 'Workplace' only apply to the 'Community' column. I do not believe the table is meant to be read across the Community/Clinic/Hospital columns. Rather, they are independent and only share the 'Adult'/'Child' labels.

joverlee521:
> Minor question: Why is `Harborview` labeled `longitudinalInpatient`?

I based this off of Mike's [comment](https://seattle-flu-study.slack.com/archives/CL1Q38ATG/p1577470665018900?thread_ts=1577397176.011700&cid=CL1Q38ATG) and the fact that all samples associated with the site `Harborview` comes from `Inpatient Swabs and Blood Sample Records`

schema/deploy/warehouse/site/data.sql
```diff
@@ -0,0 +1,45 @@
...
+        ('RetrospectiveChildrensHospitalSeattle',   '{"category": "hospital",   "type": "retrospective"}'),
+        ('RetrospectiveHarborview',                 '{"category": "hospital",   "type": "retrospective"}'),
+        ('RetrospectiveNorthwest',                  '{"category": "hospital",   "type": "retrospective"}'),
+        ('RetrospectiveUWMedicalCenter',            '{"category": "hospital",   "type": "retrospective"}'),
+        ('SeaTacDomestic',                          '{"category": "community",  "type": "publicSpace"}'),
+        ('SeaTacInternational',                     '{"category": "community",  "type": "publicSpace"}'),
+        ('SeattleCenter',                           '{"category": "community",  "type": "publicSpace"}'),
+        ('self-test',                               '{"category": "community",  "type": "self-test"}'),
+        ('StMartins',                               '{"category": "community",  "type": "shelter"}'),
+        ('swabNSend',                               '{"category": "community",  "type": "self-test"}'),
```
kairstenfay: I think that the metrics team is interested in separating self-test and swab-n-send, per [their community break-down table](https://docs.google.com/spreadsheets/d/1eAmWEULmgogciQnQtYEaTGFnl9hkt7snTgF0u-xNTi4/edit#gid=10263779).

joverlee521: I feel like that should be looking at the site identifier itself? They are both "self-test" type

tsibley: Maybe `self-swab` is a more accurate/precise type? Both study arms swab themselves, but only the `self-test` / home test arm is actually _testing_ themselves with an at-home device (though that's on hold right now!).

But maybe introducing another term is more confusing...

kairstenfay: > I feel like that should be looking at the site identifier itself? They are both "self-test" type

I'm just thinking keeping self-test and swab-n-send separate will make metrics calculations much easier, seeing as the current site types are all defined by metrics-related needs.

joverlee521: Taking another look at https://docs.google.com/document/d/1SqAlXuZttDXDdrRqdeSS910ecrM737MNRacBXHd6ePg/edit

They have `Swab-and-send` and `Swab-and-send with self-test` labeled as `Home` for site_type

tsibley: > seeing as the current site types are all defined by metrics-related needs

Nod, though not just metrics, but also modeling and the Nextstrain build.

tsibley: A site type of `home` makes sense to me.

PR #29, opened 2020-01-16
clinical parse-sch: Rename test result column

kairstenfay:
In the most recent SCH monthly clinical data pull from 15 Jan 2020, one
of the test result columns changed names. Rename it in the parsing code
to avoid a KeyError.


kairstenfay:
[Ingest Year 2 SCH encounter data (S3)](https://trello.com/c/HsOZRRvb/318-ingest-year-2-sch-encounter-data-s3)

PR #30, opened 2020-01-17
REDCap DET etl updates

joverlee521:
Bumps both kiosk and swab-n-send to REVISION 2 to include updates in response to warning logs in recent etl runs. 

Kiosks:
* Returns `project_id-record_id` if it cannot extract zip code from record. This allows ingestion of participants without a permanent residence and participants who are uncomfortable with sharing their address.
* Include `acute_symptom_onset` as a QuestionnaireResponse item. This was a request from the BBI team.

Swab-n-send:
* Make `collection_date` optional since it is not used by the FHIR etl.
* Do not create Patient Resource if a `patient_id` could not be generated.
* Require Patient and Encounter Resources to be created in order to process a bundle.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
@@ -52,6 +52,11 @@ REQUIRED_INSTRUMENTS = [
 def redcap_det_swab_n_send(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_record: dict) -> Optional[dict]:
     location_resource_entries = locations(db, cache, redcap_record)
     patient_entry, patient_reference = create_patient(redcap_record)
+
+    if not patient_entry:
+        LOG.warning("Skipping enrollment with insufficient information to construct patient")
+        return None
+
     encounter_entry, encounter_reference = create_encounter(redcap_record, patient_reference, location_resource_entries)
     questionnaire_entry = create_questionnaire_response(redcap_record, patient_reference, encounter_reference)
     specimen_entry, specimen_reference = create_specimen(redcap_record, patient_reference)
@@ -193,6 +198,9 @@ def create_patient(record: dict) -> tuple:
         gender      = gender,
         birth_date  = record['birthday'],
         postal_code = record['home_zipcode_2'])
+
+    if not patient_id:
+        return None, None
 
     patient_identifier = create_identifier(f"{INTERNAL_SYSTEM}/individual", patient_id)
     patient_resource = create_patient_resource([patient_identifier], gender)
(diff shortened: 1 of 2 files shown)


joverlee521:
[Resolve warnings during REDCap DET and FHIR ETLs on real data](https://trello.com/c/sBpnqH8H/313-resolve-warnings-during-redcap-det-and-fhir-etls-on-real-data)

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
```diff
@@ -352,18 +352,11 @@ def specimen_barcode(record: Any) -> str:
         value = barcode
     )
 
-    if not record.get('collection_date'):
-        LOG.warning("Could not create Specimen Resource due to lack of collection date.")
-        return None, None
-
     # YYYY-MM-DD in REDCap
-    collected_time = record['collection_date']
+    collected_time = record['collection_date'] if record['collection_date'] else None
```
kairstenfay: You can shorten this with `record.get('collection_date')

joverlee521: The REDCap API returns empty strings for all values that are not filled so it would never be `None`

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
```diff
@@ -352,18 +352,11 @@ def specimen_barcode(record: Any) -> str:
...
 
     # YYYY-MM-DD HH:MM:SS in REDCap
-    received_time = record['samp_process_date'].split()[0]
-    if not received_time:
-        LOG.warning("No sample process date found. Using collection date instead.")
-        received_time = collected_time
+    received_time = record['samp_process_date'].split()[0] if record['samp_process_date'] else None
```
kairstenfay: It looks like we're doing more than just changing the handling of `collection_date` here. Perhaps we can either split the commit or edit the current commit's language to reflect this.

lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py
tsibley: `record['collection_date'] or None` is the shorter version then. But the long version is also just fine. :-)

tsibley:
> * Do not create Patient Resource if a `patient_id` could not be generated.

Does this mean that we have `warehouse.individual` records with bad identifiers?

joverlee521:
> > * Do not create Patient Resource if a `patient_id` could not be generated.
> 
> Does this mean that we have `warehouse.individual` records with bad identifiers?

I don't think so, I just added this just in case. I believe what would happen is there would be FHIR documents created with Patient Resources where the `identifier.value` is null. Then the `upsert_individual` in the FHIR ETL would fail with a null identifier.

PR #32, opened 2020-01-22
Update metadata for augur build view

kairstenfay:
Update the `metadata_for_augur_build_v2` to format dates in YYYY-MM-DD format.

Create new view from v2, `metadata_for_augur_build_v3`, that
1. coalesces encounter date with collection date from sample details 
2. uses the new 'category' option from `site.details`.

Addresses two aspects of [this issue](https://github.com/seattleflu/augur-build/issues/32) by reformatting dates and adding an `age_range_coarse` column.

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -560,3 +560,28 @@ comment on view shipping.observation_with_presence_absence_result_v2 is
   'Joined view of shipping.incidence_model_observation_v3 and shipping.presence_absence_result_v1';
 
 commit;
+
+create or replace view shipping.metadata_for_augur_build_v3 as
+
+    select sample as strain,
+            encountered_date as date,
+            'seattle' as region,
+            -- XXX TODO: Change to PUMA and neighborhoods
+            residence_census_tract as location,
+            'Seattle Flu Study' as authors,
+            age_range_coarse,
+            case
+                when age_range_coarse <@ '[0 mon, 18 years)'::intervalrange then 'child'
+                else 'adult'
+            end as age_category,
+            warehouse.site.details->>'category' as site_category,
+            residence_census_tract,
+            flu_shot,
+            sex
+
+      from shipping.incidence_model_observation_v2
+      join warehouse.encounter on encounter.identifier = incidence_model_observation_v2.encounter
+      join warehouse.site on site = site.identifier;
+
+comment on view shipping.metadata_for_augur_build_v3 is
+		'View of metadata necessary for SFS augur build';
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -564,7 +564,10 @@ commit;
 create or replace view shipping.metadata_for_augur_build_v3 as
 
     select sample.identifier as strain,
-            encountered_date as date,
+            coalesce(
+              encountered_date,
+              date_or_null(sample.details->>'date')
+            ) as date,
             'seattle' as region,
             -- XXX TODO: Change to PUMA and neighborhoods
             residence_census_tract as location,
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -563,7 +563,7 @@ commit;
 
 create or replace view shipping.metadata_for_augur_build_v3 as
 
-    select sample as strain,
+    select sample.identifier as strain,
             encountered_date as date,
             'seattle' as region,
             -- XXX TODO: Change to PUMA and neighborhoods
@@ -579,9 +579,11 @@ create or replace view shipping.metadata_for_augur_build_v3 as
             flu_shot,
             sex
 
-      from shipping.incidence_model_observation_v2
-      join warehouse.encounter on encounter.identifier = incidence_model_observation_v2.encounter
-      join warehouse.site on site = site.identifier;
+      from warehouse.sample
+      left join shipping.incidence_model_observation_v2 on sample.identifier = incidence_model_observation_v2.sample
+      left join warehouse.site on site.identifier = incidence_model_observation_v2.site
+
+     where sample.identifier is not null;
 
 comment on view shipping.metadata_for_augur_build_v3 is
 		'View of metadata necessary for SFS augur build';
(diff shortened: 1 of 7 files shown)


kairstenfay:
[Update format for `date` in `shipping.augur-build-metadata`](https://trello.com/c/SJKUrMDI/298-update-format-for-date-in-shippingaugur-build-metadata)

kairstenfay:
[Include samples without encounter data in metadata table for Augur build](https://trello.com/c/VAQxlVoM/326-include-samples-without-encounter-data-in-metadata-table-for-augur-build)

joverlee521:
I _think_ you're missing a verify script for `metadata_for_augur_build_v3`

kairstenfay:
> I _think_ you're missing a verify script for `metadata_for_augur_build_v3`

Good catch. Fixed it.

schema/deploy/shipping/views.sql
```diff
@@ -558,3 +559,31 @@ comment on view shipping.observation_with_presence_absence_result_v2 is
...
+              date_or_null(sample.details->>'date')
+            ) as date,
+            'seattle' as region,
+            -- XXX TODO: Change to PUMA and neighborhoods
+            residence_census_tract as location,
+            'Seattle Flu Study' as authors,
+            case
+                when age_range_coarse <@ '[0 mon, 18 years)'::intervalrange then 'child'
+                else 'adult'
+            end as age_category,
```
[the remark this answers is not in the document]
kairstenfay: Done.

kairstenfay:
> Thanks, this looks good except for some sqitch ordering issues. It looks like the order of commits has changes to the views before the first sqitch rework is done, which messes up expectations about isolation of changes.

@tsibley I believe I've fixed this error.

kairstenfay:
I just pushed a fixed-up commit adding a `drop` on the new v3 metadata for augur build view to the revert script.

schema/deploy/shipping/views.sql
```diff
@@ -558,3 +560,35 @@ comment on view shipping.observation_with_presence_absence_result_v2 is
...
+            warehouse.site.details->>'category' as site_category,
+            residence_census_tract,
+            flu_shot,
+            sex
+
+      from warehouse.sample
+      left join warehouse.encounter using (encounter_id)
+      left join shipping.incidence_model_observation_v2 on encounter.identifier = incidence_model_observation_v2.encounter
```
tsibley: I just realized that these joins aren't quite right if an encounter has multiple samples: they'll produce the cross-product of the `warehouse.sample` rows and the `shipping.incidence_model_observation_v2`, each of which has one row per sample. The join on `shipping.incidence_model_observation_v2` should use `sample.identifier = incidence_model_observation_v2.sample`.

schema/deploy/shipping/views.sql
```diff
@@ -558,3 +560,35 @@ comment on view shipping.observation_with_presence_absence_result_v2 is
...
+            ) as date,
+            'seattle' as region,
+            -- XXX TODO: Change to PUMA and neighborhoods
+            residence_census_tract as location,
+            'Seattle Flu Study' as authors,
+            age_range_coarse,
+            case
+                when age_range_coarse <@ '[0 mon, 18 years)'::intervalrange then 'child'
+                else 'adult'
+            end as age_category,
```
tsibley: I'm also just realizing that this case is wrong. If `age_range_course` is null, then `adult` is returned. The fix is to compare directly to boolean values, e.g.:

```sql
            case age_range_coarse <@ '[0 mon, 18 years)'::intervalrange
                when 't' then 'child'
                when 'f' then 'adult'
            end as age_category,
```

which will only be `child` or `adult` when the age is known (not null) and otherwise be null.

schema/deploy/shipping/views.sql
tsibley: Similarly, I think that it'd be good to filter out `warehouse.sample` rows without an `identifier`. These will be ones for which we only know the `collection_identifier` so far.

schema/deploy/shipping/views.sql
```diff
@@ -558,3 +560,35 @@ comment on view shipping.observation_with_presence_absence_result_v2 is
...
 
 commit;
+
+create or replace view shipping.metadata_for_augur_build_v3 as
+
+    select sample as strain,
+            coalesce(
+              cast(encountered as date),
+              case
+                when date_or_null(sample.details->>'date') <= current_timestamp::date
```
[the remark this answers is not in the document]
tsibley: (Update: I wrote this comment before realizing I would do some fixups, so I did end up changing it.)

schema/deploy/shipping/views.sql
```diff
@@ -48,17 +48,19 @@ create or replace view shipping.reportable_condition_v1 as
     order by encountered desc;
 
 
+drop view shipping.metadata_for_augur_build_v2;
 create or replace view shipping.metadata_for_augur_build_v2 as
 
-    select  sample as strain,
-            encountered as date,
+    select sample as strain,
+            cast(encountered as date) as date,
```
kairstenfay: @tsibley should I also changed to using `encounter_date` instead of casting `encountered` like you did in v3? It would require joining on a different version of `shipping.incidence_model_observation*`, though.

schema/deploy/shipping/views.sql
kairstenfay: Nice catch.

kairstenfay:
Thanks, Thomas for catching those bugs. I have one more question above. Otherwise is this good to merge.

schema/deploy/shipping/views.sql
tsibley: I used `encounter_date` in `metadata_for_augur_build_v3` to a) get [correct timezone handling](https://github.com/seattleflu/id3c-customizations/blob/c8d3f7abb889d330b243c2529c017c2eadd9d955/schema/deploy/shipping/views.sql#L267) and b) remove the need for the join `warehouse.encounter`.

For `metadata_for_augur_build_v2` here, I wouldn't change the join, but use the same timezone conversion logic. It may also be that we want to set the default database timezone to US/Pacific.

tsibley:
I'd say good to merge as-is since we're going to eventually drop `metadata_for_augur_build_v2`. But would also be fine to fix the timezone handling first and then merge.

kairstenfay:
I fixed the timezone handling in v2 (in a fixed-up commit), even though it will be dropped eventually.

tsibley:
Thanks!

PR #33, opened 2020-01-24
views: Add new view of best available site data

kairstenfay:
Metrics calculations, especially for community samples, have become
increasingly important in Y2 of the SFS. These calculations are often
requested for samples even before encounter their data is received. To
limit the number of long, ugly queries in Metabase used to calculate
these metrics, create a view that contains some hard-coded [information
from Peter](https://seattle-flu-study.slack.com/archives/CCAA9RBFS/p1579234521000200?thread_ts=1579205839.056600&cid=CCAA9RBFS)
about how the specimen manifest `swab_site` column maps to our site
types from `warehouse.site.details`. Namely, we're creating two new
sources of information: `best_available_site_type` and
`best_available_site_category` columns.


kairstenfay:
[Define master view or two wrapping up ugly queries](https://trello.com/c/YcXue8Pw/286-define-master-view-or-two-wrapping-up-ugly-queries)

kairstenfay:
Added a new `swab_site` that came in recently, `capitol_hill`.

kairstenfay:
@tsibley I ended up with a merge conflict because you merged a commit with a new sqitch tag since I started my rework. Because you didn't edit the same file (`views.sql`), do I still need to update the tag in my rework to match your latest tag?

kairstenfay:
> Added a new `swab_site` that came in recently, `capitol_hill`.

I just fixed this addition (I had accidentally placed it under `workplace` instead of `publicSpace`).

Also, to save a future rework, took a guess at the Westlake Station `swab_site` label by capturing `westlake%`.

schema/deploy/shipping/views.sql
```diff
@@ -557,4 +557,104 @@ create or replace view shipping.observation_with_presence_absence_result_v2 as
...
+
+    with specimen_manifest_site_data as (
+        select distinct on (manifest_id)
+            cast(value->>'sample_id' as integer) as sample_id,
+            trim(both ' ' from document->>'swab_site') as swab_site,
+            trim(both ' ' from document->>'sample_origin') as sample_origin
+        from
+            receiving.manifest,
+            jsonb_array_elements(processing_log)
+        where value->'sample_id' is not null
```
tsibley: Providing a table alias for `jsonb_array_elements` would make it a little more clear where `value` is coming from.

Or, perhaps even more clear, using `jsonb_to_recordset(processing_log) as log(sample_id integer)`, since that both extracts the JSON `sample_id` field, casts it, and assigns a meaningful name.

schema/deploy/shipping/views.sql
```diff
@@ -557,4 +557,104 @@ create or replace view shipping.observation_with_presence_absence_result_v2 as
...
+        order by manifest_id desc
```
tsibley: Since this is a `distinct on (manifest_id)`, there should be an ordering on another column as well so the row that's kept is stable. Refer to https://www.postgresql.org/docs/10/sql-select.html#SQL-DISTINCT for more details.

In this case, I think we want the `sample_id` from the _last_ `processing_log` entry, not the first. This can be done by ordering by the log's timestamp field, for example.

schema/deploy/shipping/views.sql
```diff
@@ -557,4 +557,104 @@ create or replace view shipping.observation_with_presence_absence_result_v2 as
...
+      select
+        sample_id,
+        sample.identifier as sample,
+        coalesce(site_type,
+          case
+              -- First handle sample_origin
+              when sample_origin = 'es'
+                then 'environmental'
+
+              when sample_origin like '%_retro'
```
tsibley: It'd be a little more robust to match case-insensitively using `ilike` instead of `like`.

tsibley:
Not saying anything needs to change now, but one thought on an alternative way of doing this that I think might require less maintenance in the long term is to map `swab_site` names to our actual `warehouse.site` records. This way if we just update a `warehouse.site`'s type or category, the associated `swab_site` values also produce the updated type and category automatically. It would then also, for example, let us store the associated `swab_site` aliases in `warehouse.site.details` and use that for lookup.

kairstenfay:
> Not saying anything needs to change now, but one thought on an alternative way of doing this that I think might require less maintenance in the long term is to map `swab_site` names to our actual `warehouse.site` records. This way if we just update a `warehouse.site`'s type or category, the associated `swab_site` values also produce the updated type and category automatically. It would then also, for example, let us store the associated `swab_site` aliases in `warehouse.site.details` and use that for lookup.

I like this idea so I created https://github.com/seattleflu/id3c-customizations/pull/39

schema/deploy/shipping/views.sql
kairstenfay: I addressed this in a new commit.

schema/deploy/shipping/views.sql
kairstenfay: I addressed this in a new commit.

schema/deploy/shipping/views.sql
kairstenfay: I addressed this in a new commit.

tsibley: :disappointed: Realizing now that there's no equivalent for `similar to`.

kairstenfay:
@tsibley this view has changed dramatically now that it sources new specimen manifest site details from `warehouse.site`. 

There are still many sites that do not exist in warehouse but are being tracked in metrics (especially shelters and environmental sample sites) that I could hard-code into the new view, but I'm hesitant to have two sources of truth for the specimen manifest site matching regex.

schema/deploy/shipping/views.sql
```diff
@@ -593,3 +592,73 @@ create or replace view shipping.metadata_for_augur_build_v3 as
...
+          end,
+
+          manifest.sample_origin
+        ) as site_manifest_details,
+
+        case
+          when site_id is null
+            then 'f'
+          else 't'
+        end as has_encounter_data
```
joverlee521: It's slightly weird to me that `has_encounter_data` is text and not just a boolean.

kairstenfay: Agreed. Good catch. 
I was thinking that because boolean values in Postgres can be matched with ` = 't'` or ` = 'f'` that they could also be inserted as such and convert to boolean. I was mistaken. I just changed this commit to use `FALSE` and `THEN`, and now the `pg_typeof()` of this column is boolean as expected.

joverlee521: FYI, if all you want is the boolean value, this can be shortened to 
`site_id is not null as has_encounter_data`

kairstenfay: Thanks for the tip! Fixed up.

kairstenfay:
> These are likely introduced by joins matching multiple rows.

I found the culprit of the problems. One was that I was selecting `distinct on (manifest_id)` when I should've been using `distinct on (sample_id)`. This change is reflected in a new commit. 

The second problem was that two of our sites are mapped to the same regex: `community_cl?i?nic`. I wasn't sure how to deal with this best so I created a new site called `UnknownClinic` (category & type `clinic`) that stores this regex. 

While in there, I also added an `UnknownAirport` site for the relatively new `airport` label from the manifest data. 

Let me know if you think I should be handling this vague specimen manifest site labels differently.

kairstenfay:
> The second problem was that two of our sites are mapped to the same regex: `community_cl?i?nic`. I wasn't sure how to deal with this best so I created a new site called `UnknownClinic` (category & type `clinic`) that stores this regex.
> 
> While in there, I also added an `UnknownAirport` site for the relatively new `airport` label from the manifest data. I fixed up these two changes into the existing site details commit.
> 
> Let me know if you think I should be handling this vague specimen manifest site labels differently.

I decided after consulting with Jover to abandon this for now. Specimen manifest sites that map to multiple warehouse sites will be ignored.

kairstenfay:
> Looks good. One thing I did notice is this includes samples that do not have identifier. Do we want to filter those out or should there be a `collection_identifier` column in the view?

Good question. The purpose of this view is to provide best available encounter site data for all samples, so I think we should filter out encounters without sample identifiers. I'll change this.

kairstenfay:
> > Looks good. One thing I did notice is this includes samples that do not have identifier. Do we want to filter those out or should there be a `collection_identifier` column in the view?
> 
> Good question. The purpose of this view is to provide best available encounter site data for all samples, so I think we should filter out encounters without sample identifiers. I'll change this.

I made a few more changes in addition to this one plus the previous one about ignoring specimen manifest sites that map to multiple warehouse sites (e.g. `seatac_airport`). Some changes are in new commits and some fixed up into previous commits.

The changes are:
* Samples without identifiers are now removed
* We have three new columns: 
    1. `best_available_encounter_date` (uses collection date from specimen manifest. Similar to what we did in `metadata_for_augur_build` but sources from `receiving.manifest` instead of sample details)
    2. season (currently Y1 or Y2). Uses `best_available_encounter_date`
    3. `best_available_site` because Robin now asks for more specific metrics other than at the `site_type` and `site_category` level. Also because why not :woman_shrugging: 
* The query is more complete, joining on the `site_id` from the samples with specimen manifest auxiliary query rather than relying on the regex in `warehouse.site` to join site data to the final table.

kairstenfay:
Would appreciate your feedback on this @tsibley when you get a chance so I can proceed with refactoring Metabase metrics dashboards.

kairstenfay:
Just force-pushed a small stray tab fixup. I'm not making further changes at this time.

schema/deploy/shipping/views.sql
```diff
@@ -593,3 +592,82 @@ create or replace view shipping.metadata_for_augur_build_v3 as
...
+
+
+create or replace view shipping.sample_with_best_available_encounter_data_v1 as
+
+    with specimen_manifest_data as (
+        select distinct on (sample_id)
+            sample_id,
+            date_or_null(document->>'date') as collection_date,
+            trim(both ' ' from document->>'swab_site') as swab_site,
+            trim(both ' ' from document->>'sample_origin') as sample_origin
```
joverlee521: I'm confused why these have to be pulled from `receiving.manifest`. They should all be within `sample.details`

kairstenfay: I can't remember why I originally used `receiving.manifest` for this view. I may not have realized these details were also in `sample.details`...

kairstenfay: OK, I rewrote the query to use `sample.details` over `receiving.manifest`.

kairstenfay:
I also pushed a small commit that reflects something lost in rebasing: adding the `best_available_site` column.

kairstenfay:
I'm wondering if it would be useful to add `best_available_site_id` in addition to `best_available_site`.

joverlee521:
> I'm wondering if it would be useful to add `best_available_site_id` in addition to `best_available_site`.

Maybe? It really depends on how you are going to use this view. Make it whatever you need ti to be 😄

PR #34, opened 2020-01-25
redcap-det: Fallback to record id-based individual identifier generation

tsibley:
Instead of special casing only zip code and bailing out if other pieces
of information are missing, take a best-effort approach and assume the
record represents a unique participant who hasn't participated
previously and won't participate again in the future.

This is a broadening of the logic we previously used for zip code, and I
think it makes more sense in the bigger picture.  I only realized it
when I was digging into noisy logs, though!  Those same noisy logs are
silenced since missing personal info is now less of an issue with this
more general fallback.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -400,7 +400,7 @@ def generate_patient_hash(names: Iterable[str], gender: str, birth_date: str, po
         ]
 
     if missing(personal_information):
-        LOG.warning(f"All personal information is required to generate a robust patient hash; missing {missing(personal_information)}")
+        LOG.debug(f"All personal information is required to generate a robust patient hash; missing {missing(personal_information)}")
         return None
 
     return generate_hash("\N{UNIT SEPARATOR}".join(personal_information))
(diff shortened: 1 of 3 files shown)


tsibley:
Thanks for the eyes on this! I'll plan to deploy this tomorrow. One expected outcome is that any records we've previously processed that were missing information will cause their existing `warehouse.individual` rows to be orphaned when we next re-process those records (if we do). That's ok, I think, and we can clean them up after the fact.

Which brings me to the question of bumping the ETL revision for both of these routines, so that everything we previously skipped due is imported. Seems like the right thing, yeah?

tsibley:
> Which brings me to the question of bumping the ETL revision for both of these routines, so that everything we previously skipped due [to missing info] is imported. Seems like the right thing, yeah?

@kairstenfay @joverlee521 Would love to get a sanity check on this question from you before I go ahead.

joverlee521:
> > Which brings me to the question of bumping the ETL revision for both of these routines, so that everything we previously skipped due [to missing info] is imported. Seems like the right thing, yeah?
> 
> @kairstenfay @joverlee521 Would love to get a sanity check on this question from you before I go ahead.

If you mean both `kiosk` and `swab-n-send` ETL revision numbers, then yes, that sounds good to me.

PR #35, opened 2020-01-27
etl redcap-det kiosk: Add new site

kairstenfay:
Add a new site for Capitol Hill Light Rail Station with spaces so that
the ETL doesn't break.

This is the minimal effort solution to fix the broken Kiosk ETL. The other solution I considered is to strip spaces from site names when comparing them to our `warehouse.site` identifier equivalent. Thoughts?


tsibley:
> The other solution I considered is to strip spaces from site names when comparing them to our warehouse.site identifier equivalent. Thoughts?

I would leave the mapping explicit for now unless this is a very common issue.

PR #36, opened 2020-01-27
Add new Kiosk sites to warehouse

kairstenfay:
Add the Capitol Hill and Westlake light rail stations from the Kiosk ETL
to the warehouse.

Skip `WestlakeCenter` for now, because Izzy said in the #redcap channel
that we likely won't start enrolling participants there anytime soon, if
at all.

### DIFF
--- a/schema/revert/warehouse/site/data.sql
+++ b/schema/revert/warehouse/site/data.sql
@@ -1,8 +1,11 @@
--- Revert seattleflu/id3c-customizations:warehouse/site/data from pg
+-- Deploy seattleflu/id3c-customizations:warehouse/site/data to pg
 -- requires: seattleflu/schema:warehouse/site
 
 begin;
 
--- No need to revert data.
+delete from warehouse.site
+  where identifier = 'CapitolHillLightRailStation'
+  or identifier = 'WestlakeLightRailStation'
+;
 
 commit;
--- a/schema/deploy/warehouse/site/data@2020-01-30.sql
+++ b/schema/deploy/warehouse/site/data@2020-01-30.sql
@@ -5,7 +5,6 @@ begin;
 
 insert into warehouse.site(identifier, details)
     values
-        ('CapitolHillLightRailStation',             '{"category": "community",  "type": "publicSpace"}'),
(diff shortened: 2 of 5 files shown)


schema/revert/warehouse/site/data.sql
```diff
@@ -1,8 +1,45 @@
--- Revert seattleflu/id3c-customizations:warehouse/site/data from pg
+-- Deploy seattleflu/id3c-customizations:warehouse/site/data to pg
 -- requires: seattleflu/schema:warehouse/site
 
 begin;
 
--- No need to revert data.
+insert into warehouse.site(identifier, details)
```
joverlee521: Do we need the revert script?

kairstenfay: Not sure. I just left the automatically generated one from sqitch.

tsibley: Revert scripts for data modification statements (DML), as opposed to data definition statements (DDL), are tricky, particularly when they update data in place like this.

I'd either:

1. Amend the current revert script here to include statements removing the two new sites, so it's actually a revert in concept.

2. Remove the revert script entirely, acknowledging explicitly that automatic revert isn't really something we're trying to support here.

The first seems nicer.

kairstenfay: Deleting a site from warehouse will only work if there is encounter that references its site_id.

tsibley: I assume you meant if there is _not_ an encounter. That is true, but I think in general revert scripts for DML (not DDL) don't need to be expected to run at arbitrary points in the future, only immediately after a failed deploy (when there shouldn't be linked data yet).

kairstenfay: Correct, I meant _not_ an encounter.

joverlee521:
This looks good to me!

kairstenfay:
I found one small mistake that occurred when rebasing onto master: the new sites slipped into the tagged `data.sql` file. I just removed them and re-pushed. Merging now.

PR #37, opened 2020-01-28
Pumas & Neighborhood Districts

joverlee521:
This PR is dependent on https://github.com/seattleflu/id3c/pull/103

Adds download & import for PUMAs and neighborhood districts.

Uses the new `--hierarchy-by-feature` option that is added in https://github.com/seattleflu/id3c/pull/103 to include PUMAs and neighborhood district for hierarchy of tracts in ID3C. 

Neighborhood districts are created via a union of Community Reporting Areas(CRAs) and looks like it matches what is currently in [seattle-geojson](https://github.com/seattleflu/seattle-geojson/blob/master/seattle_geojsons/2016_seattle_neighborhoods.geojson) but doesn't include bodies of water!


joverlee521:
[Switch to exporting neighborhood and PUMA instead of census tracts](https://trello.com/c/F4RMJCzJ/262-switch-to-exporting-neighborhood-and-puma-instead-of-census-tracts)

locations/Snakefile
```diff
@@ -73,17 +86,81 @@ rule download_state_tracts_boundaries:
...
+        neighborhood_hierarchies = (
+            pandas
+            .read_excel(input.neighborhood, dtype = object)
+            .assign(
+                feature_identifier = lambda df: df.GEOID10.apply(lambda x: x[:-4]),
+                neighborhood_district = lambda df: df.NEIGHBORHOOD_DISTRICT_NAME.apply(lambda x: hierarchy_neighborhood(x))
+            )
+        )
+
+        hierarchies = puma_hierarchies.merge(neighborhood_hierarchies, on="feature_identifier", how="outer")
```
tsibley: I think something weird is happening in this join, but it might be elsewhere. The resulting CSV has lots of duplicate rows for Washington, e.g.:

image

locations/Snakefile
```diff
@@ -94,7 +171,58 @@ rule import_state:
...
+        geojson = ancient(rules.download_community_reporting_areas.output)
+    output:
+        geojson = "data/neighborhood_district/seattle_neighborhood_district.geojson"
+    shell:
+        """
+        ogr2ogr \
+            -f GeoJSON \
+            {output.geojson} {input.geojson} \
+            -dialect sqlite \
+            -sql "SELECT replace(replace(NEIGHDIST,' ','_'), '/', '_') as NEIGHDIST, ST_Union(geometry) as geometry FROM seattle_cra GROUP BY CRA_GRP"
```
tsibley: Hmm, this throws an error for me:

```
ERROR 1: In ExecuteSQL(): sqlite3_prepare_v2(SELECT replace(replace(NEIGHDIST,' ','_'), '/', '_') as NEIGHDIST, ST_Union(geometry) as geometry FROM seattle_cra GROUP BY CRA_GRP):
  no such table: seattle_cra
```

locations/Snakefile
joverlee521: This is probably due to the `SeattleCensusBlocksandNeighborhoodCorrelationFile.xlsx` file. The file contains all census **blocks** within Seattle, so there are repeated rows for census tracts.

locations/Snakefile
joverlee521: I think this might be version problem?
According to the GDAL [docs](https://gdal.org/drivers/vector/geojson.html): 
```
Starting with GDAL 2.2, the layer name is built with the following logic:
1. If a “name” member is found at the FeatureCollection level, it is used.
2. Otherwise if the filename is regular (ie not a URL with query parameters), then the filename without extension and path is used as the layer name.
3. Otherwise OGRGeoJSON is used.
```

tsibley:
Pushed a minor change to a pre-existing part of the Snakefile that I noticed while reviewing.

locations/Snakefile
tsibley: Hmm:

```
tom@whunk locations (pumas=) $ apt policy gdal-bin
gdal-bin:
  Installed: 2.2.3+dfsg-2
  …
```

joverlee521: Running `ogrinfo -ro data/cra/seattle_cra.geojson` gives this layer name for me:
```
INFO: Open of `data/cra/seattle_cra.geojson'
      using driver `ESRIJSON' successful.
1: seattle_cra (Polygon)
```

tsibley: Looks like the table name that works for me is `OGRGeoJSON`.

tsibley: Ah, which makes sense given the docs you quoted, since there is no `type: FeatureCollection` field in the input GeoJSON (making it not a valid GeoJSON!).

tsibley:
```
$ ogrinfo --version
GDAL 2.2.3, released 2017/11/20
$ ogrinfo -ro data/cra/seattle_cra.geojson
INFO: Open of `data/cra/seattle_cra.geojson'
      using driver `GeoJSON' successful.
1: OGRGeoJSON (Polygon)
```

tsibley: It looks like changing the download URL parameter from `f=json` to `f=geojson` gets us standard GeoJSON instead of ESRI's JSON format.

joverlee521:
```
$ ogrinfo --version
GDAL 2.4.2, released 2019/06/28
```

tsibley: If I download with `f=geojson`, then `ogrinfo` returns `seattle_cra` for me too.

kairstenfay:
When I test this locally, I get an error at either `download_state_tracts` or `download_state_tract_boundaries`.

The logging changes each time. 

[First attempt](https://github.com/seattleflu/id3c-customizations/files/4173288/snakemake.log0.txt)
* Job failed at `download_state_tracts_boundaries`. On the first attempt, I did not invoke `snakemake` with the `--max-jobs-per-second 0.5` option. For the following two attempts, I did use this option.

[Second attempt](https://github.com/seattleflu/id3c-customizations/files/4173292/snakemake.log1.txt)
* Job failed at `download_state_tracts`

[Third attempt](https://github.com/seattleflu/id3c-customizations/files/4173299/snakemake.log2.txt)
* Two jobs (?) failed at `download_state_tracts`. Unsure why this logging/output is different from the second attempt.

joverlee521:
@kairstenfay I was able to get it to run with `--max-jobs-per-second 0.05`

PR #39, opened 2020-01-31
Add swab_site and sample_origin to site details

kairstenfay:
Metrics calculations for year 2 have increasingly called for counts of
all samples including those without encounter data. To provide better
real-time estimate for sample collection whose encounter data we don't
yet have, map several values from the `swab_site` and `sample_origin`
columns of Peter's specimen manifest sheet to their `warehouse.site`
equivalents.

Note: the mapping from Peter's sheet to our warehouse site is not always
accurate, e.g. we can observe a `swab_site` of "suzzallo" mapped to the
`site.identifier` "HUB", etc.

Also note: some `site.identifier`s have multiple `swab_site`s that map
to them (see `ChildrensHospitalSeattle`), and some `swab_site`s appear
in multiple `site.identifier`s (see `community_clinic`).

---

The revert script doesn't work as expected. I'm thinking of either
* removing it
* writing line-by-line `UPDATE`s to remove the `sample_origin` or `swab_site` keys from affected rows. 

Postgres doesn't seem to support bulk updates like it does for inserts.


tsibley:
> The revert script doesn't work as expected. I'm thinking of either
> 
>     * removing it
> 
>     * writing line-by-line `UPDATE`s to remove the `sample_origin` or `swab_site` keys from affected rows.
> 
> 
> Postgres doesn't seem to support bulk updates like it does for inserts.

What do you mean by bulk updates here? It is possible, for example, to write something like:

```sql
update warehouse.site set details = details - '{sample_origin, swab_site}'::text[] where identifier in (…);
```

tsibley:
I think it would also be reasonable to decide that this sqitch change is our single source of site details and change the upsert's

```
set details = coalesce(site.details, '{}') || EXCLUDED.details
```

clause to 

```
set details = EXCLUDED.details
```

kairstenfay:
> I think it would also be reasonable to decide that this sqitch change is our single source of site details and change the upsert's
> 
> ```
> set details = coalesce(site.details, '{}') || EXCLUDED.details
> ```
> 
> clause to
> 
> ```
> set details = EXCLUDED.details
> ```

I chose to implement this suggestion in the deploy and revert scripts in sqitch.

kairstenfay:
Opened for visibility; closing in favor of #33

PR #40, opened 2020-02-04
Update api for new metadata view

kairstenfay:
~Question here: should I also delete `shipping.metadata_for_augur_build_v2`?~

~My guess is I should not before we deploy https://github.com/seattleflu/website/pull/28 and https://github.com/seattleflu/backoffice/pull/13 which both depend on this PR.~

We decided not to mark v2 as gone until after merging the dependent PRs.

### DIFF
--- a/lib/seattleflu/id3c/api/routes.py
+++ b/lib/seattleflu/id3c/api/routes.py
@@ -10,6 +10,9 @@ LOG = logging.getLogger(__name__)
 api_v2 = Blueprint('api_v2', 'api_v2', url_prefix='/v2')
 blueprints.append(api_v2)
 
+api_v3 = Blueprint('api_v3', 'api_v3', url_prefix='/v3')
+blueprints.append(api_v3)
+
 @api_v1.route("/shipping/return-results/<barcode>", methods = ['GET'])
 @cross_origin(origins=[
     "https://seattleflu.org",
@@ -42,6 +45,19 @@ def get_metadata_v2(session):
     LOG.debug("Exporting metadata for SFS augur build")
 
     metadata = datastore.fetch_rows_from_table(session, ("shipping", "metadata_for_augur_build_v2"))
+
+    return Response((row[0] + '\n' for row in metadata), mimetype="application/x-ndjson")
+
+
+@api_v3.route("/shipping/augur-build-metadata", methods = ['GET'])
+@authenticated_datastore_session_required
+def get_metadata_v3(session):
+    """
+    Export metadata needed for SFS augur build
+    """
+    LOG.debug("Exporting metadata for SFS augur build")
+
+    metadata = datastore.fetch_rows_from_table(session, ("shipping", "metadata_for_augur_build_v3"))
 
     return Response((row[0] + '\n' for row in metadata), mimetype="application/x-ndjson")
 
--- a/lib/seattleflu/id3c/api/routes.py
+++ b/lib/seattleflu/id3c/api/routes.py
@@ -41,7 +41,7 @@ def get_metadata_v2(session):
     """
     LOG.debug("Exporting metadata for SFS augur build")
 
-    metadata = datastore.fetch_metadata_for_augur_build(session)
+    metadata = datastore.fetch_rows_from_table(session, ("shipping", "metadata_for_augur_build_v2"))
 
     return Response((row[0] + '\n' for row in metadata), mimetype="application/x-ndjson")
 
--- a/lib/seattleflu/id3c/api/datastore.py
+++ b/lib/seattleflu/id3c/api/datastore.py
@@ -1,7 +1,33 @@
 from typing import Any, Iterable, Tuple
+from psycopg2.sql import SQL, Identifier
 from id3c.db.session import DatabaseSession
 from id3c.api.datastore import catch_permission_denied
 from id3c.api.utils import export
+
+@export
+@catch_permission_denied
+def fetch_rows_from_table(session: DatabaseSession,
+                          qualified_table: Tuple) -> Iterable[Tuple[str]]:
+    """
+    Exports all rows in a given *qualified_table* and yields them as JSON (one
+    per line) in a generative fashion.
+
+    *qualified_table* should be a tuple of (schema, table). All identifiers will
+    be properly quoted by this method.
+    """
+    assert len(qualified_table) == 2, \
+        "A schema and table name must be included in the qualified table tuple"
+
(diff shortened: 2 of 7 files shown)


kairstenfay:
[Update database roles for `metadata_for_augur_build_v2` view and new `metadata_for_augur_build_v3` view](https://trello.com/c/IaKFRlZq/332-update-database-roles-for-metadataforaugurbuildv2-view-and-new-metadataforaugurbuildv3-view)

kairstenfay:
[Create new API endpoint for `shipping.metadata_for_augur_build_v3`.](https://trello.com/c/Gz5ZVbzR/331-create-new-api-endpoint-for-shippingmetadataforaugurbuildv3)

lib/seattleflu/id3c/api/routes.py
```diff
@@ -35,7 +38,16 @@ def get_metadata_v1():
 
 @api_v2.route("/shipping/augur-build-metadata", methods = ['GET'])
 @authenticated_datastore_session_required
-def get_metadata_v2(session):
+def get_metadata_v2():
+    """
+    Mark old endpoint as 410 Gone
+    """
+    return "use /v3/shipping/augur-build-metadata instead", 410
```
kairstenfay: I believe we don't actually want to mark this endpoint as gone, but rather delete the v2 endpoint after we merge the related backoffice and website PR's (linked in top-level comment). 

Seeking feedback here.

lib/seattleflu/id3c/api/routes.py
```diff
@@ -35,7 +38,16 @@ def get_metadata_v1():
 
 @api_v2.route("/shipping/augur-build-metadata", methods = ['GET'])
 @authenticated_datastore_session_required
-def get_metadata_v2(session):
+def get_metadata_v2():
```
tsibley: `@authenticated_datastore_session_required` either needs to be removed or the `session` argument needs to be added back. I think the former as the hardcoded 410 Gone response doesn't need to be behind auth, and this is what the v1 route does.

In the current state, if you do authenticate, an exception is thrown instead of a 410 Gone because the decorator tries to pass the `session` argument but the function doesn't accept it.

```
[2020-02-04 10:58:24] Connected to user=test dbname=seattleflu port=5432 sslmode=prefer
[2020-02-04 10:58:24] Exception on /v2/shipping/augur-build-metadata [GET]
Traceback (most recent call last):
  File "/home/tom/.local/share/virtualenvs/id3c-customizations-Q1pOp1dl/lib/python3.6/site-packages/flask/app.py", line 2446, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/tom/.local/share/virtualenvs/id3c-customizations-Q1pOp1dl/lib/python3.6/site-packages/flask/app.py", line 1951, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/tom/.local/share/virtualenvs/id3c-customizations-Q1pOp1dl/lib/python3.6/site-packages/flask/app.py", line 1820, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/tom/.local/share/virtualenvs/id3c-customizations-Q1pOp1dl/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/tom/.local/share/virtualenvs/id3c-customizations-Q1pOp1dl/lib/python3.6/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/tom/.local/share/virtualenvs/id3c-customizations-Q1pOp1dl/lib/python3.6/site-packages/flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/tom/seattleflu/id3c/lib/id3c/api/utils/routes.py", line 37, in wrapped_route
    return route(*args, **kwargs, session = session)
TypeError: get_metadata_v2() got an unexpected keyword argument 'session'
```

lib/seattleflu/id3c/api/routes.py
tsibley: I think 410 Gone is fine and slightly more useful if someone runs an old saved command against those endpoints. But removing them after we think all uses are updated would also be ok, and in that case we'd want to remove both the v1 and v2 endpoints.

tsibley:
> Question here: should I also delete `shipping.metadata_for_augur_build_v2`?
> 
> My guess is I should not before we deploy [seattleflu/website#28](https://github.com/seattleflu/website/pull/28) and [seattleflu/backoffice#13](https://github.com/seattleflu/backoffice/pull/13) which both depend on this PR.

Yep, I agree.

kairstenfay:
> Well, with the caveat that we should check any Metabase queries using it too.

I created a [Trello card](https://trello.com/c/i5khEuyi) tracking this and the possibility of removing the v1 and v2 endpoints instead of marking them as gone.

lib/seattleflu/id3c/api/routes.py
kairstenfay: Ah, sorry @tsibley, I was unclear. 
I am ok with marking the v2 endpoint as "gone" but only after we merge the dependent PRs linked at the top. 
I should not have used the word "delete"

kairstenfay: Essentially, I am considering reverting this part of the commit to keep the v2 endpoint functioning for now.

tsibley: Ah, I see. Yep. Although those PRs at the top will require the `augur-build-exporter` role updates in this PR.

kairstenfay: Right, so we should merge this PR first. 
I don't believe the `augur-build-exporter` changes only add privileges, not revoke any.

kairstenfay: @tsibley In my latest fixed-up commit, the v2 endpoint is back to its original and is no longer marked as gone.

kairstenfay: Agh I just realized that doesn't work because the datastore function is updated to use v3. Unsure now how to proceed.

kairstenfay: Thomas and I discussed in person the idea to create a new, generic function that exports all rows from any qualified (schema, table) couple to ndjson. This exists as a new commit in this PR.

lib/seattleflu/id3c/api/routes.py
kairstenfay: Resolving this conversation as we are no longer marking the v2 endpoint as gone at this point.

PR #41, opened 2020-02-06
Add role to perform idle database disconnects

kairstenfay:
Create a new database role that has permissions to disconnect database
sessions. This role will eventually be used in a cron job on backoffice
for disconnecting sessions that have been idle for > 30 minutes.

### DIFF
--- a/schema/deploy/roles/idle-session-disconnector/grants.sql
+++ b/schema/deploy/roles/idle-session-disconnector/grants.sql
@@ -1,0 +1,44 @@
+-- Deploy seattleflu/id3c-customizations:roles/idle-session-disconnector/grants to pg
+
+begin;
+
+revoke all on database :"DBNAME" from "idle-session-disconnector";
+revoke all on schema receiving, warehouse, shipping from "idle-session-disconnector";
+revoke all on all tables in schema receiving, warehouse, shipping from "idle-session-disconnector";
+
+grant connect on database :"DBNAME" to "idle-session-disconnector";
+
+-- Regular users cannot see the state of sessions that they don't own. Create a
+-- new view using security definer with limited columns that this role can
+-- access to see a session's state.
+create or replace function public.pg_stat_get_activity_nonsuperuser() returns table(
+    pid integer, usename name, application_name text, client_addr inet,
+    state_change timestamp with time zone, state text) as
+    $$
+    select
+        pid, usename, application_name, client_addr, state_change, state
+        from pg_catalog.pg_stat_activity;
+    $$
+    language sql
+    volatile
+    security definer
+    set search_path = pg_catalog;
+
+revoke execute
+    on function public.pg_stat_get_activity_nonsuperuser
+  from public;
+
+grant execute
+    on function public.pg_stat_get_activity_nonsuperuser
+    to "idle-session-disconnector";
+
+create or replace view public.pg_stat_activity_nonsuperuser as
+    select * from public.pg_stat_get_activity_nonsuperuser();
+
+grant select
+    on public.pg_stat_activity_nonsuperuser
+    to "idle-session-disconnector";
+
+grant pg_signal_backend to "idle-session-disconnector";
+
+commit;
--- a/schema/verify/roles/idle-session-disconnector/grants.sql
+++ b/schema/verify/roles/idle-session-disconnector/grants.sql
@@ -1,0 +1,14 @@
+-- Verify seattleflu/id3c-customizations:roles/idle-session-disconnector/grants on pg
+
+begin;
+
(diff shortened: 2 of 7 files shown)


kairstenfay:
[Idle disconnect for database](https://trello.com/c/ygWRstQJ/322-idle-disconnect-for-database)

schema/deploy/roles/idle-session-disconnector/grants.sql
```diff
@@ -0,0 +1,33 @@
...
+revoke all on database :"DBNAME" from "idle-session-disconnector";
+revoke all on schema receiving, warehouse, shipping from "idle-session-disconnector";
+revoke all on all tables in schema receiving, warehouse, shipping from "idle-session-disconnector";
+
+grant connect on database :"DBNAME" to "idle-session-disconnector";
+
+-- Regular users cannot see the state of sessions that they don't own. Create a
+-- new view using security definer with limited columns that this role can
+-- access to see a session's state.
+create or replace function get_stat_activity() returns table(
```
tsibley: For our setup, it's generally good to explicitly specify a schema name for functions/tables/view/etc when creating them. This makes it very clear where they will exist. Otherwise, they're created in the first schema listed in `search_path`, which might be variable across deploys.

The `public` schema is an ok place to put this function, but note that functions in the `public` schema are by default executable by anyone (i.e. the pseudo-role also named `public`). From the Pg 10 docs on GRANT:

> PostgreSQL grants default privileges on some types of objects to PUBLIC. No privileges are granted to PUBLIC by default on tables, table columns, sequences, foreign data wrappers, foreign servers, large objects, schemas, or tablespaces. For other types of objects, the default privileges granted to PUBLIC are as follows: CONNECT and TEMPORARY (create temporary tables) privileges for databases; EXECUTE privilege for functions; and USAGE privilege for languages and data types (including domains). The object owner can, of course, REVOKE both default and expressly granted privileges. (For maximum security, issue the REVOKE in the same transaction that creates the object; then there is no window in which another user can use the object.) Also, these initial default privilege settings can be changed using the ALTER DEFAULT PRIVILEGES command.

Since this is a sensitive function (security definer), we'll want to revoke `execute` privileges from everyone but `idle-session-disconnector`.

schema/deploy/roles/idle-session-disconnector/grants.sql
```diff
@@ -0,0 +1,33 @@
...
+        pid, usename, application_name, client_addr, state_change, state
+        from pg_catalog.pg_stat_activity;
+    $$
+    language sql
+    volatile
+    security definer
+    set search_path = pg_catalog;
+
+create view pg_stat_activity_nonsuperuser as
+    select * from get_stat_activity();
```
tsibley: Similarly, would be good to explicitly name the schema the view should be created in.

I would also make the function and view names parallel the ones used by the `pg_stat_activity` view and its `pg_stat_get_activity` function, so renaming our function to `pg_stat_get_activity_nonsuperuser`.

schema/verify/roles/idle-session-disconnector/grants.sql
```diff
@@ -0,0 +1,12 @@
...
+
+begin;
+
+select 1/pg_catalog.has_database_privilege('idle-session-disconnector', :'DBNAME', 'connect')::int;
+select 1/pg_catalog.has_schema_privilege('idle-session-disconnector', 'public', 'usage')::int;
+select 1/pg_catalog.has_table_privilege('idle-session-disconnector', 'pg_stat_activity_nonsuperuser', 'select')::int;
+
+select 1/(not pg_catalog.has_table_privilege('idle-session-disconnector', 'pg_stat_activity', 'insert,update,delete'))::int;
+select 1/(not pg_catalog.has_table_privilege('idle-session-disconnector', 'pg_stat_activity_nonsuperuser', 'insert,update,delete'))::int;
+
```
tsibley: Would be good to test execute privs for the security-definer function, to verify that we got them right.

schema/deploy/roles/idle-session-disconnector/grants.sql
kairstenfay: Renamed

schema/deploy/roles/idle-session-disconnector/grants.sql
kairstenfay: I specified the public schema and revoked execute privileges from `public` before granting execute privileges to `idle-session-disconnector`.

schema/verify/roles/idle-session-disconnector/grants.sql
kairstenfay: Added a test to check that `idle-session-disconnector` has execute privileges but `public` does not.

PR #44, opened 2020-02-19
Distinguish sch sea mar from uw

kairstenfay:
Waiting on a response to [this slack message](https://seattle-flu-study.slack.com/archives/CPQHRGVEK/p1582071101004300) from Amanda or Kirsten for clarification. 

---
DO NOT merge this for now.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py
@@ -508,10 +508,13 @@ def determine_site_name(redcap_record: dict) -> Optional[str]:
 
     site = potential_site_names[0]
 
+    # Handle our special SeaMar edge case that relies on staff organization
+    if site == 'SeaMar':
+        return distinguish_seamar_affiliation(redcap_record)
+
     site_name_map = {
         'UW HUB': 'HUB',
         'UW Suzzallo Library': 'UWSuzzalloLibrary',
-        'SeaMar': 'UWSeaMar',
         'UW Hall Health': 'UWHallHealth',
         "Seattle Children's: Seattle Children's campus site": 'ChildrensHospitalSeattle',
         "Seattle Children's: Seattle outpatient clinic": 'ChildrensHospitalSeattle',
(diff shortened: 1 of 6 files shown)


kairstenfay:
[Y2 metrics](https://trello.com/c/ZpLpqy5G/260-y2-metrics)

tsibley:
> According to Robin, `home` should not be a site type. This is confusing the metrics team with household observation and intervention studies. Instead, Robin says to keep the site type for Swab N Send and Self-Test `swab-n-send` and `self-test` (or something similar), respectively.

I worry this change is misguided on Robin's part, particularly because her usage for metrics isn't the only usage of site type by study teams. For example, other analyses in the study may want to lump together this sort of "participation from home".

It probably isn't worth following up about, but I would not be surprised if we end up having to change this back at some point.

tsibley:
(Noting also that Robin is not the person who originally defined these types and categories. They came out of a lengthy back-and-forth between Helen and Mike's needs.)

kairstenfay:
> > According to Robin, `home` should not be a site type. This is confusing the metrics team with household observation and intervention studies. Instead, Robin says to keep the site type for Swab N Send and Self-Test `swab-n-send` and `self-test` (or something similar), respectively.
> 
> I worry this change is misguided on Robin's part, particularly because her usage for metrics isn't the only usage of site type by study teams. For example, other analyses in the study may want to lump together this sort of "participation from home".
> 
> It probably isn't worth following up about, but I would not be surprised if we end up having to change this back at some point.

I see. Changing them back to `home` eventually is OK with me if it means metrics get done more quickly and smoothly for now. 

Also while you were out, I cherry-picked this commit and merged #50 which converts the site_type of Self-Test and Swab N Send to their eponymous site type labels. 

I'm going to hold off on the other commits in this PR for now. I think we'll end up dropping the SCH/UW SeaMar distinction.

tsibley:
Ah, I was working through notifications in order and didn't see the later PR which was merged. Sounds good. Thanks for managing all this!

kairstenfay:
> Ah, I was working through notifications in order and didn't see the later PR which was merged. Sounds good. Thanks for managing all this!

No worries. I do not expect my sometimes anachronistic commits to be intuitive to all.

kairstenfay:
Adding a new site for SCH SeaMar is no longer going to happen.

PR #45, opened 2020-02-19
Update site data

kairstenfay:
I met with Robin, Kirsten and Amanda a couple weeks ago. They requested some changes to the way we currently calculate site metrics. These changes reflect:
* Creation of new site, `SCHSeaMar`, that will be added to FHIR documents by the kiosk ETL once #44 is merged. 
* Stopping the usage of site type `home` for swab-n-send and self-test in favor of eponymous site types

It's worth investigating why SCH doesn't want to use the `ChildrensSeaMar` site from Year 1. They claim it was something Audere created, and I think it might be a different site entirely.


kairstenfay:
[Y2 metrics](https://trello.com/c/ZpLpqy5G/260-y2-metrics)

kairstenfay:
> Looks like #44 includes these commits already?

Yep, not sure what I was doing here and how #44 ended up including these changes. I think I forgot to delete this branch/PR after some more consideration. It's also worth noting that I've already merged the `self-test` and `swab-n-send` site_type changes (via #50 which I cherrypicked from either #44 or this PR).

PR #48, opened 2020-02-20
Link multi organism targets to generic parent organism

joverlee521:
This is just a temporary solution for return of results to include targets that are linked to multiple organism subtypes. 

In the future we will want to be more specific and link these targets to multiple organisms.


joverlee521:
[Link multi-subtype targets to generic parent organism](https://trello.com/c/4CDkK0jj/349-link-multi-subtype-targets-to-generic-parent-organism)

joverlee521:
Rebased on to master to resolve conflicts in sqitch.plan

PR #49, opened 2020-02-21
New Swab & Send + Home Flu

joverlee521:
I copied over the ETL process for the old swab-n-send project and edited it for the new project. 
(The old project is still live and has data trickling in from people who signed up over the last two weeks.)

The major change is how to determine site for the encounter, which now depends on the records `study_type`

The two studies share a lot of the same field names so a majority of the ETL process did not need to be changed. In cases where the same data is provided under different field names, I just used the `or` operator in the variable assignment.


joverlee521:
[Update REDCap swab-n-send ETL to new project](https://trello.com/c/6f0OaALq/304-update-redcap-swab-n-send-etl-to-new-project)

PR #51, opened 2020-02-26
Return only Cepheid results

joverlee521:
This PR depends on https://github.com/seattleflu/id3c/pull/119

Limit `shipping.return_results_v1` to only Cepheid results and also return Y2 self-test results. 

Limits results to Influenza and RSV and separates `Influenza.A` and `Influenza.B` results


joverlee521:
Ah, I forgot the distinction of Flu A vs Flu B.

joverlee521:
[Scope return of results view to just Cepheid results](https://trello.com/c/eZeY6sk2/347-scope-return-of-results-view-to-just-cepheid-results)

schema/deploy/shipping/views.sql
```diff
@@ -139,26 +139,37 @@ create or replace view shipping.return_results_v1 as
...
                                   'collections-swab&send',
+                                  'collections-self-test',
                                   'collections-household-observation',
                                   'collections-household-intervention') and
           (organism is null or
           -- We only return results for these organisms, so omit all other presence/absence results
-          organism <@ '{"Adenovirus", "Human_coronavirus", "Enterovirus", "Influenza", "Human_metapneumovirus", "Human_parainfluenza", "Rhinovirus", "RSV"}'::ltree[])
+          organism <@ '{"Influenza.A", "Influenza.B", "RSV"}'::ltree[]) and
+          -- Limit to only Cepheid results
+          presence_absence_result_v2.details ->> 'device' = 'Cepheid'
```
tsibley: It's a small thing, but may make a difference: `details @> '{"device":"Cepheid"}'` can use the index on `details`, but `details->>'device' = 'Cepheid'` cannot. Details in the doc: https://www.postgresql.org/docs/10/datatype-json.html#JSON-INDEXING

schema/deploy/shipping/views.sql
```diff
@@ -139,26 +139,37 @@ create or replace view shipping.return_results_v1 as
...
+              when 'Influenza'::ltree @> organism then subpath(organism, 0, 2)::text
+              else subpath(organism, 0, 1)::text
+             end)
+           ) filter (where present and organism is not null)  as organisms_present,
+           array_agg(distinct
+             (case
+              when 'Influenza'::ltree @> organism then subpath(organism, 0, 2)::text
+              else subpath(organism, 0, 1)::text
+             end)
+           ) filter (where not present and organism is not null)  as organisms_absent
```
tsibley: The `subpath` stuff isn't needed and this can be much simpler since we're only selecting results for `Influenza.A`, `Influenza.B`, and `RSV` now: `array_agg(distinct organism::text)`.

schema/deploy/shipping/views.sql
tsibley: I believe this condition will also negate the `left join` to presence/absence results and filter out rows which would have status of `notReceived` or `processing`.

tsibley:
> Looks good. I only wonder if these changes are enough to warrant a version bump to `return_results_v2`. I don't have strong feelings on this, though.

My thought is that it doesn't warrant a bump because a) the view's interface ("API", e.g. schema) is not changing, just the data returned by it, and b) we should be the only consumers of this view so signaling a big difference in data isn't necessary.

schema/deploy/shipping/views.sql
joverlee521: I thought it would still include the subtypes of `Influenza.A` and `Influenza.B` without the subpath stuff. 

Although all Cepheid results don't include subtypes, so it might be overly complicated.

tsibley: Ahh, yeah, i was thinking only about the Cepheid results, but now recall we do have those subtypes.

You're definitely correct that the code will still guard against Cepheid being more specific (or us otherwise introducing results that are more specific). I'm not sure that'll ever happen, but no problem being defensive. No need to change this then.

joverlee521: I can also change to the where clause to `organism ? '{"Influenza.A", "Influenza.B", "RSV"}'::lquery[]` for exact matches only

tsibley: Nod. I'm ok with either way.

joverlee521:
Rebased on to master to resolve merge conflict and tagged schema

kairstenfay:
> This PR also looks good to me, so let's merge it now to build on top of it for COVID-19 work.

Err, merge it once merge conflicts are resolved.

joverlee521:
> > This PR also looks good to me, so let's merge it now to build on top of it for COVID-19 work.
> 
> Err, merge it once merge conflicts are resolved.

I thought we were putting a hold on returning Cepheid results? Or did I miss something?

kairstenfay:
> > > This PR also looks good to me, so let's merge it now to build on top of it for COVID-19 work.
> > 
> > 
> > Err, merge it once merge conflicts are resolved.
> 
> I thought we were putting a hold on returning Cepheid results? Or did I miss something?

kairstenfay:
Oh snap I didn't mean to close, just canceled my comment.

tsibley:
> I thought we were putting a hold on returning Cepheid results? Or did I miss something?

Only a hold because of the COVID-19 response is taking priority. I think this would be good to merge now even if we don't yet do the corresponding website work to make use of it.

schema/deploy/shipping/views.sql
```diff
@@ -138,27 +138,27 @@ create or replace view shipping.return_results_v1 as
...
                                   'collections-swab&send',
+                                  'collections-self-test',
                                   'collections-household-observation',
                                   'collections-household-intervention') and
           (organism is null or
           -- We only return results for these organisms, so omit all other presence/absence results
-          organism <@ '{"Adenovirus", "Human_coronavirus", "Enterovirus", "Influenza", "Human_metapneumovirus", "Human_parainfluenza", "Rhinovirus", "RSV"}'::ltree[])
+          organism ? '{"Influenza.A", "Influenza.B", "RSV"}'::lquery[]) and
+          -- Limit to only Cepheid results
+          (presence_absence_result_v2.details is null or presence_absence_result_v2.details @> '{"device":"Cepheid"}')
```
tsibley: It may not matter in practice, but there might be non-Cepheid results with null details and they would be included by this condition. To detect the case of no matching row for the left join, better to condition on the PK or equivalent of the joined table, e.g. `presence_absence_result_v2.sample is null`.

tsibley:
@joverlee521 Seems like we could rebase and merge this to clean up open PRs a little? Is there a reason not to?

tsibley:
Nod. Sounds good to me. Closing.

PR #54, opened 2020-02-27
etl redcap-det kiosk: Update Cepheid results

joverlee521:
Makes the handling of Cepheid results less convoluted.

Only include "Inconclusive" result observation when the test is inconclusive and do not include pathogen results if "Inconclusive".


joverlee521:
@seattleflu/core-devs I believe this PR can be closed since we are no longer using Cepheid.

PR #59, opened 2020-03-05
Create new RoR view for COVID-19

kairstenfay:
Create a new view, `shipping.return_results_v2`, for returning COVID-19/
HCoV-19 results to participants. This view differs from v1 by returning:
* only `Human_coronavirus.2019` results
* results only after Izzy updated the Swab & Send / Home Test REDCap
  consent form using timestamp "2020-03-04 08:35:00-8"
* organism subtype to distinguish clearly from other Human coronavirus
* only results from the `collections-kiosks` and `collections-swab&send`
  identifier sets

----

This PR does not reflect the Cepheid work that @joverlee521 started (https://github.com/seattleflu/id3c-customizations/pull/51), because as far as I know, the organisms we will / will not be able to return and through which devices they must be returned are still in flux. 

Therefore, this view only returns results for `Human_coronavirus.2019`. 

Here, I'm using the timestamp for Izzy's consent change on REDCap to the Swab & Send ./ Home Test project, but I'm also including samples from the Kiosk project. My question here is: do I drop kiosk samples for now or wait for a potential consent form update to this project as well, applying the timestamp before merging?


kairstenfay:
[Return of COVID-19 results](https://trello.com/c/ELYj9ohx/361-return-of-covid-19-results)

schema/deploy/shipping/views.sql
```diff
@@ -669,4 +669,39 @@ comment on view shipping.sample_with_best_available_encounter_data_v1 is
...
+
+      from warehouse.identifier
+      join warehouse.identifier_set using (identifier_set_id)
+      left join warehouse.sample on uuid::text = sample.collection_identifier
+      left join warehouse.encounter using (encounter_id)
+      left join shipping.presence_absence_result_v2 on sample.identifier = presence_absence_result_v2.sample
+
+    --These are all past and current collection identifier sets not including self-test
+    where identifier_set.name in ('collections-kiosks',
+                                  'collections-swab&send')
```
tsibley: I think this should exclude kiosks (which for sure aren't running and it also seems don't have updated consent) and include `collections-self-test`, which shares the same REDCap project as swab and send now.

schema/deploy/shipping/views.sql
```diff
@@ -669,4 +669,39 @@ comment on view shipping.sample_with_best_available_encounter_data_v1 is
...
+      and (organism is null or
+          -- We only return results for COVID-19, so omit all other presence/absence results
+          organism <@ '{"Human_coronavirus.2019"}'::ltree[])
+          -- We only want results collected after Izzy updated the REDCap consent form in swab & send.
+          -- This filter-by timestamp comes from REDCap
+      and coalesce(encountered, date_or_null(warehouse.sample.details->>'date')) > '2020-03-04 08:35:00-8'::date
```
tsibley: The cast of the timestamp to `::date` means that the time part is ignored (i.e. treated as 00:00:00). Changing to `::timestamp with time zone` or using `2020-03-05` (one day later) would both be acceptable.

schema/deploy/shipping/views.sql
kairstenfay: Perhaps swab & send asymptomatic should also be added, too?

schema/deploy/shipping/views.sql
kairstenfay: Resolved

schema/deploy/shipping/views.sql
tsibley: If the consent was updated there at the same time, then yes.

kairstenfay: It doesn't appear to be updated, yet. So I will leave it off for now.

kairstenfay:
Just pushed a reword commit.

kairstenfay:
@joverlee521 @tsibley I pushed a new commit with relevant changes to the API including deprecating the old v1 `return-results` endpoint.

kairstenfay:
Also I updated the "return-results-exporter" grants via a rework to account for `shipping.return_results_v2`.

lib/seattleflu/id3c/api/routes.py
```diff
@@ -14,12 +14,20 @@
 blueprints.append(api_v3)
 
 @api_v1.route("/shipping/return-results/<barcode>", methods = ['GET'])
+def get_barcode_results_v1(barcode=None):
```
joverlee521: Not sure we want to mark v1 as gone, we will probably use it for Cepheid results in the future?

kairstenfay: I assumed we would integrate Cepheid results into the v2 view.

joverlee521: I was imagining the normal Flu/RSV results would be on a completely separate page than HCoV-19 results 🤷‍♀

kairstenfay: Oh interesting. I was thinking they'd be on the same page and we would have a `union` query joining Cepheid results and HCoV-19 results.

tsibley: Yeah, I think it's just unclear how either set of results will ultimately be returned, and this iss reflected in being unclear on what the final implementation might look like. Doing this view + API work now means we more quickly return them when decisions are made, though. I think marking the v1 endpoint 410 Gone is fine for now. We can always revert it.

lib/seattleflu/id3c/api/routes.py
```diff
@@ -14,12 +14,21 @@
...
+    """
+    return "use /v2/shipping/return-results/<barcode>", 410
+
+
+@api_v2.route("/shipping/return-results/<barcode>", methods = ['GET'])
 @cross_origin(origins=[
     "https://seattleflu.org",
     "https://dev.seattleflu.org",
+    "http://localhost:3000",
     "http://localhost:8080"])
```
tsibley: By "moving" the `@cross_origin` decorator from v1 → v2 instead of copying it, the v1 endpoint won't actually ever get a chance to respond 410 Gone to a cross-origin request from the website frontend. It'll be blocked by CORS first.

This doesn't really matter, as the end result is similar (the request still fails), but I wanted to point out this nuance as I don't think it's what you intended.

kairstenfay: That makes sense. I'll duplicate the `@cross_origin` decorator across v1 and v2.

lib/seattleflu/id3c/api/routes.py
kairstenfay: Ok, I'm going to mark v1 as gone for now since we aren't supposed to be currently using it anyway. I'm happy to revert this down the road as soon as we have more clarity.

PR #60, opened 2020-03-06
Asymptomatic swab n send

joverlee521:
Very similar to old swab-n-send project, just had to edit some field names and delete fields that are not present in the asymptomatic project.


joverlee521:
[Ingest asymptomatic swab & send records](https://trello.com/c/1ksfbtGL/365-ingest-asymptomatic-swab-send-records)

PR #65, opened 2020-03-09
Report hCoV-19

joverlee521:
Report hCoV-19 in a separate channel and provide a separate metabase url for hCoV-19 specific results.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/reportable_conditions.py
+++ b/lib/seattleflu/id3c/cli/command/reportable_conditions.py
@@ -67,6 +67,7 @@ def notify(*, action: str):
 
     SLACK_WEBHOOK_REPORTING_GENERAL = webhook("GENERAL")
     SLACK_WEBHOOK_REPORTING_CHILDRENS = webhook("CHILDRENS")
+    SLACK_WEBHOOK_REPORTING_HCOV19 = webhook("HCOV19")
 
     childrens_sites = get_childrens_sites(db)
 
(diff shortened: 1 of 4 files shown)


joverlee521:
[HCoV-19 alerts](https://trello.com/c/9zW4fLMd/363-hcov-19-alerts)

joverlee521:
> The changes look good to me, but I get this error when I try to sqitch deploy to my local dev instance or id3c-testing:
> 
> ```
> Deploying changes to dev
>   + warehouse/target/data @2020-03-09b .. Verify script schema/verify/warehouse/target/data.sql does not exist
> ERROR:  duplicate key value violates unique constraint "changes_project_script_hash_key"
> DETAIL:  Key (project, script_hash)=(seattleflu/id3c-customizations, 743893add0978aaf8323c138197bd093f836a84b) already exists.
> schema/revert/warehouse/target/data.sql: No such file or directory
> "psql" unexpectedly returned exit value 1
> 
> not ok
> Deploy failed
> ```
> 
> Not sure if this is actually a Sqitch bug or what. :-/

Ah, it's because `data.sql` is the exact same as `data@2020-03-07.sql` so it thinks that the change already exists...the deploy works if I add an extra newline in the file....

tsibley:
> Ah, it's because `data.sql` is the exact same as `data@2020-03-07.sql` so it thinks that the change already exists...the deploy works if I add an extra newline in the file....

This strikes me as pretty weird behaviour from Sqitch. Workaround is good for now, but I'd love to eventually track down why Sqitch cares here.

tsibley:
Ah, I see: https://github.com/sqitchers/sqitch/issues/367

PR #66, opened 2020-03-10
Add hCoV-19 observation w/ presence-absence view

kairstenfay:
Add a new similar to `observation_with_presence_absence_result` specific
to samples tested for hCoV-19.

Currently this view includes all non-control samples, with or without
encounter data, that have been tested for hCoV-19 and all encounters
since March 1, 2020.

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -757,11 +757,11 @@ create or replace view shipping.hcov19_observation_v1 as
     left join warehouse.location using (location_id)
     where
         lineage = 'Human_coronavirus.2019'
-        -- If no test results are available, select encounters on or after 1 March 2020
-        -- Because Mike said so
+        -- If no test results are available, select encounters on or after 22 Feb 2020.
+        -- Date chosen because of this tweet: https://mobile.twitter.com/trvrb/status/1237395936193605633
         or (
           lineage is null and
-          best_available_encounter_date >= '2020-03-01'::date
+          best_available_encounter_date >= '2020-02-22'::date
         )
         and (not control or control is null)
         and sample.details->>'sample_origin' != 'es'
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -730,42 +730,89 @@ comment on view shipping.return_results_v2 is
 
 create or replace view shipping.hcov19_observation_v1 as
 
-    select
-        presence_absence.created::date as sample_tested_date,
-        present,
-        sample.details->>'sample_origin' as sample_manifest_origin,
-        best_available_encounter_date as encounter_date,
-        age_bin_fine.range as age_range_fine,
-        lower(age_bin_fine.range) as age_range_fine_lower,
-        upper(age_bin_fine.range) as age_range_fine_upper,
-        sex,
-        hierarchy->'puma' as puma,
-        best_available_site as site,
-        individual,
-        sample.identifier as sample,
-        best_available_site_type as site_type
+    with hcov19_presence_absence as (
+        -- Collapse potentially multiple hCoV-19 results
+        select distinct on (sample_id)
+            sample_id,
+            pa.created::date as hcov19_result_received,
+            pa.present as hcov19_present
+        from
+            warehouse.presence_absence as pa
+            join warehouse.target using (target_id)
+            join warehouse.organism using (organism_id)
+        where
+            organism.lineage <@ 'Human_coronavirus.2019'
+            and not control
+        order by
+            sample_id,
+            present desc nulls last -- t → f → null
+    )
 
-    from shipping.sample_with_best_available_encounter_data_v1
-    join warehouse.sample using (sample_id)
-    full outer join warehouse.encounter using (encounter_id)
-    left join warehouse.individual using (individual_id)
-    left join shipping.age_bin_fine on age_bin_fine.range @> ceiling(age_in_years(age))::int
-    left join warehouse.presence_absence using (sample_id)
-    left join warehouse.target using (target_id)
-    left join warehouse.organism using (organism_id)
-    left join warehouse.encounter_location using (encounter_id)
-    left join warehouse.location using (location_id)
+    select
+        sample_id,
+        sample.identifier as sample,
+
+        -- Lab testing-related columns
+        hcov19_result_received,
+        hcov19_present,
+
+        -- Encounter-related columns
+        encounter_id,
+        best_available_encounter_date as encountered,
+        to_char(best_available_encounter_date, 'IYYY-"W"IW') as encountered_week,
+
+        best_available_site as site,
+        best_available_site_type as site_type,
+
+        location.hierarchy->'puma' as puma,
+
+        -- Individual-related columns
+        individual_id,
+        age_bin_fine_v2.range as age_range_fine,
+        lower(age_bin_fine_v2.range) as age_range_fine_lower,
+        upper(age_bin_fine_v2.range) as age_range_fine_upper,
+        sex,
+
+        -- Misc cruft
+        sample.details->>'sample_origin' as manifest_origin
+
+        /* XXX TODO
+         *   → symptoms (Mike says can be JSON blob)
+         *   → symptom onset (date)
+         *   → race, SES, housing, etc
+         */
+    from
+        warehouse.sample
+        left join shipping.sample_with_best_available_encounter_data_v1 using (sample_id)
+        left join warehouse.encounter using (encounter_id)
+        left join warehouse.individual using (individual_id)
+        left join shipping.age_bin_fine_v2 on age_bin_fine_v2.range @> age
+        left join warehouse.primary_encounter_location using (encounter_id)
+        left join warehouse.location using (location_id)
+        left join hcov19_presence_absence using (sample_id)
     where
-        lineage = 'Human_coronavirus.2019'
-        -- If no test results are available, select encounters on or after 22 Feb 2020.
-        -- Date chosen because of this tweet: https://mobile.twitter.com/trvrb/status/1237395936193605633
-        or (
-          lineage is null and
-          best_available_encounter_date >= '2020-02-22'::date
-        )
-        and (not control or control is null)
-        and sample.details->>'sample_origin' != 'es'
-    ;
+        /* All tested samples plus samples and encounters after 22 Feb 2020, as
+         * we presume those _will be_ tested.  This criteria comes from Mike
+         * Famulare:
+         *
+         *   https://seattle-flu-study.slack.com/archives/GU24NGD18/p1583684258051000
+         *
+         * Note that when comparing some row-valued X, the expressions "X is
+         * not null" and "X is distinct from null" behave differently.  We want
+         * the latter.
+         */
+        (hcov19_presence_absence is distinct from null or best_available_encounter_date >= '2020-02-22')
+
+        /* Exclude environmental swabs.
+         *
+         * Note that the standard index on details which supports containment
+         * (@>) can't/won't be used by the planner because of both the negated
+         * condition ("not … @>") and coupled "is null" check.  If this
+         * condition is slow, we could directly index
+         * details->>'sample_origin'.
+         */
+        and (sample.details is null or sample.details->>'sample_origin' != 'es')
+;
 
 
 /* The shipping.hcov19_observation_v1 view needs hCoV-19 visibility, so
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -816,8 +816,8 @@ create or replace view shipping.hcov19_observation_v1 as
 
 
 /* The shipping.hcov19_observation_v1 view needs hCoV-19 visibility, so
- * remains owned by postgres, but it should only be accessible by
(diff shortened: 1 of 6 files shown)


schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,53 @@ comment on view shipping.return_results_v2 is
...
+      )
+      and sample.details->>'sample_origin' != 'es'
+      and not control
+    ;
+
+
+comment on view shipping.hcov19_observation_with_presence_absence_result_v1 is
+  'Custom view similar to shipping.observation_with_presence_absence_result_v2 but for hCoV-19 positive samples';
+
+
```
kairstenfay: Not sure what kind of grants I should assign here given the new RLS. 
Maybe this?
```
revoke all
    on shipping.hcov19_observation_with_presence_absence_result_v1
  from "hcov19-visibility";

grant select
   on shipping.hcov19_observation_with_presence_absence_result_v1
   to "hcov19-visibility";
```

kairstenfay: Or perhaps a revoke all from public?

kairstenfay:
Just pushed a reword fixup.

kairstenfay:
[Data export for Mike & PH](https://trello.com/c/2x9cYe7R/370-data-export-for-mike-ph)

schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,53 @@ comment on view shipping.return_results_v2 is
...
+    -- TODO we'll replace this via a replacement join on
+    -- incidence_model_observation_v4 with PUMAs once it's merged to master
+    left join (
+      select
+        encounter_id,
+        tract.hierarchy->'puma' as residence_puma,
+        tract.hierarchy->'neighborhood_district' as residence_neighborhood_district
+      from warehouse.encounter_location
+      left join warehouse.location as address using (location_id)
+      left join warehouse.tract as tract on (address.hierarchy -> 'tract') = tract.identifier
```
tsibley: This join is weird to me. The point of the location.hierarchy is to avoid this kind of double join to location. For any location row at the scale of puma or below, the containing puma should be directly accessible in `hierarchy->'puma'`.

It does look like some records are indeed missing puma in hierarchy and have tract, but I consider this a data loading bug we should (and can) fix.

Note that there's also no guarantee that an encounter location points to an address, so the table alias is slightly misleading.

schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,53 @@ comment on view shipping.return_results_v2 is
...
+        best_available_site,
+        individual,
+        sample.identifier as sample,
+        symptoms,
+        best_available_site_type
+
+    from shipping.sample_with_best_available_encounter_data_v1
+    join warehouse.sample using (sample_id)
+    full outer join shipping.incidence_model_observation_v3 on
+      (sample.identifier = incidence_model_observation_v3.sample)
```
tsibley: What columns ultimately come from the `incidence_model_observation_v3` view joined here? I'd guess it would be cleaner to grab them from core fact tables than re-use that view. Too many cross-view dependencies.

schema/deploy/shipping/views.sql
tsibley: This view will need the same grants/revokes as the reportable conditions view.

schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,53 @@ comment on view shipping.return_results_v2 is
     'Version 2 of view of barcodes and presence/absence results for return of results on website';
 
 
+create or replace view shipping.hcov19_observation_with_presence_absence_result_v1 as
+
+select
+        presence_absence.created::date as sample_tested_date,
+        present as test_result,
+        sample.details->>'sample_origin' as sample_manifest_origin,
+        best_available_encounter_date,
```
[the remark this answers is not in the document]
kairstenfay: Yes, I agree it is a mouthful, but I just wanted to underscore that it could contain some unvalidated (i.e. specimen manifest) data.

schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,53 @@ comment on view shipping.return_results_v2 is
     'Version 2 of view of barcodes and presence/absence results for return of results on website';
 
 
+create or replace view shipping.hcov19_observation_with_presence_absence_result_v1 as
```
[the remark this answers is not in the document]
kairstenfay: > What do you think about making this name a bit simpler? `shipping.hcov19_observation` seems like a nicer alternative.

That sounds great to me! I was trying to be consistent with the `observation_` view naming. 

>     * Because of those things, I would really like to stop using the `shipping.observation_with_presence_absence_result` views for new things moving forward. I don't think it's a well-designed query (it doesn't de-duplicate p/a results, for example), and I realize that's 100% my own doing.

Good to know! I didn't realize you wanted to ditch this view. 

>     * Finally, the columns and data in this view don't seem to me to bear much resemblance to `shipping.observation_with_presence_absence_result`.

Fair enough. 

I'm happy to drop all mention of `observation_with_presence_absence_result`.

schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,53 @@ comment on view shipping.return_results_v2 is
...
+
+select
+        presence_absence.created::date as sample_tested_date,
+        present as test_result,
```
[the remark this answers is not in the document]
kairstenfay: Makes sense to me.

schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,53 @@ comment on view shipping.return_results_v2 is
...
+        sample.details->>'sample_origin' as sample_manifest_origin,
+        best_available_encounter_date,
+        age_range_fine,
```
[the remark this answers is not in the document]
kairstenfay: Added it.

schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,53 @@ comment on view shipping.return_results_v2 is
...
+        sex,
+        residence_puma,
+        best_available_site,
+        individual,
+        sample.identifier as sample,
+        symptoms,
+        best_available_site_type
```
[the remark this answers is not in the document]
kairstenfay: Yep. I documented in the comment that this view includes the best available encounter data, so hopefully that message is sufficient.

schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,53 @@ comment on view shipping.return_results_v2 is
...
+        tract.hierarchy->'neighborhood_district' as residence_neighborhood_district
+      from warehouse.encounter_location
+      left join warehouse.location as address using (location_id)
+      left join warehouse.tract as tract on (address.hierarchy -> 'tract') = tract.identifier
+      where relation = 'residence'
+      or relation = 'lodging'
+    ) as residence using (encounter_id)
+    where (
+        lineage ~ 'Human_coronavirus.2019'
+        or best_available_encounter_date >= '2020-03-01'::date
```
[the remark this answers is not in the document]
kairstenfay: > Hmm, won't this return a row for every non-hCoV-19 presence absence result too? I might be missing something, but I think this filter isn't doing what we want.

Oh, derp. I'll fix that.

> The `~ [lquery]` comparison is equivalent to `= [ltree]` when the `lquery` contains no wildcards, but the latter seems more clear. I had to go lookup what `~` was doing here, but everyone knows what `=` means.

Thanks for the info. I found the `~` operator on the [table in these Postgres docs](https://www.postgresql.org/docs/10/ltree.html) but I missed the paragraph at the top saying `=` was also available. 

> It would also be good to document with a comment why this magic date value is what it is.

Yes, good point. As far as I know it is what it is because Mike said so. But I'll investigate and add a comment. :smile:

schema/deploy/shipping/views.sql
kairstenfay: Not too many columns. I thought it would make for cleaner `age_range` selection, but I agree that reducing cross-view dependencies is probably a good thing.

kairstenfay: The list of columns from `incidence_model_observation_v3` is:
* `age_range_fine`
* `age_range_fine_upper`
* `age_range_fine_upper`
* `sex`
* `symptoms`

schema/deploy/shipping/views.sql
kairstenfay: Got it. I'm using `hierarchy->'puma'` now instead of this join.

kairstenfay:
@tsibley I re-pushed with fixups from your requested changes.

I dropped `symptoms` from the columns for now (which should be OK given that Mike listed this as a "nice to have" initially). Now that we're no longer using `shipping.incidence_model_observation_v3` as a dependency, I have some questions about how to best bring in these data that would be nice to address over Slack or on a call tomorrow.

schema/deploy/shipping/views.sql
```diff
@@ -768,6 +768,27 @@ create or replace view shipping.hcov19_observation_v1 as
     ;
 
 
+/* The shipping.hcov19_observation_v1 view needs hCoV-19 visibility, so
+ * remains owned by postgres, but it should only be accessible by
+ * reportable-condition-notifier.  Revoke existing grants to every other role.
```
kairstenfay: I copied this comment from the `return_results` view. I don't think it's correct as is. Should `reportable-condition-notifier` have access to this view? By revoking access to `reporter`, don't we leave `fhir-processor` and `presence-absence-processor` with access, too? I'd like to discuss this with you tomorrow if possible.

schema/deploy/shipping/views.sql
```diff
@@ -728,4 +728,48 @@ comment on view shipping.return_results_v2 is
...
+    left join shipping.age_bin_fine on age_bin_fine.range @> ceiling(age_in_years(age))::int
+    left join warehouse.presence_absence using (sample_id)
+    left join warehouse.target using (target_id)
+    left join warehouse.organism using (organism_id)
+    left join warehouse.encounter_location using (encounter_id)
+    left join warehouse.location using (location_id)
+    where
+        lineage = 'Human_coronavirus.2019'
+        -- If no test results are available, select encounters on or after 1 March 2020
+        -- Because Mike said so
```
kairstenfay: This is a placeholder for a real answer from Mike.

schema/deploy/shipping/views.sql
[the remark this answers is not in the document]
tsibley: > Good to know! I didn't realize you wanted to ditch this view.

Yeah, and to be clear, totally didn't expect you to. :-) Just something I've been ruminating on, but that's been super low priority.

tsibley:
@kairstenfay Thanks for all this work. I'm still struggling to reason to myself that this query is actually returning what we want. I'm pretty sure it's not because of the way the joins and where conditions interact. Some samples are definitely duplicated, for example, and the where conditions assume all targets have an organism (and the filter doesn't do what we want if that's not true). I think it'd help (at least me) to restructure this query to be easier to reason about. I'm going to try and do that both to convince myself that we're doing what we intend to and to figure out concrete changes to suggest.

tsibley:
Pushed a commit with my suggested changes after working out a bunch of problematic joins and conditions. Lots of details in the commit message. I abstracted out a generic piece that seemed useful to have in core as https://github.com/seattleflu/id3c/pull/131, which this PR now depends on.

I would definitely appreciate feedback on this!

schema/deploy/shipping/views.sql
tsibley: See db7f9df, which I just pushed to correct the comment and add a grant for hcov19-visibility.

kairstenfay:
> Pushed a commit with my suggested changes after working out a bunch of problematic joins and conditions. Lots of details in the commit message. I abstracted out a generic piece that seemed useful to have in core as [seattleflu/id3c#131](https://github.com/seattleflu/id3c/pull/131), which this PR now depends on.
> 
> I would definitely appreciate feedback on this!

Thanks @tsibley. Using a CTE makes this view so much easier to understand and see potential pitfalls. 

I took a look at it locally. It looks good. The only issue I've encountered is that a `select *` on the view is extremely slow. I'm curious if you experienced this, too. 

Also, would now be a good time to add in those extra variables Mike requested? e.g. symptoms or race

kairstenfay:
@tsibley It looks like the sqitch tag also got out of order.

tsibley:
> It looks like the sqitch tag also got out of order.

Hmm, how so? 

> Also, would now be a good time to add in those extra variables Mike requested? e.g. symptoms or race

I'd like to get this merged first so we can get Mike and his team a first draft. Can add the other things immediately after.

tsibley:
> The only issue I've encountered is that a select * on the view is extremely slow. I'm curious if you experienced this, too.

Hmm, it's quick for me locally. We can try to troubleshoot this, but let's see what behaviour is like in production first since performance depends a lot on specific environments.

kairstenfay:
> > It looks like the sqitch tag also got out of order.
> 
> Hmm, how so?

Nevermind, I thought this PR already had a sqitch tag but I just added it. 

> > Also, would now be a good time to add in those extra variables Mike requested? e.g. symptoms or race
> 
> I'd like to get this merged first so we can get Mike and his team a first draft. Can add the other things immediately after.

Sounds good. I'm happy to merge if you approve.

PR #67, opened 2020-03-11
Scan redcap etl

joverlee521:
My remaining question is whether we need to decode the answers within `QuestionnaireResponse` or do we just leave them as codes.

If we do want to decode them, what do we decode them to? I'm assuming we will be returning these to the study group, so do we want to make them even more readable?

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -249,7 +249,7 @@ def create_encounter(record: dict, patient_reference: dict, locations: list) -> 
     site_reference = {
         "location": create_reference(
             reference_type = "Location",
-            identifier = create_identifier(f"{INTERNAL_SYSTEM}/site", 'scan')
+            identifier = create_identifier(f"{INTERNAL_SYSTEM}/site", 'SCAN')
         )
     }
     non_tract_references.append(site_reference)
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -1,0 +1,572 @@
+"""
+Process REDCap DETs that are specific to the
+COVID-19 Public Health Surveillance (SCAN) Project.
+"""
+import re
+import click
+import json
+import logging
+from uuid import uuid4
+from typing import Any, Callable, Dict, List, Mapping, Match, Optional, Union, Tuple
+from datetime import datetime
+from cachetools import TTLCache
+from id3c.db.session import DatabaseSession
+from id3c.cli.command.etl import redcap_det
+from id3c.cli.command.geocode import get_response_from_cache_or_geocoding
+from id3c.cli.command.location import location_lookup
+from seattleflu.id3c.cli.command import age_ceiling
+from .redcap_map import *
+from .fhir import *
+from . import race
+
+
+LOG = logging.getLogger(__name__)
+
+
+REVISION = 1
+
+REDCAP_URL = 'https://redcap.iths.org/'
+INTERNAL_SYSTEM = "https://seattleflu.org"
+
+PROJECT_ID = 20759
+REQUIRED_INSTRUMENTS = [
+    'screening',
+    'shipping_information',
+    'back_end_mail_scans',
+    'illness_questionnaire'
+    'nasal_swab_collection',
+    'post_collection_data_entry_qc'
+]
+
+
+@redcap_det.command_for_project(
+    "scan",
+    redcap_url = REDCAP_URL,
+    project_id = PROJECT_ID,
+    required_instruments = REQUIRED_INSTRUMENTS,
+    revision = REVISION,
+    help = __doc__)
+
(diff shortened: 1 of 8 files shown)


joverlee521:
[Ingest new REDCap project once survey available](https://trello.com/c/tLwh4CV7/368-ingest-new-redcap-project-once-survey-available)

kairstenfay:
> My remaining question is whether we need to decode the answers within `QuestionnaireResponse` or do we just leave them as codes.
> 
> If we do want to decode them, what do we decode them to? I'm assuming we will be returning these to the study group, so do we want to make them even more readable?

What's the list of answers you're considering decoding? And what are some examples of the coded values? Are we talking like 1/2 for yes/no?

joverlee521:
> > My remaining question is whether we need to decode the answers within `QuestionnaireResponse` or do we just leave them as codes.
> > If we do want to decode them, what do we decode them to? I'm assuming we will be returning these to the study group, so do we want to make them even more readable?
> 
> What's the list of answers you're considering decoding? And what are some examples of the coded values? Are we talking like 1/2 for yes/no?

These are the codes that Misja had put in place for us instead of the REDCap default numbers. Mostly referring to fields under `illness_questionnaire` instrument, relevant fields being:
income, housing_type, doctor_3e8fae, hospital_where, hospital_ed, smoke_9a005a, chronic_illness, overall_risk, country, state,

kairstenfay:
> These are the codes that Misja had put in place for us instead of the REDCap default numbers. Mostly referring to fields under `illness_questionnaire` instrument, relevant fields being:
> income, housing_type, doctor_3e8fae, hospital_where, hospital_ed, smoke_9a005a, chronic_illness, overall_risk, country, state,

Ah, I see now what you mean. I think it's fine to leave the coded questions as they are, because as you highlighted, we don't have any better values to code them to for now.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -0,0 +1,534 @@
...
+        source = f"{REDCAP_URL}{PROJECT_ID}/{redcap_record['record_id']}",
+        entries = list(filter(None, resource_entries))
+    )
+
+
+def locations(db: DatabaseSession, cache: TTLCache, record: dict) -> list:
+    """ Creates a list of Location resource entries from a REDCap record. """
+    lodging_options = [
+        'shelter',
+        'afl',
```
tsibley: I think an assisted-living facility is a residence, not lodging.

How are shelters categorized in our existing data sources?

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -0,0 +1,534 @@
...
+    """ Returns a FHIR Questionnaire Response resource entry. """
+
+    def combine_checkbox_answers(coded_question: str) -> Optional[List]:
+        """
+        Handles the combining "select all that apply"-type checkbox
+        responses into one list.
+
+        Uses our in-house mapping for race.
+        """
+        regex = rf'{coded_question}___[\w]*$'
```
tsibley: I think in our specific current usages it's not an issue, but generally-speaking it is best to be defensive and [escape](https://docs.python.org/3/library/re.html#re.escape) interpolations into regex patterns, e.g.:

```python
regex = rf'{re.escape(coded_question)}___[\w]*$'
```

This prevents metachars in the interpolation from being misinterpreted.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -0,0 +1,534 @@
...
+
+        return False
+
+
+    def combine_multiple_fields(field_prefix: str) -> Optional[List]:
+        """
+        Handles the combining of multiple fields asking the same question such
+        as country and state traveled.
+        """
+        regex = rf'^{field_prefix}[0-9]+$'
```
tsibley: Ditto re: escaping.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -0,0 +1,534 @@
...
+    'shipping_information',
+    'back_end_mail_scans',
+    'illness_questionnaire'
+    'nasal_swab_collection',
+    'post_collection_data_entry_qc'
+]
+
+
+@redcap_det.command_for_project(
+    "scan",
```
tsibley: I'd probably make this `scan-en` for English in anticipation of expanding to other languages. We can bundle all `scan-*` commands into this file as future work, and they can call the exact same command function.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
joverlee521: I based this off of our exisiting REDCap ETLs for swab and send, should I update those as well?

joverlee521:
Rebased to fix sqitch conflicts

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
tsibley: Ah, I missed that. Let's leave it this way for now then. Should be fine since we're going to use the primary-location concept in ID3C anyway to fallback from residence → lodging.

PR #68, opened 2020-03-14
Additional uw retro metadata

kairstenfay:
Extracts clinical test results, encounter status, encounter class, and discharge disposition from UW Retrospective data. 

Depends on https://github.com/seattleflu/id3c/pull/132. 

Note that, while the DiagnosticReport resources from clinical test result end up in `warehouse.presence_absence`, none of the other variables end up in encounter details. Should we change how our `warehouse.encounter` details document is structured? Should we add new columns to `warehouse.encounter`? Other thoughts?

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -229,10 +229,7 @@ def create_encounter_resource(encounter_identifier: List[dict],
         "class": encounter_class,
         "identifier": encounter_identifier,
         "status": encounter_status,
-        "period": {
-            "start": encounter_date,
-            "end": encounter_date,
-        },
+        "period": period,
         "subject": patient_reference,
     }
 
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -151,6 +151,7 @@ def create_encounter_resource(encounter_identifier: List[dict],
                               location_references: List[dict],
                               diagnosis: List[dict] = None,
                               contained: List[dict] = None,
+                              encounter_status = 'finished',
                               hospitalization: dict = None) -> dict:
     """
     Create encounter resource following the FHIR format
@@ -160,7 +161,7 @@ def create_encounter_resource(encounter_identifier: List[dict],
         "resourceType": "Encounter",
         "class": encounter_class,
         "identifier": encounter_identifier,
-        "status": "finished",
+        "status": encounter_status,
         "period": {
             "start": encounter_date,
             "end": encounter_date,
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -150,7 +150,8 @@ def create_encounter_resource(encounter_identifier: List[dict],
                               patient_reference: dict,
                               location_references: List[dict],
                               diagnosis: List[dict] = None,
-                              contained: List[dict] = None) -> dict:
+                              contained: List[dict] = None,
+                              hospitalization: dict = None) -> dict:
     """
     Create encounter resource following the FHIR format
     (http://www.hl7.org/implement/standards/fhir/encounter.html)
@@ -173,6 +174,8 @@ def create_encounter_resource(encounter_identifier: List[dict],
         encounter_resource["diagnosis"] = diagnosis
     if contained:
         encounter_resource["contained"] = contained
+    if hospitalization:
+        encounter_resource["hospitalization"] = hospitalization
 
     return encounter_resource
 
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -202,6 +202,28 @@ def create_encounter_resource(encounter_identifier: List[dict],
     Create encounter resource following the FHIR format
     (http://www.hl7.org/implement/standards/fhir/encounter.html)
     """
+    # XXX FIXME: We are currently using Encounter.period incorrectly, which
+    # became apparently only after we started using hospitalization data (from
+    # applicable retrospective data). An Encounter.period's start and end dates
+    # should reflect the time period covering the hospital admission and
+    # discharge.
+    #
+    # Currently, the FHIR ETL upserts an ID3C encounter (e.g. time of sample
+    # collectoin date) using the Encounter.period.start date. Ideally, the FHIR
+    # ETL would pull the encounter date (i.e. sample collection date) from
+    # Specimen.collection, but there are some organizational hurdles in the
+    # current code we'd have to address before being able to do so. For example,
+    # Encounters are processed before Specimens, and the two are only linked by
+    # Observations.
+    #
+    # kfay, 20 March 2020
+    period =  {
+        "start": encounter_date
+    }
+
+    if encounter_status == 'finished' or encounter_status == 'cancelled':
+        period['end'] = encounter_date
+
     encounter_resource = {
         "resourceType": "Encounter",
         "class": encounter_class,
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -4,7 +4,7 @@ REDCap DET ETL shared functions to create FHIR documents
 import logging
 import regex
 from itertools import filterfalse
-from typing import Iterable, NamedTuple, Optional, List
+from typing import Iterable, NamedTuple, Optional, List, Any
 from uuid import uuid4
 from datetime import datetime
 from id3c.cli.command.de_identify import generate_hash
@@ -12,6 +12,7 @@ from id3c.cli.command.de_identify import generate_hash
 
 LOG = logging.getLogger(__name__)
 
+SFS = "https://seattleflu.org"
 
 # CREATE FHIR RESOURCES
 def create_reference(reference_type: str = None,
@@ -440,3 +441,25 @@ def canonicalize_name(*parts: Iterable[str]) -> str:
         return collapse_whitespace(remove_non_word_chars(part)).strip().upper()
 
     return " ".join(map(canonicalize, parts))
+
+
+# XXX TODO: Define this as a TypedDict when we upgrade from Python 3.6 to
+# 3.8.  Until then, there's no reasonable way to type this data structure
+# better than Any.
+#   -trs, 24 Oct 2019
+observation_resource: Any = {
+    'resourceType': 'Observation',
+    'id': '',
+    'status': 'final',
+    'code': {
+        'coding': []
+    },
+    'valueBoolean': None,
+    'device': create_reference(
+        reference_type = 'Device',
+        identifier = create_identifier(
+            system = f'{SFS}/device',
+            value = 'Cepheid'
+        )
+    )
+}
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -4,7 +4,7 @@ REDCap DET ETL shared functions to create FHIR documents
 import logging
 import regex
 from itertools import filterfalse
-from typing import Iterable, NamedTuple, Optional, List, Any
+from typing import Iterable, NamedTuple, Optional, List, Any, Callable
 from uuid import uuid4
 from datetime import datetime
 from id3c.cli.command.de_identify import generate_hash
@@ -49,6 +49,53 @@ def create_patient_resource(patient_identifier: List[dict], gender: str) -> dict
         "identifier": patient_identifier,
         "gender": gender
     })
+
+
+def create_diagnostic_report(redcap_record:dict,
+                             patient_reference: dict,
+                             specimen_reference: dict,
+                             create_device_result_observation_resource: Callable) -> Optional[dict]:
+    """
+    Create FHIR diagnostic report from given *redcap_record*. Attaches
+    observation resources to it using the given, device-specific
+    *create_device_result_observation_resource* function.
+
+    Links the generated diagnostic report to a specific *patient_reference* and
+    *specimen_reference*.
+    """
+    clinical_results = create_device_result_observation_resource(redcap_record)
+    if not clinical_results:
+        return None
+
+    diagnostic_result_references = []
+    for result in clinical_results:
+        reference = create_reference(
+            reference_type = 'Observation',
+            reference = '#' + result['id']
+        )
+        diagnostic_result_references.append(reference)
+
+    collection_datetime = redcap_record['collection_date']
+
+    diagnostic_code = create_codeable_concept(
+        system = 'http://loinc.org',
+        code = '85476-0',
+        display = 'FLUAV and FLUBV and RSV pnl NAA+probe (Upper resp)'
+    )
+
+    diagnostic_report_resource = create_diagnostic_report_resource(
+        datetime = collection_datetime,
+        diagnostic_code = diagnostic_code,
+        patient_reference  = patient_reference,
+        specimen_reference = specimen_reference,
+        result = diagnostic_result_references,
+        contained = clinical_results
+    )
+
+    return (create_resource_entry(
+        resource = diagnostic_report_resource,
+        full_url = generate_full_url_uuid()
+    ))
 
 
 def create_diagnostic_report_resource(datetime: str,
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -447,19 +447,24 @@ def canonicalize_name(*parts: Iterable[str]) -> str:
 # 3.8.  Until then, there's no reasonable way to type this data structure
 # better than Any.
 #   -trs, 24 Oct 2019
-observation_resource: Any = {
-    'resourceType': 'Observation',
-    'id': '',
-    'status': 'final',
-    'code': {
-        'coding': []
-    },
-    'valueBoolean': None,
-    'device': create_reference(
-        reference_type = 'Device',
-        identifier = create_identifier(
-            system = f'{SFS}/device',
-            value = 'Cepheid'
+def observation_resource(device: str) -> Any:
+    """
+    Returns a minimally-filled FHIR Observation Resource with a given
+    *device* value.
+    """
+    return {
+        'resourceType': 'Observation',
+        'id': '',
+        'status': 'final',
+        'code': {
+            'coding': []
+        },
+        'valueBoolean': None,
+        'device': create_reference(
+            reference_type = 'Device',
+            identifier = create_identifier(
+                system = f'{SFS}/device',
+                value = device
+            )
         )
-    )
-}
+    }
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -54,14 +54,18 @@ def create_patient_resource(patient_identifier: List[dict], gender: str) -> dict
 def create_diagnostic_report(redcap_record:dict,
                              patient_reference: dict,
                              specimen_reference: dict,
+                             diagnostic_code: dict,
                              create_device_result_observation_resource: Callable) -> Optional[dict]:
     """
-    Create FHIR diagnostic report from given *redcap_record*. Attaches
-    observation resources to it using the given, device-specific
-    *create_device_result_observation_resource* function.
+    Create FHIR diagnostic report from given *redcap_record*.
 
     Links the generated diagnostic report to a specific *patient_reference* and
     *specimen_reference*.
+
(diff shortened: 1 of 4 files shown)


lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -94,7 +103,7 @@ def create_specimen(record: dict, patient_reference: dict) -> tuple:
     """ Returns a FHIR Specimen resource entry and reference. """
     barcode = record["barcode"]
     specimen_identifier = create_identifier(f"{SFS}/sample", barcode)
-    specimen_type = "NSECR" # Nasal swab.
+    specimen_type = "NSECR" # Nasal swab
```
kairstenfay: oops stray period deletion

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -275,6 +284,187 @@ def determine_questionnaire_items(record: dict) -> List[dict]:
...
+        patient_reference  = patient_reference,
+        specimen_reference = specimen_reference,
+        result = diagnostic_result_references,
+        contained = clinical_results
+    )
+
+    return (create_resource_entry(
+        resource = diagnostic_report_resource,
+        full_url = generate_full_url_uuid()
+    ))
```
kairstenfay: There's a lot of duplicated code here between this redcap-det etl and the kiosk one.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -275,6 +284,187 @@ def determine_questionnaire_items(record: dict) -> List[dict]:
...
+            'display': 'Coronavirus infection',  # NOT clinical finding
+        },
+        '440925005': {
+            'system': 'http://snomed.info/sct',
+            'code': '440925005',
+            'display': 'Human rhinovirus present',
+        },
+        '440930009': {
+            'system': 'http://snomed.info/sct',
+            'code': '440930009',
```
kairstenfay: Having snomed codes in both the key and value in this dict is a bit odd to me, but it was the easiest way to de-duplicate clinical test results without creating some ad-hoc, arbitrary intermediary name like `Flu A` to code against.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -275,6 +284,187 @@ def determine_questionnaire_items(record: dict) -> List[dict]:
...
+
+    results: Dict[str, bool] = {}
+
+    # Populate dict of tests administered during encounter by filtering out
+    # null results. Map the REDCap test variable to the snomed code. In the
+    # event of duplicate clinical findings, prioritize keeping positive results.
+    for test in redcap_to_snomed_map:
+        code = redcap_to_snomed_map[test]
+
+        # Skip updating positive results
```
kairstenfay: copy-edit: Skip updating results for tests already marked as positive

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -248,6 +251,61 @@ def find_sample_origin_by_barcode(db: DatabaseSession, barcode: str) -> Optional
...
+    """
+    Returns an Encounter.hospitalization entry created from a given *redcap_record*.
+    (https://www.hl7.org/fhir/encounter-definitions.html#Encounter.hospitalization)
+    """
+    disposition = discharge_disposition(redcap_record)
+
+    # For now, dischargeDisposition is the only info we store in
+    # Encounter.hospitalization. If this info isn't available, skip creating
+    # this resource entry.
+    if not discharge_disposition:
```
kairstenfay: bug: change to `if not disposition`

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -306,6 +304,32 @@ def discharge_disposition(redcap_record: dict) -> Optional[str]:
     return mapper[disposition]
 
 
+def create_encounter_class(redcap_record: dict) -> dict:
+    """
+    Creates an Encounter.class coding from a given *redcap_record*.
+    This attribute is required by FHIR for an Encounter resource.
+    (https://www.hl7.org/fhir/encounter-definitions.html#Encounter.class)
+    """
```
kairstenfay: copy-edit: describe default of `AMB`

kairstenfay:
[Ingest patient outcome, inpatient/outpatient, and COVID-19 lab results for UW retros](https://trello.com/c/o43z6zyu/371-ingest-patient-outcome-inpatient-outpatient-and-covid-19-lab-results-for-uw-retros)

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
kairstenfay: I'll probably push a commit to refactor the code between these two ETLs at the risk of the code looking slightly more convoluted

kairstenfay: Now done in the latest commit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -275,6 +395,145 @@ def determine_questionnaire_items(record: dict) -> List[dict]:
...
+        return None
+
+    test_result_map = {
+        'Positive': True,
+        'Negative': False,
+        'None detected.': False,
+        'Test not applicable': None,
+        'Canceled by practitioner': None,
+        'Duplicate request': None,
+    }
```
joverlee521: Looks like we need to add `Detected` to this map.

lib/seattleflu/id3c/cli/command/etl/fhir.py
```diff
@@ -50,6 +51,53 @@ def create_patient_resource(patient_identifier: List[dict], gender: str) -> dict
...
+        )
+        diagnostic_result_references.append(reference)
+
+    collection_datetime = redcap_record['collection_date']
+
+    diagnostic_code = create_codeable_concept(
+        system = 'http://loinc.org',
+        code = '85476-0',
+        display = 'FLUAV and FLUBV and RSV pnl NAA+probe (Upper resp)'
+    )
```
joverlee521: This diagnostic code is specifice to the Cepheid assay.

lib/seattleflu/id3c/cli/command/etl/fhir.py
```diff
@@ -158,7 +208,7 @@ def create_encounter_resource(encounter_identifier: List[dict],
         "resourceType": "Encounter",
         "class": encounter_class,
         "identifier": encounter_identifier,
-        "status": "finished",
+        "status": encounter_status,
         "period": {
             "start": encounter_date,
             "end": encounter_date,
```
joverlee521: Do we need to also change the period if the `Encounter.status` is different?

lib/seattleflu/id3c/cli/command/etl/fhir.py
kairstenfay: I see. Do you know how to find the right code for these tests?

lib/seattleflu/id3c/cli/command/etl/fhir.py
kairstenfay: Ah, right. If a status is not finished, we would not want an end date for the period.

lib/seattleflu/id3c/cli/command/etl/fhir.py
joverlee521: I think this might work for the clinical report: https://loinc.org/77029-7/
It seems general enought to encompass all the different pathogens we ingest.

joverlee521: Or this one https://loinc.org/60566-7/ is even more general

kairstenfay: Thank you!

joverlee521: This is another one https://loinc.org/78922-2/
There's so many I don't even know which one fits the best here....
(I'm just googling "loinc respiratory pathogen")

lib/seattleflu/id3c/cli/command/etl/fhir.py
kairstenfay: Oh but wait. I remember why I didn't adjust encounter period based on the hospitalization start or end date. 
We're not tracking the hospitalization period. That's a separate entity. We're only interested in the encounter where a swab was taken. So that always has a start/end date of the swab collection date.

kairstenfay: This contradicts the encounter status change I made, though. Perhaps we should continue setting all Encounter.status to 'finished'.

joverlee521: Huh, then shouldn't the status of the encounter always be finished (if this is tracking the encounter where the swab was taken)?

kairstenfay: Exactly :smile:

kairstenfay: Hmm. Jover just informed me on Slack she found this in the FHIR docs:
> The duration recorded in the period of this encounter covers the entire scope of this hospitalization record.

https://www.hl7.org/fhir/encounter-definitions.html#Encounter.hospitalization

If we stay true to the docs and change our Encounter.period.start and Encounter.period.end to reflect the hospitalization start/end times, we will need to update the FHIR ETL to extract a date other than start date (if available) for our swab collection (ID3C encounter) date.

joverlee521: The other option is to keep the Encounter pointed at when/where the swab was taken and move the hospitalization details to QuestionnaireResponse

kairstenfay: That's also a possibility.

If we lean toward more structure (using Encounter.hospitalization), it looks like we should be able to use [Specimen.collection](https://www.hl7.org/fhir/specimen-definitions.html#Specimen.collection) to store collection date.

kairstenfay: Even though it's slightly more work, I would prefer to use Encounter.hospitalization, Encounter.status and Encounter.period (accordingly), and Specimen.collection. But I also understand if we want to punt on this for now.

lib/seattleflu/id3c/cli/command/etl/fhir.py
kairstenfay: There are so many... I think I will go with the most general ("Unspecified") one for now. I'm not sure how to choose the appropriate one.

joverlee521: Found the code on [UW](https://testguide.labmed.uw.edu/public/view/REVSQT) under `Billing/Coding`. It's https://loinc.org/50219-5/

lib/seattleflu/id3c/cli/command/etl/fhir.py
joverlee521: Yeah, I feel like we should follow FHIR, but we might want to punt it depending on how much ROR work we're getting...

lib/seattleflu/id3c/cli/command/etl/fhir.py
kairstenfay: > Found the code on [UW](https://testguide.labmed.uw.edu/public/view/REVSQT) under `Billing/Coding`. It's https://loinc.org/50219-5/

Awesome find! I used the [UW Laboratory Test Guide](https://testguide.labmed.uw.edu/public/index) website you linked to look up the codes for all the tests.

kairstenfay: Ughh. Another FHIR rabbit hole.
We can only have one `DiagnosticReport.code` value per resource. However, it is possible to have multiple entries for a `codeableConcept`. From the [FHIR docs](http://hl7.org/fhir/datatypes.html#CodeableConcept): 
> More than one code may be used in CodeableConcept. The concept may be coded multiple times in different code systems (or even multiple times in the same code systems, where multiple forms are possible, such as with SNOMED CT). 

Right now we have three LOINC codes for a DiagnosticReport: 
```
    uw_billing_to_loinc_map = {
        'NCVQLT': '94306-8',
        'REVSQT': '50219-5',
        'FLUMC': '48509-4',
    }
```

I'm not sure if it's appropriate to include all three LOINC codes in the `DiagnosticReport.code.codeableConcept` list, because they're all mutually exclusive. I think the true-to-FHIR way would be to insert a new DiagnosticReport for each LOINC code. Thoughts?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
tsibley: Yep, this makes sense to me. :+1:

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -275,6 +403,146 @@ def determine_questionnaire_items(record: dict) -> List[dict]:
...
+        '441344004': {
+            'system': 'http://snomed.info/sct',
+            'code': '441344004',
+            'display': 'Human parainfluenza virus present',
+        },
+        '186747009': {
+            'system': 'http://snomed.info/sct',
+            'code': '186747009',
+            'display': 'Coronavirus infection',  # NOT clinical finding
+        },
```
tsibley: Looks like this is unused?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -239,6 +259,114 @@ def find_sample_origin_by_barcode(db: DatabaseSession, barcode: str) -> Optional
...
+        "ED": "EMER",  # can also code as "AMB"
+    }
+
+    if encounter_class and encounter_class not in mapper:
+        raise Exception(f"Unknown encounter class «{encounter_class}».")
+
+    # Default to 'AMB' if encounter_class not defined
+    return create_coding(
+        system = "http://terminology.hl7.org/CodeSystem/v3-ActCode",
+        code = encounter_class or 'AMB'
```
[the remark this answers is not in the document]
kairstenfay: Thanks for catching that bug

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -239,6 +259,114 @@ def find_sample_origin_by_barcode(db: DatabaseSession, barcode: str) -> Optional
...
+    (https://www.hl7.org/fhir/encounter-definitions.html#Encounter.status)
+    """
+    status = redcap_record['encounter_status']
+    if not status:
+        return 'finished'
+
+    mapper = {
+        'ARRIVED': 'arrived',
+        'DISCHARGED': 'finished',
+        'LWBS': 'cancelled +',  # LWBS = left without being seen.
```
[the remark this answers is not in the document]
kairstenfay: Ha, yes you're exactly right. Thanks.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -239,6 +259,114 @@ def find_sample_origin_by_barcode(db: DatabaseSession, barcode: str) -> Optional
...
+        'Disch/Trans to a distinct Rehab Unit/Hospital': 'rehab',
+        'SNF-SKILLED NURSING FACILITY': 'snf',
+        'DISCH/TRANS TO COURT/LAW ENFORCEMENT': 'oth',
+        'Other Institution - Not Defined Elsewhere': 'oth',
+    }
+
+    if disposition not in mapper:
+        raise Exception(f"Unknown discharge disposition value «{disposition}».")
+
+    return mapper[disposition]
```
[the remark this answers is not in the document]
kairstenfay: Easy enough to fix. I stripped leading/trailing white space and cast to lowercase before mapping. I'm not sure it makes sense to remove all spaces, though, because our codes are all either one word/initialism or are longer strings of text.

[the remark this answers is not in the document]
tsibley: Ah, not removing all spaces, but normalizing it for comparison, e.g. from the example I linked to:

```python
    def standardize_whitespace(string):
        return re.sub(r"\s+", " ", string.strip())
```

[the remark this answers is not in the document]
kairstenfay: Ah I misread the function. I can standardize white space, too.

lib/seattleflu/id3c/cli/command/etl/fhir.py
kairstenfay: I'm ok with marking the Encounter.period as a future `FIXME`.

Good find on the Encounter.partOf. I agree, it sounds like unnecessary work, and I _really_ don't want to process a nested Encounter resource right now. I'd prefer to kick this can down the road.

lib/seattleflu/id3c/cli/command/etl/fhir.py
kairstenfay: @tsibley @joverlee521 any thoughts here?

joverlee521: I agree that the true FHIR way is a new DiagnosticReport for each diagnostic code. I want to say that the FHIR ETL is capable of handling multiple DiagnosticReports in one Bundle.

lib/seattleflu/id3c/cli/command/etl/fhir.py
```diff
@@ -149,16 +195,40 @@ def create_encounter_resource(encounter_identifier: List[dict],
...
+    # Encounters are processed before Specimens, and the two are only linked by
+    # Observations.
+    #
+    # kfay, 20 March 2020
+    period =  {
+        "start": encounter_date
+    }
+
+    if encounter_status == 'finished' or encounter_status == 'cancelled':
+        period['end'] = encounter_date
```
tsibley: `period` needs to be used below within `encounter_resource`… :-)

kairstenfay: derp

kairstenfay:
> One last fix, but otherwise looks good. Thanks!

Thanks. Unfortunately there's still one more piece of work: to use the correct LOINC codes for a DiagnosticReport, I have to actually generate multiple DiagnosticReports. I'll push this as a new commit then re-request review.

tsibley:
Oh, I misunderstood what point of the WIP commit then. Ok!

tsibley:
> Thanks. Unfortunately there's still one more piece of work: to use the correct LOINC codes for a DiagnosticReport, I have to actually generate multiple DiagnosticReports. I'll push this as a new commit then re-request review.

Hmm. So we don't actually use the DiagnosticReport.code in our ingest, right? The point of including it is to specify what the set of Observations are expected to be and to satisfy the spec, but our ingest ignores it. Instead of doing this work to generate multiple DiagnosticReports, it'd be entirely within spec to just use an internal code. This seems fine to me since the data for these samples is really a purely-synthetic "diagnostic panel" assembled after the fact. That is, we're not trying to represent the _actual_ panels that were run.

kairstenfay:
> Hmm. So we don't actually use the DiagnosticReport.code in our ingest, right? The point of including it is to specify what the set of Observations are expected to be and to satisfy the spec, but our ingest ignores it. Instead of doing this work to generate multiple DiagnosticReports, it'd be entirely within spec to just use an internal code. This seems fine to me since the data for these samples is really a purely-synthetic "diagnostic panel" assembled after the fact. That is, we're not trying to represent the _actual_ panels that were run.

No, we don't; it's just another measure we've taken to produce valid FHIR documents. What would you suggest as an internal diagnostic code, @tsibley ?

tsibley:
Something like this might be fine? I don't think it matters too much at the moment.

```json
{"system": "https://seattleflu.org/presence-absence-panel", "code": "uw-retrospective"}
```

If you're already well on your way to multiple DiagnosticReports, then feel free to continue. Just wanted to offer an easier way out (which still results in a valid FHIR document).

kairstenfay:
> Something like this might be fine? I don't think it matters too much at the moment.
> 
> ```json
> {"system": "https://seattleflu.org/presence-absence-panel", "code": "uw-retrospective"}
> ```
> 
> If you're already well on your way to multiple DiagnosticReports, then feel free to continue. Just wanted to offer an easier way out (which still results in a valid FHIR document).

That sounds great to me. I had not started creating multiple DiagnosticReports yet, just had mapped UW's LOINC codes to the tests administered. I'm super happy to just apply a blanket, internal code here. Pushed in a new commit.

PR #69, opened 2020-03-16
etl redcap-det uw-retro: fix sample barcode lookup

joverlee521:
Instead of using `identifer like`, join with `warehouse.identifier`
do directly compare with citext barcode for case-insensitivity and
indexed operation.


joverlee521:
[Fix barcode lookup in REDCap ETL for UW Retros](https://trello.com/c/Z6542Hf7/396-fix-barcode-lookup-in-redcap-etl-for-uw-retros)

PR #70, opened 2020-03-19
Additional hCoV view variables

joverlee521:
* My attempt to optimize the FHIR encounter details query decreased execution time from ~2500ms to ~450ms locally.

* I created a new view `shipping.fhir_encounter_details_v2` which has all of the illness questionnaire fields from the SCAN project. 

* I've joined the new fhir view with hCoV19 view, and it slows down the query. I'm wondering if we should leave the hCoV19 view as is and ask Mike's team to join the two views when they need the extra variables. 

Edit: ~I'm also punting on CT values for now because not sure how that's going to look with the CLIA targets from Samplify....~

Edit2: This now includes CT values for hCoV19


joverlee521:
[Add additional variables for `hcov19_observation_v1`](https://trello.com/c/W9Zu5P7Z/402-add-additional-variables-for-hcov19observationv1)

joverlee521:
Oh gosh. Now that we are ingesting SCAN encounters without samples first, do we need to change `shipping.hcov19_observation_v1` to pull `from warehouse.encounter` instead of `from warehouse.sample`? 😑

schema/deploy/shipping/views.sql
```diff
@@ -1029,6 +1029,12 @@ grant select
    to "incidence-modeler";
 
 
+revoke all
+    on shipping.hcov19_observation_v1
+  from "hcov19-visibility";
```
kairstenfay: Is this step necessary before dropping a view just to replace it?

schema/revert/shipping/views.sql
```diff
@@ -740,6 +740,16 @@ comment on view shipping.return_results_v2 is
...
 
+revoke all
+    on shipping.hcov19_observation_v1
+  from "hcov19-visibility";
+drop view shipping.hcov19_observation_v1;
+
+revoke all
+    on shipping.fhir_encounter_details_v2
+  from "incidence-modeler";
+drop view shipping.fhir_encounter_details_v2;
```
tsibley: Why the revokes here before the drop?

schema/deploy/shipping/views.sql
```diff
@@ -740,24 +754,316 @@ comment on view shipping.return_results_v2 is
...
+      left join hospital_leave using (encounter_id)
+      left join smoking using (encounter_id)
+      left join chronic_illness using (encounter_id)
+      left join overall_risk_health using (encounter_id)
+      left join overall_risk_setting using (encounter_id)
+      left join long_term_type using (encounter_id)
+;
+
+comment on view shipping.fhir_encounter_details_v1 is
+  'A view of encounter details that are in FHIR format';
```
tsibley: This should be `_v2`.

schema/deploy/shipping/views.sql
tsibley: (It isn't.)

schema/deploy/shipping/views.sql
```diff
@@ -740,24 +754,316 @@ comment on view shipping.return_results_v2 is
...
+            from questionnaire_responses
+          where "linkId" in ('travel_countries','travel_countries_phs')
+          group by encounter_id
+        ),
+
+        travel_states as (
+          select encounter_id,
+                 bool_or(boolean_response) as travel_states
+            from questionnaire_responses
+          where "linkId" in ('travel_states', 'travel_countries_phs')
```
tsibley: This should be `travel_states_phs`.

schema/deploy/shipping/views.sql
```diff
@@ -740,24 +754,316 @@ comment on view shipping.return_results_v2 is
...
+                 jsonb_to_recordset("answer") as answer("valueString" text, "valueBoolean" bool, "valueDate" text, "valueCoding" jsonb),
+                 jsonb_to_record("valueCoding") as code("code" text)
+          where "linkId" in ('redcap_event_name',
+                             'vaccine',
+                             'race',
+                             'insurance',
+                             'ethnicity',
+                             'travel_countries',
+                             'travel_countries_phs',
+                             'country',
```
tsibley: This `linkId` isn't used below. Should it be?

schema/deploy/shipping/views.sql
```diff
@@ -740,24 +754,316 @@ comment on view shipping.return_results_v2 is
...
+        income as (
+          select encounter_id,
+                 string_response as income
+            from questionnaire_responses
+          where "linkId" = 'income'
+        ),
+
+        housing_type as (
+          select encounter_id,
+                 string_response as housing_type
```
tsibley: This is a single-choice field, so I think would safe to grab just the first element and make analysis easier.

Some other string fields are the same way, e.g. house_members and clinical_care. I didn't check others, but would be good to. Some of this is clear from looking at existing data (e.g.  house_members), others have to look at REDCap field type.

Think this is worth the extra work to check?

schema/deploy/shipping/views.sql
joverlee521: Huh, not sure how I got it in my head that it was necessary...will remove

schema/revert/shipping/views.sql
joverlee521: 🤷‍♀ I have no idea how I got it in my head that was necessary...

schema/deploy/shipping/views.sql
joverlee521: Yes! Thank you, I missed `country` and `state` 😑

schema/deploy/shipping/views.sql
joverlee521: Yup, will update!

joverlee521:
Updated to now separate UW clinical and BBI hCoV-19 results in `hcov19_observation_v1`

Also thoughts on my previous question:
>Oh gosh. Now that we are ingesting SCAN encounters without samples first, do we need to change shipping.hcov19_observation_v1 to pull from warehouse.encounter instead of from warehouse.sample? 😑

kairstenfay:
> Also thoughts on my previous question:
> 
> > Oh gosh. Now that we are ingesting SCAN encounters without samples first, do we need to change shipping.hcov19_observation_v1 to pull from warehouse.encounter instead of from warehouse.sample? expressionless

If this view is supposed to be the central source of truth for all samples or encounters tested for hCoV-19 or since Jan 1 then I think we may want to do a `full outer join` b/w samples and encounters.

joverlee521:
> > Also thoughts on my previous question:
> > > Oh gosh. Now that we are ingesting SCAN encounters without samples first, do we need to change shipping.hcov19_observation_v1 to pull from warehouse.encounter instead of from warehouse.sample? expressionless
> 
> If this view is supposed to be the central source of truth for all samples or encounters tested for hCoV-19 or since Jan 1 then I think we may want to do a `full outer join` b/w samples and encounters.
> 
> Alternatively, if we're letting customers like Roy join encounter data to this view, could we guarantee that they could do a full outer join b/w samples (this view) and encounters (the other view) safely?

Ah, yeah forgot to update on here. I think it would be easiest to just have a completely separate view for SCAN encounters and Roy can use that to see the up-to-date demographic data

PR #71, opened 2020-03-19
views: Add SCAN return of results

kairstenfay:
Add a new view for the SCAN return of results (RoR) data that will be
used with the UW Lab Med API.

----

Opening as a DRAFT as this view contains some columns that are used in debugging the soon-to-be-submitted backoffice cron job for joining exported REDCap data with an export of this view.


schema/deploy/shipping/views.sql
```diff
@@ -844,4 +844,95 @@ comment on view shipping.hcov19_observation_v1 is
...
+create or replace view shipping.scan_return_results_v1 as
+
+    with status_codes as (
+        -- Create status codes as required by Lab Med's results portal
+        with x(status_code, status) as (
+            values
+            ('DET', 'positive result, colored red + bold'),
+            ('INCON', 'inconclusive, colored orange; treated as positive with respect to information given'),
+            ('NDET', 'not detected/negative, colored blue'),
+            ('pending-other', null),
```
kairstenfay: I'm currently treating `pending-other` as the scenario where a sample tests positive but the participant has not yet been contacted. No sample in this view (as it stands) will ever contain the status `pending-other`.

schema/deploy/shipping/views.sql
```diff
@@ -844,4 +844,95 @@ comment on view shipping.hcov19_observation_v1 is
...
+            ('pending-lab', 'We are working on it!'),
+            ('cancelled', null),
+            ('notReceived', 'This is our code we made up')
```
kairstenfay: This code is not contained in the codes UW Lab Med sent us.

schema/deploy/shipping/views.sql
```diff
@@ -844,4 +844,95 @@ comment on view shipping.hcov19_observation_v1 is
...
+            case
+                when sample_id is null then 'notReceived'  -- todo new code
+                when sample_id is not null and present is null then 'pending-lab'
+                when target.identifier is not null and present is null then 'INCON'
+                when present is true then 'DET'
+                when present is false then 'NDET'
+            end as status_code,
+            target.identifier,  -- todo debug
+            lineage,  -- todo debug
+            present  -- todo debug
```
kairstenfay: These columns will be removed before merging.

schema/deploy/shipping/views.sql
```diff
@@ -844,4 +844,95 @@ comment on view shipping.hcov19_observation_v1 is
...
+
+        from warehouse.identifier
+        join warehouse.identifier_set using (identifier_set_id)
+        left join warehouse.sample on uuid::text = sample.collection_identifier
+        left join warehouse.encounter using (encounter_id)
+        left join warehouse.presence_absence using (sample_id)
+        left join warehouse.target using (target_id)
+        left join warehouse.organism using (organism_id)
+
+    where identifier_set.name in('collections-scan', 'collections-swab&send') -- todo
```
kairstenfay: only `collections-scan` will be used in production. Including an additional collection is useful for debugging.

joverlee521: I thought we were reporting on all prospective samples? Should this include prospective SFS collections?

schema/deploy/shipping/views.sql
joverlee521: If we can't have extra codes, I think we just want to merge this with `pending-lab`?

schema/deploy/shipping/views.sql
```diff
@@ -844,4 +844,95 @@ comment on view shipping.hcov19_observation_v1 is
...
+
+    results as (
+        select distinct on (barcode, lineage)
+            sample_id,
+            barcode as qrcode,
+            encountered as collect_ts,
+            case
+                when sample_id is null then 'notReceived'  -- todo new code
+                when sample_id is not null and present is null then 'pending-lab'
+                when target.identifier is not null and present is null then 'INCON'
```
joverlee521: I think with this current logic it won't ever reach 'INCON'.

schema/deploy/shipping/views.sql
```diff
@@ -844,4 +844,95 @@ comment on view shipping.hcov19_observation_v1 is
...
+        left join warehouse.sample on uuid::text = sample.collection_identifier
+        left join warehouse.encounter using (encounter_id)
+        left join warehouse.presence_absence using (sample_id)
+        left join warehouse.target using (target_id)
+        left join warehouse.organism using (organism_id)
+
+    where identifier_set.name in('collections-scan', 'collections-swab&send') -- todo
+        and target.control is distinct from true
+        and (lineage is null or
+            lineage <@ '{"Human_coronavirus.2019"}'::ltree[])
```
joverlee521: I think we might want to add a date cutoff? We don't want to report on earlier samples.

schema/deploy/shipping/views.sql
kairstenfay: `pending-lab` as I have it is our equivalent for `processing`. I'd ideally like to keep "processing" distinct from "not received".

kairstenfay: We could make it "pending-other", but I was planning on using "pending-other" as our internal sign that the test is positive but participant hasn't yet been contacted, so hold results.

schema/deploy/shipping/views.sql
kairstenfay: I think it's fine if we just stick to `collections-scan`.

schema/deploy/shipping/views.sql
kairstenfay: Thanks for pointing that out. I wasn't sure what classified as an inconclusive result. Do you have any insight here? This may not be determinable until we see the new Samplify target data.

kairstenfay:
[Command to return Name + Results if given correct Barcode + DOB](https://trello.com/c/K2uPFEMN/404-command-to-return-name-results-if-given-correct-barcode-dob)

schema/deploy/shipping/views.sql
joverlee521: Yeah, just not sure we want to report `notReceived`...Well I think this also loops back to the language issue. No one has given us any guidance on the language to use for the return results site...

schema/deploy/shipping/views.sql
joverlee521: I think you just need to flip the order of the two `when...` statements? But yeah, also not entirely sure how we are modeling inconclusive results within our database yet...

schema/deploy/shipping/views.sql
joverlee521: Yeah, made that comment when I thought we were also returning SFS results through the same website.

schema/deploy/shipping/views.sql
kairstenfay: We chatted on slack and decided to use `pending` (no distinction)

schema/deploy/shipping/views.sql
kairstenfay: I updated it to make it simpler: `if presence_absence_id is not null and present is null then 'inconclusive'`

kairstenfay:
This is ready to be merged once the `--todo`s are removed. I've left them in there for now because it will make testing https://github.com/seattleflu/backoffice/pull/25 easier.

schema/deploy/shipping/views.sql
```diff
@@ -844,4 +844,73 @@ comment on view shipping.hcov19_observation_v1 is
...
 
 
+create or replace view shipping.scan_return_results_v1 as
+    select distinct on (barcode, lineage)
+        sample_id,
+        barcode as qrcode,
+        encountered as collect_ts,
+        case
+            when sample_id is null then 'not-received'
+            when sample_id is not null and present is null then 'pending'
```
joverlee521: I think `sample.sample_id is not null and presence_absence_id is null` is more specific for pending. Or else this will never reach 'inconclusive' case

kairstenfay: Ah right, I'm using `presence_absence_id` to check for a test result anyway, so using that instead of a presence result is also more consistent.

schema/deploy/shipping/views.sql
```diff
@@ -844,4 +844,73 @@ comment on view shipping.hcov19_observation_v1 is
   'Custom view of hCoV-19 samples with presence-absence results and best available encounter data';
 
 
+create or replace view shipping.scan_return_results_v1 as
+    select distinct on (barcode, lineage)
+        sample_id,
+        barcode as qrcode,
+        encountered as collect_ts,
```
tsibley: I'd make this a date instead of a timestamp. The time may give a false sense of precision and is not necessary for returning results.

schema/deploy/shipping/views.sql
```diff
@@ -844,4 +844,73 @@ comment on view shipping.hcov19_observation_v1 is
...
+    left join warehouse.sample on uuid::text = sample.collection_identifier
+    left join warehouse.encounter using (encounter_id)
+    left join warehouse.presence_absence using (sample_id)
+    left join warehouse.target using (target_id)
+    left join warehouse.organism using (organism_id)
+
+    where identifier_set.name in('collections-scan', 'collections-swab&send') -- todo
+        and target.control is distinct from true
+        and (lineage is null or
+            lineage <@ '{"Human_coronavirus.2019"}'::ltree[])
```
tsibley: The combination of left joins across multiple tables above + the where condition `lineage is null or lineage <@ …` here suffers from the same problems as we discussed in the `hcov19_observation_v1` view. It works now, but there is the chance for future targets without a linked organism to slip into these results. That'd be bad, as the view would report a result for a non-hCoV-19 target. I think better to join to a subquery or CTE which selects just the hCoV-19 results, similar to what was done in `hcov19_observation_v1`.

kairstenfay: Ah, right. I pushed a new commit that addresses this bug.

PR #72, opened 2020-03-20
hCoV Slack alerts

joverlee521:

The one thing I'm not sure about is how to filter for only prospective samples. I've currently set that filter as `sample.details -> sample_origin` is either `kiosk` or `mail`. We might add `kp` to that list if Kaiser gets back to us on how their consents are set up (see [slack](https://seattle-flu-study.slack.com/archives/CCAEWSFTK/p1584651186083000?thread_ts=1584567603.073900&cid=CCAEWSFTK))

### DIFF
--- a/lib/seattleflu/id3c/cli/command/reportable_conditions.py
+++ b/lib/seattleflu/id3c/cli/command/reportable_conditions.py
@@ -92,11 +92,7 @@ def notify(*, action: str):
                     LOG.info(f"No site found for presence_absence_id «{record.id}». " +
                         "Inferring site from manifest data.")
 
-                if record.lineage == 'Human_coronavirus.2019':
-                    metabase_link = "https://backoffice.seattleflu.org/metabase/question/466"
-                    url = SLACK_WEBHOOK_REPORTING_HCOV19
-
-                response = send_slack_post_request(record, url, metabase_link)
+                response = send_slack_post_request(record, SLACK_WEBHOOK_REPORTING_HCOV19)
 
                 if response.status_code == 200:
                     mark_processed(db, record.id, {"status": "sent Slack notification"})
@@ -138,7 +134,7 @@ def notify(*, action: str):
             db.rollback()
 
 
-def send_slack_post_request(record: Any, url: str, metabase_link: str) -> requests.Response:
+def send_slack_post_request(record: Any, url: str) -> requests.Response:
     """
     Sends a POST request to a channel-specific Slack webhook *url*. The payload
     of this POST request is composed using Slack blocks. These blocks provide
@@ -148,9 +144,12 @@ def send_slack_post_request(record: Any, url: str, metabase_link: str) -> reques
     containing minimal sample details.
     """
     data = {
-        "sample": record.barcode,
+        "sample_barcode": record.sample_barcode,
+        "collection_barcode": record.collection_barcode,
+        "clia_barcode": record.clia_barcode,
         "site": record.site,
-        "condition": record.lineage
+        "condition": record.lineage,
+        "language": record.language
     }
 
     if not record.site:
@@ -173,8 +172,7 @@ def send_slack_post_request(record: Any, url: str, metabase_link: str) -> reques
             "text": {
                 "type": "mrkdwn",
                 "text": dedent(f"""
-                :rotating_light: @channel {record.lineage} detected. \n
-                *<{metabase_link}|Go to Metabase>*
+                :rotating_light: @channel {record.lineage} detected.
                 """)
             }
         },
--- a/lib/seattleflu/id3c/cli/command/reportable_conditions.py
+++ b/lib/seattleflu/id3c/cli/command/reportable_conditions.py
@@ -14,10 +14,8 @@ be defined as environment variables outside of this command.
 
 \b
 Environment variables required:
-    * SLACK_WEBHOOK_REPORTING_GENERAL: incoming webhook URL for sending Slack
-        messages to the Seattle Flu Study #reporting-general channel
-    * SLACK_WEBHOOK_REPORTING_CHILDRENS: incoming webhook URL for sending Slack
-        messages to the Seattle Flu STudy #reporting-childrens channel
+    * SLACK_WEBHOOK_REPORTING_HCOV19: incoming webhook URL for sending Slack
+        messages to the Seattle Flu Study #ncov-reporting channel
 """
 import os
 import json
@@ -65,11 +63,7 @@ def notify(*, action: str):
         return os.environ.get("SLACK_WEBHOOK_ALERTS_TEST") \
             or os.environ[f"SLACK_WEBHOOK_REPORTING_{suffix}"]
 
-    SLACK_WEBHOOK_REPORTING_GENERAL = webhook("GENERAL")
-    SLACK_WEBHOOK_REPORTING_CHILDRENS = webhook("CHILDRENS")
     SLACK_WEBHOOK_REPORTING_HCOV19 = webhook("HCOV19")
-
-    childrens_sites = get_childrens_sites(db)
 
     # Fetch and iterate over reportable condition records that aren't processed
     #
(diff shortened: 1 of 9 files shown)


joverlee521:
[hCoV-19 Slack alerts](https://trello.com/c/Pxs9mTou/406-hcov-19-slack-alerts)

schema/deploy/shipping/views.sql
```diff
@@ -13,37 +13,36 @@ begin;
...
-        select array_agg(lineage) as lineages
-        from warehouse.organism
-        where details @> '{"report_to_public_health":true}'
-    )
-
     select
         presence_absence.presence_absence_id,
         organism.lineage::text,
         sample.identifier as sample,
-        barcode,
```
kairstenfay: I think we should bump the revision number because we're removing existing columns from the view.

schema/deploy/shipping/views.sql
```diff
@@ -13,37 +13,36 @@ begin;
...
+    join warehouse.identifier sample_id on (sample.identifier = cast(sample_id.uuid as text))
+    join warehouse.identifier collection_id on (sample.collection_identifier = cast(collection_id.uuid as text))
     left join warehouse.encounter using (encounter_id)
     left join warehouse.site using (site_id)
 
-    where organism.lineage <@ (table reportable)
+    where organism.lineage <@ 'Human_coronavirus.2019'::ltree
     and present
+    -- We can only report on prospective samples
+    and sample.details ->> 'sample_origin' in ('kiosk', 'mail')
```
kairstenfay: This seems fine but may require more manual updating if we get more sample_origin sites added. A couple other options (not necessarily robust to manual updates):
* exclude all `sample_origin` != '*_retro' (or explicitly list known retros, `sch_retro`, `uwmc_retro`, `hmc_retro`, `nwh_retro`)
* exclude all where `swab_site` is null (I believe this is only for retros) plus those where `sample_origin` == 'es' (environmental sample)

Side note: I see a couple new `sample_origin` labels in there: 'critical' and 'critical_sample'. Not sure what they mean.

schema/deploy/shipping/views.sql
joverlee521: Since we're the only consumers of this view I didn't think the version number matters

schema/deploy/shipping/views.sql
joverlee521: I'd rather be explicit about what _is_ allowed so we don't accidently leak any results we are not supposed to export.

joverlee521: I asked Peter about `critical_sample` on [slack](https://seattle-flu-study.slack.com/archives/CCAEWSFTK/p1584656638083700?thread_ts=1584645092.080500&cid=CCAEWSFTK). They are internal labels for the lab and were supposed to be scrubbed

schema/deploy/shipping/views.sql
kairstenfay: We do have at least two Metabase queries that use this view.

joverlee521: ohhh right...I guess we'll keep v1 and I'll update to v2

joverlee521: Actually I'll just rename to `reportable_hcov19_v2` since this new view is specific to hCoV19

tsibley: > We do have at least two Metabase queries that use this view.

@kairstenfay Which queries? I didn't see these when I went searching.

schema/deploy/shipping/views.sql
tsibley: I concur with Jover re: being explicit about what is allowed.

In order not to solely rely on the manifest data being accurate though, I do wonder about making this:

```sql
collection_identifier_set.name in (…)
```

where we list prospective identifier sets. Thoughts?

schema/deploy/shipping/views.sql
```diff
@@ -930,4 +930,59 @@ comment on view shipping.scan_return_results_v1 is
...
+        sample.details->>'sample_origin' as sample_origin,
+        sample.details->>'swab_site' as swab_site,
+        encounter.details ->> 'language' as language
+
+    from warehouse.presence_absence
+    join warehouse.target using (target_id)
+    join warehouse.organism using (organism_id)
+    join warehouse.sample using (sample_id)
+    join warehouse.identifier sample_id on (sample.identifier = cast(sample_id.uuid as text))
+    join warehouse.identifier collection_id on (sample.collection_identifier = cast(collection_id.uuid as text))
```
tsibley: This join requires the sample to have a collection identifier, and although I expect that to always be true, we probably shouldn't assume lest we miss reportables due to a manifest data problem. (Because p/a results can come in for a sample without a collection identifier.)

schema/deploy/shipping/views.sql
```diff
@@ -930,4 +930,59 @@ comment on view shipping.scan_return_results_v1 is
...
 
 
+create or replace view shipping.reportable_hcov19_v1 as
+
+    select
+        presence_absence.presence_absence_id,
+        organism.lineage::text,
+        sample.identifier as sample,
+        sample_id.barcode as sample_barcode,
+        collection_id.barcode as collection_barcode,
```
tsibley: Izzy et al. need two identifiers for CLIA, but I believe the two identifiers they need are specifically:

1. The collection barcode (e.g. from `collections-scan`).
2. The extra compliance barcode (e.g. from `collections-clia-compliance`). This should be found in `sample.details->>'scan_barcode'` now, I think.

Both of these are in REDCap and must both be cross-checked. The sample barcode isn't assigned until the lab processes the sample, and that information is not in REDCap AFAIK...

I'll double-check on this with Izzy/Misja/Peter/etc.

lib/seattleflu/id3c/cli/command/reportable_conditions.py
```diff
@@ -79,7 +73,7 @@ def webhook(suffix):
 
     reportable_conditions = db.cursor("reportable_conditions")
     reportable_conditions.execute("""
-        select reportable_condition_v1.*, presence_absence_id as id
+        select reportable_hcov19_v1.*, presence_absence_id as id
             from shipping.reportable_condition_v1
```
tsibley: The select-list is using the new view name but the from clause is using the old view name (although in my other comments I think it should stay that way). This generates a SQL error from Pg.

schema/deploy/shipping/views.sql
tsibley: The SCAN REDCap project's reporting instrument suggests this as well.

tsibley: Asked in Slack: https://seattle-flu-study.slack.com/archives/GUUDWB9BN/p1584946163113500

tsibley:
My thinking here with tweaking the `reportable_condition_v1` view vs. creating a new one:

1. We can't report on retrospectives, only prospectives, so that additional query constraint is ok to add to the existing view.

2. We can't report _at all_ on anything but hCoV-19, so instead of adjusting the view's query, we should instead remove the `report_to_public_health` flag on all other organisms.

3. The existing `barcode` field could/should become collection barcode. I don't know why we were using sample barcode; that just added an additional lookup step for Izzy! orz

4. The new CLIA barcode can be an extra new column.

schema/deploy/shipping/views.sql
kairstenfay: I'm sorry, I thought the linked Metabase queries in the Slack messages used the `shipping.reportable_condition_v1` view. I guess I remembered incorrectly.

kairstenfay:
>     2. Right now we can't report _at all_ on anything but hCoV-19, so instead of adjusting the view's query, we should instead remove the `report_to_public_health` flag on all other organisms. When other organisms become CLIA-certified in the future, they can be re-flagged as reportable then.

Removing `report_to_public_health` from all other organisms and making fewer adjustments on the existing view's query makes a lot of sense to me. 

>     3. The existing `barcode` field could/should become collection barcode. I don't know why we were using sample barcode; that just added an additional lookup step for Izzy! orz

This was the part that made me think we should create a `v2`. But as we discussed above, @joverlee521's intuition was right, and apparently nothing else uses this view, not even Metabase. So I am fine with keeping view's name/version the same and making adjustments in place.

schema/deploy/shipping/views.sql
joverlee521: @tsibley So would this be 
```
((collection_identifier_set.name is null or collection_identifier_set.name in (...))
or (sample.details ->> 'sample_origin' in ('kiosk', 'mail')))
```

schema/deploy/shipping/views.sql
joverlee521: Can we even report these results back to participants if there are not two identifiers to verify?

tsibley:
Repushed to fix a SQL syntax error:

```diff
-    and collection_id_set.name  == 'collections-scan'
+    and collection_id_set.name = 'collections-scan'
```

and while I was at it corrected a couple typos in commit messages. Ok to merge and deploy!

PR #75, opened 2020-03-23
Ingest SCAN encounter without sample

joverlee521:
Removed 3 instruments from the required instruments list that were only required to get the
sample barcode. 

If a sample has not been received yet, then we only create encounter, patient, questionnaire response for the record. Later on the sample will be created once the record has been updated with `post_collection_data_entry_qc` complete.

(Also, the instrument names have been changed at some point??? 🤷‍♀ )


joverlee521:
[Ingest data from SCAN screening questionnaire before smaple received by lab (e.g. early REDCap metadata)](https://trello.com/c/VbQXhRCV/430-ingest-data-from-scan-screening-questionnaire-before-smaple-received-by-lab-eg-early-redcap-metadata)

PR #77, opened 2020-03-24
Link new targets

joverlee521:
This includes new Snomed targets from UW retro pulls and the two new COVID targets from NWGC. 

I want to point out that the UW retros include hCoV-19 results. I don't think we want to be including these in our analyses/reporting?

### DIFF
--- a/schema/deploy/warehouse/target/data.sql
+++ b/schema/deploy/warehouse/target/data.sql
@@ -80,6 +80,8 @@ with target_lineage (identifier, lineage) as (
         ('CoV_229E_CoV_OC43',           'Human_coronavirus'),
         ('CoV_HKU1_CoV_NL63',           'Human_coronavirus'),
         ('COVID-19',                    'Human_coronavirus.2019'),
+        ('COVID-19_Orf1b',              'Human_coronavirus.2019'),
+        ('COVID-19-S_gene',             'Human_coronavirus.2019'),
         ('nCoV',                        'Human_coronavirus.2019'),
         ('CoV_HKU1',                    'Human_coronavirus.HKU1'),
         ('CoV_NL63',                    'Human_coronavirus.NL63'),
--- a/schema/deploy/warehouse/target/data@2020-03-24b.sql
+++ b/schema/deploy/warehouse/target/data@2020-03-24b.sql
@@ -80,6 +80,8 @@ with target_lineage (identifier, lineage) as (
         ('CoV_229E_CoV_OC43',           'Human_coronavirus'),
         ('CoV_HKU1_CoV_NL63',           'Human_coronavirus'),
         ('COVID-19',                    'Human_coronavirus.2019'),
+        ('COVID-19_Orf1b',              'Human_coronavirus.2019'),
+        ('COVID-19-S_gene',             'Human_coronavirus.2019'),
         ('nCoV',                        'Human_coronavirus.2019'),
         ('CoV_HKU1',                    'Human_coronavirus.HKU1'),
         ('CoV_NL63',                    'Human_coronavirus.NL63'),
@@ -121,13 +123,10 @@ with target_lineage (identifier, lineage) as (
         ('S. pneumoniae_APZTD4A',       'Streptococcus_pneumoniae'),
         ('S.pneumoniae',                'Streptococcus_pneumoniae'),
 
-        -- Snomed CT targets used by Ellume, Cepheid, UW Retros
-        ('http://snomed.info/id/440930009',        'Adenovirus'),
-        ('http://snomed.info/id/1240581000000104', 'Human_coronavirus.2019'),
-        ('http://snomed.info/id/181000124108',     'Influenza.A'),
-        ('http://snomed.info/id/441345003',        'Influenza.B'),
-        ('http://snomed.info/id/440925005',        'Rhinovirus'),
-        ('http://snomed.info/id/441278007',        'RSV')
+        -- Targets reported by Ellume & Cepheid using SNOMED CT
+        ('http://snomed.info/id/181000124108', 'Influenza.A'),
+        ('http://snomed.info/id/441345003',    'Influenza.B'),
+        ('http://snomed.info/id/441278007',    'RSV')
 )
 
 insert into warehouse.target (identifier, organism_id, control)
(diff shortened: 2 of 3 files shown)


joverlee521:
[Link new targets to organisms](https://trello.com/c/FNzWXPNV/435-link-new-targets-to-organisms)

kairstenfay:
> I want to point out that the UW retros include hCoV-19 results. I don't think we want to be including these in our analyses/reporting?

My understanding is that we do want to ingest hCoV-19 retrospective results but not in a way that's reportable (e.g. not in the reportable conditions view or the return of results view). I think we do want to include those results in the `hcov19_observation_v1` view.

joverlee521:
> > I want to point out that the UW retros include hCoV-19 results. I don't think we want to be including these in our analyses/reporting?
> 
> My understanding is that we do want to ingest hCoV-19 retrospective results but not in a way that's reportable (e.g. not in the reportable conditions view or the return of results view). I think we do want to include those results in the `hcov19_observation_v1` view.

Hmm, I thought `hcov19_observation_v1` is completely for analyzing hCoV results produced by SFS or SCAN.

kairstenfay:
> Hmm, I thought `hcov19_observation_v1` is completely for analyzing hCoV results produced by SFS or SCAN.

it's a good point to raise that some people may be operating under that assumption, but I was not when I created the view.

schema/deploy/warehouse/target/data.sql
```diff
@@ -80,6 +80,8 @@ with target_lineage (identifier, lineage) as (
         ('CoV_229E_CoV_OC43',           'Human_coronavirus'),
         ('CoV_HKU1_CoV_NL63',           'Human_coronavirus'),
         ('COVID-19',                    'Human_coronavirus.2019'),
+        ('COVID-19_Orf1b',              'Human_coronavirus.2019'),
+        ('COVID-19-S_gene',             'Human_coronavirus.2019'),
```
kairstenfay: Is the plan now to drop this commit because we're deleting the affected records?

schema/deploy/warehouse/target/data.sql
```diff
@@ -121,10 +121,13 @@ with target_lineage (identifier, lineage) as (
...
-        ('http://snomed.info/id/181000124108', 'Influenza.A'),
-        ('http://snomed.info/id/441345003',    'Influenza.B'),
-        ('http://snomed.info/id/441278007',    'RSV')
+        -- Snomed CT targets used by Ellume, Cepheid, UW Retros
+        ('http://snomed.info/id/440930009',        'Adenovirus'),
+        ('http://snomed.info/id/1240581000000104', 'Human_coronavirus.2019'),
+        ('http://snomed.info/id/181000124108',     'Influenza.A'),
+        ('http://snomed.info/id/441345003',        'Influenza.B'),
+        ('http::/snomed.info/id/440925005',        'Rhinovirus'),
+        ('http://snomed.info/id/441278007',        'RSV')
```
kairstenfay: Thanks for doing this!

schema/deploy/warehouse/target/data.sql
```diff
@@ -121,10 +121,13 @@ with target_lineage (identifier, lineage) as (
...
-        -- Targets reported by Ellume & Cepheid using SNOMED CT
-        ('http://snomed.info/id/181000124108', 'Influenza.A'),
-        ('http://snomed.info/id/441345003',    'Influenza.B'),
-        ('http://snomed.info/id/441278007',    'RSV')
+        -- Snomed CT targets used by Ellume, Cepheid, UW Retros
+        ('http://snomed.info/id/440930009',        'Adenovirus'),
+        ('http://snomed.info/id/1240581000000104', 'Human_coronavirus.2019'),
+        ('http://snomed.info/id/181000124108',     'Influenza.A'),
+        ('http://snomed.info/id/441345003',        'Influenza.B'),
+        ('http::/snomed.info/id/440925005',        'Rhinovirus'),
```
kairstenfay: typo on line 129; should be `http://`

joverlee521: thanks for catching that 🤦‍♀

schema/deploy/warehouse/target/data.sql
joverlee521: I think we still want to link them in case for whatever reason they get pushed to ID3C again. That way these results are hidden from public view...

schema/deploy/warehouse/target/data.sql
```diff
@@ -121,10 +121,13 @@ with target_lineage (identifier, lineage) as (
         ('S. pneumoniae_APZTD4A',       'Streptococcus_pneumoniae'),
         ('S.pneumoniae',                'Streptococcus_pneumoniae'),
 
-        -- Targets reported by Ellume & Cepheid using SNOMED CT
-        ('http://snomed.info/id/181000124108', 'Influenza.A'),
-        ('http://snomed.info/id/441345003',    'Influenza.B'),
-        ('http://snomed.info/id/441278007',    'RSV')
+        -- Snomed CT targets used by Ellume, Cepheid, UW Retros
+        ('http://snomed.info/id/440930009',        'Adenovirus'),
+        ('http://snomed.info/id/1240581000000104', 'Human_coronavirus.2019'),
```
kairstenfay: Ugghh I forgot this SNOMED id is specific to the UK SNOMED system, not the international one. there's a comment about it in the `redcap-det uw-retrospectives` command, but I forgot it ended up creating a full URL given the SNOMED id... Do you think this is fine to leave for now, or should the `redcap-det` command be updated to use a different URL?

schema/deploy/warehouse/target/data.sql
tsibley: Agreed.

schema/deploy/warehouse/target/data.sql
tsibley: It should be updated, but let's leave it for now unless you don't think it'll take long. None of our other refs make the SNOMED edition or version explicit either.

joverlee521:
oh my, just realized this wasn't even a full sqitch rework 🤦‍♀

tsibley:
Haha, oh my. Mmm, the rework you just pushed though is not quite right, I think, as it used your existing changes when copying the old files to `views@….sql`.

PR #78, opened 2020-03-25
Add "kp" regex to site details

kairstenfay:
The specimen manifest sheet labels Kaiser Permanent data's sample origin
as "kp". Add this to our site.details column.


kairstenfay:
[Add "kp" regex to warehouse.site.details](https://trello.com/c/F2XteMM8/445-add-kp-regex-to-warehousesitedetails)

PR #79, opened 2020-03-25
redcap-det uw-retrospectives: Add discharge disposition

kairstenfay:
Add an additional discharge disposition key to the UW Retrospective
REDCap-DET ETL.


kairstenfay:
[Update uw-retrospectives discharge dispositions](https://trello.com/c/U8N1RdHa/447-update-uw-retrospectives-discharge-dispositions)

PR #80, opened 2020-03-26
views: Drop individual COVID-19 targets from SCAN RoR

kairstenfay:
We shouldn't be receiving individual COVID-19 targets from Samplify, but
it still happens on occasion. Make sure we don't return inaccurate
results to participants by blocking these targets from the view
entirely.


kairstenfay:
[Stop returning individual COVID-19 targets in scan_return_results_v1 view](https://trello.com/c/DuWFfmqB/437-stop-returning-individual-covid-19-targets-in-scanreturnresultsv1-view)

kairstenfay:
Rebased onto master and updated sqitch tags

PR #81, opened 2020-03-26
etl redcap-det scan: add assertions to check barcodes match

joverlee521:
We've run into records where the sample collection barcode from
different fields don't match. This will alert us if this happens
again and prevent ingestion of the wrong barcode.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -316,24 +316,7 @@ def create_resource_condition(record: dict, symptom_name: str, patient_reference
 
 def create_specimen(record: dict, patient_reference: dict) -> tuple:
     """ Returns a FHIR Specimen resource entry and reference """
-    def specimen_barcode(record: Any) -> str:
-        """
-        Given a REDCap *record*, returns the barcode or corrected barcode if it
-        exists.
-        """
-        barcode = record['return_utm_barcode'] or record['pre_scan_barcode']
-
-        if not barcode:
-            barcode = record['utm_tube_barcode_2']
-            reentered_barcode = record['reenter_barcode']
-
-            if record['barcode_confirm'] == "No":
-                #TODO: Figure out why 'corrected_barcode' doesn't always exist?
-                barcode = record.get('corrected_barcode')
-
-        return barcode
-
-    barcode = specimen_barcode(record)
+    barcode = record['return_utm_barcode']
     if not barcode:
         LOG.warning("Could not create Specimen Resource due to lack of barcode.")
         return None, None


joverlee521:
[Add assertion to check all barcodes match to etl redcap-det scan](https://trello.com/c/TE1QslDn/448-add-assertion-to-check-all-barcodes-match-to-etl-redcap-det-scan)

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -318,18 +318,34 @@ def create_specimen(record: dict, patient_reference: dict) -> tuple:
...
+        barcode_fields = ['return_utm_barcode', 'pre_scan_barcode', 'utm_tube_barcode_2', 'reenter_barcode']
+
+        for field in barcode_fields:
+            record[field] = record[field].strip().lower()
+
+        record_id = record['record_id']
+        barcode = record['return_utm_barcode']
+        pre_scan_barcode = record['pre_scan_barcode']
+
+        if record_id not in known_problem_record_ids and barcode and pre_scan_barcode:
```
kairstenfay: Why do we want to raise an AssertionError if the barcodes don't match but also ingest records with known mismatched barcodes? Is this just because the known barcodes have already entered our system?

joverlee521: The `return_utm_barcode` is what the lab see's so that should be the true barcode. Once we confirm that is true, add to the known list of the problem record ids.

joverlee521:
With how many barcode errors we've seen via scan return of results, I'm not sure we want to include this assertion anymore. We'd be dealing with the same problem in two separate processes...

tsibley:
> With how many barcode errors we've seen via scan return of results, I'm not sure we want to include this assertion anymore. We'd be dealing with the same problem in two separate processes...

Good to consider, yeah. Would be nice not to have two places, but I think still ideally we'd keep some check like this to avoid ingesting bad data further down the pipeline (where it would still likely require manual intervention, just in a different place).

We could, in this ETL, totally ignore the participant-entered barcodes and only accept the `return_utm_barcode` or `pre_scan_barcode`, which would reduce bookkeeping to just those (which are hopefully less frequent).

joverlee521:
> > With how many barcode errors we've seen via scan return of results, I'm not sure we want to include this assertion anymore. We'd be dealing with the same problem in two separate processes...
> 
> Good to consider, yeah. Would be nice not to have two places, but I think still ideally we'd keep some check like this to avoid ingesting bad data further down the pipeline (where it would still likely require manual intervention, just in a different place).
> 
> We could, in this ETL, totally ignore the participant-entered barcodes and only accept the `return_utm_barcode` or `pre_scan_barcode`, which would reduce bookkeeping to just those (which are hopefully less frequent).

A majority of the problem seems to be the `pre_scan_barcode` field. Can we just ingest `return_utm_barcode` since that is the barcode that the lab sees?

kairstenfay:
> A majority of the problem seems to be the `pre_scan_barcode` field. Can we just ingest `return_utm_barcode` since that is the barcode that the lab sees?

That sounds good to me. If we find that participants are repeatedly inquiring about incorrectly entered barcodes, we can always retroactively ingest them in some way.

tsibley:
> A majority of the problem seems to be the `pre_scan_barcode` field. Can we just ingest `return_utm_barcode` since that is the barcode that the lab sees?

Yes, with the understanding that this will mean we can't create a sample record until the lab receives it (that's fine).

PR #83, opened 2020-03-28
etl redcap-det scan: Ingest 'able_to_test'

kairstenfay:
If a sample arrives in a condition that is unable to be tested, it will
be marked never_tested at unboxing.

If later during lab testing/processing it falls into the never_tested
category (such as a PCR fails), the REDCap record will be updated w/
"no" for the able_to_test variable.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -256,7 +256,8 @@ def create_specimen_resource(specimen_identifier: List[dict],
                              patient_reference: dict,
                              specimen_type: str,
                              received_datetime: str = None,
-                             collection_datetime: str = None) -> dict:
+                             collection_datetime: str = None,
+                             note: str = None) -> dict:
     """
     Create specimen resource following the FHIR format
     (http://www.hl7.org/implement/standards/fhir/specimen.html)
@@ -275,6 +276,9 @@ def create_specimen_resource(specimen_identifier: List[dict],
         specimen_resource["collection"] = {
             "collectedDateTime": collection_datetime
         }
+
+    if note:
+        specimen_resource["note"] = [{"text": note}]
 
     return specimen_resource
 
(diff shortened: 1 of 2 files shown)


kairstenfay:
[UPDATED -- ingest new `able_to_test` variable (PREVIOUSLY: "never-tested": when does this situation arise in REDCap? How do we determine if something is untestable?)](https://trello.com/c/1mOk8jxj/431-updated-ingest-new-abletotest-variable-previously-never-tested-when-does-this-situation-arise-in-redcap-how-do-we-determine-if-s)

joverlee521:
Hmm, I've been thinking whether this should be in Specimen rather than in QuestionnaireResponse. 
Specimen has two fields where this may fit: [condition](https://www.hl7.org/fhir/specimen-definitions.html#Specimen.condition) and [note](https://www.hl7.org/fhir/specimen-definitions.html#Specimen.note). 

I'm not sure it's worth digging into those now, but I think this `able_to_test` field should eventually live in `sample.details`

kairstenfay:
> Hmm, I've been thinking whether this should be in Specimen rather than in QuestionnaireResponse.
> Specimen has two fields where this may fit: [condition](https://www.hl7.org/fhir/specimen-definitions.html#Specimen.condition) and [note](https://www.hl7.org/fhir/specimen-definitions.html#Specimen.note).
> 
> I'm not sure it's worth digging into those now, but I think this `able_to_test` field should eventually live in `sample.details`

Interesting. It looks like `condition` is more about the sample's status in processing. It looks like we'd have to create our own coding system to capturing "never-tested", "unable-to-test" or whatever we would code it on our end. 

`note` might be better because it accepts free text, and we're not actually interested (at this point) in tracking samples' statuses in real time like `condition` might be used for.

kairstenfay:
@joverlee521 I think your suggestion of placing it somewhere on a Specimen makes sense, because we'd want to ingest this as a sample detail, not something related to the encounter.

kairstenfay:
@joverlee521 I repushed, this time adding "never-tested" to Specimen.note iff `able_to_test` (from REDCap) == 'no'.

kairstenfay:
[UPDATED -- ingest new `able_to_test` variable (PREVIOUSLY: "never-tested": when does this situation arise in REDCap? How do we determine if something is untestable?)](https://trello.com/c/1mOk8jxj/431-updated-ingest-new-abletotest-variable-previously-never-tested-when-does-this-situation-arise-in-redcap-how-do-we-determine-if-s)

PR #87, opened 2020-03-30
Include `apartment_number` field in address

joverlee521:
This was a new field added to make the shipping process easier.
We need to ingest it to create the correct hashed address identifier.

I've tested locally and the canonicalized address returned by SmartyStreets __does__ include
the apartment number.


joverlee521:
[Include new apartment_number field in address](https://trello.com/c/Axjcyq6Y/474-include-new-apartmentnumber-field-in-address)

joverlee521:
I had to update the REVISION number because the Specimen change merged yesterday had changed the revision number on master.

PR #88, opened 2020-03-31
views: Add "never-tested" to SCAN RoR

kairstenfay:
Now that 'never-tested' lives under "note" in sample.details, add this
status code to our SCAN return of results view.


kairstenfay:
[Use `able_to_test` as `never-tested` in SCAN RoR view](https://trello.com/c/Qf6GyDm3/465-use-abletotest-as-never-tested-in-scan-ror-view)

kairstenfay:
[Use `able_to_test` as `never-tested` in SCAN RoR view](https://trello.com/c/Qf6GyDm3/465-use-abletotest-as-never-tested-in-scan-ror-view)

kairstenfay:
There are no more blockers here, as the PDF generation has been updated to accept `never-tested` status codes.

joverlee521:
While reviewing this PR I realized there's a logic flaw in our REDCap DET ETL where if samples are marked `never-tested` then later updated to `can-test` this never gets updated in `sample.details`. Not sure if this is a blocker for this specific PR, but would like to fix that bug before we report `never-tested` to participants.

kairstenfay:
Rebased onto master.

kairstenfay:
Rebased again onto master.

PR #89, opened 2020-04-01
Update Samples & hCoV19 views

joverlee521:
Updated both `shipping.sample_with_best_available_encounter_data_v1` and `shipping.hcov19_observation_v1` to include more SCAN samples. 

1. Removed where filter from `shipping.sample_with_best_available_encounter_data_v1` so that samples without an identifier are included (this mostly happens for SCAN samples that have been ingested through REDCap but hasn't been aliquoted yet)

2. In `shipping.hcov19_observation_v1` allow `sample.details ->> sample_origin` to be null, which again includes SCAN samples from REDCap that haven't been aliquoted yet.


joverlee521:
[Remove filter from `shipping.sample_with_best_available_encounter_data_v1`](https://trello.com/c/E4HmgDzU/494-remove-filter-from-shippingsamplewithbestavailableencounterdatav1)

kairstenfay:
Don't forget to tag schema

schema/deploy/shipping/views.sql
```diff
@@ -727,7 +727,6 @@ create or replace view shipping.sample_with_best_available_encounter_data_v1 as
   left join samples_with_manifest_data using (sample_id)
   left join site_details on (site_manifest_details similar to manifest_regex)
   left join warehouse.site on (samples_with_manifest_data.site_id = site.site_id)
-  where sample.identifier is not null
```
kairstenfay, line 730: I finally found [the context for this filter](https://github.com/seattleflu/id3c-customizations/pull/33#pullrequestreview-361337238). You suggested adding a collection identifier column and it looks like I didn't totally get it then. But I think that might be nice to have now.

joverlee521: Ohhhhhhh, right. I can add collection identifier to this view then.

joverlee521: Adding a column means we'd have to bump the version number right?

kairstenfay: No, only taking away existing columns or substantially changing what's being returned would warrant a version bump.

joverlee521: But adding a column breaks the REVERT script...I think this is why I'm having so much trouble with when to bump the version number

kairstenfay: Oh, right. There is the weird workaround I've been doing in the revert script where I drop then re-create the view afterward. But then we may also have to drop the views that are dependent on this view before re-creating it.

I think it can technically be done without bumping the version number, but it does require more `revert` script manipulation than we'd typically like to do.

kairstenfay, line 730: I wanted to start a separate thread on this filter removal:

Just wanted to confirm that by removing this, we're increasing both the number of samples with and without encounter data returned by this view. Is that desirable?

joverlee521: I'm not sure what you mean. I thought this view was to provide the best available metadata for all samples in `warehouse.samples`

kairstenfay: Ok, that seems fine to me.

tsibley: I'd say don't worry about adding `collection_identifier` to `shipping.sample_with_best_available_encounter_data_v1` for now. It has `sample_id`, which is what matters most.

You may know this, but in case not: the reason the revert script needs to be updated is because `create or replace view` only works if you don't change the name, type, or order of any existing columns. So with `create or replace view` you can add new columns after the existing columns or change the values in an existing column (using the same name and type), but to insert a new column amidst the existing columns, you must do `drop + create`, which then has to cascade to dependent views. One pattern for managing this, which we're not doing yet, is to always drop all views at the top of the file and then issue the creates, i.e. drop+recreate on every deploy even if you technically don't need to.

PR #90, opened 2020-04-01
etl redcap-det uw-retrospectives: Add encounter status

kairstenfay:
Add a new encounter status code to our mapping logic.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
@@ -362,6 +362,7 @@ def create_encounter_status(redcap_record: dict) -> str:
 
     mapper = {
         'arrived': 'arrived',
+        'preadmit': 'arrived',
         'discharged': 'finished',
         'lwbs': 'cancelled',  # LWBS = left without being seen.
     }


lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -362,6 +362,7 @@ def create_encounter_status(redcap_record: dict) -> str:
 
     mapper = {
         'arrived': 'arrived',
+        'preadmit': 'arrived',
```
kairstenfay, line 365: 'arrived' seemed like the best Encounter.status code here, but I'm open to suggestions.
https://www.hl7.org/fhir/valueset-encounter-status.html

kairstenfay:
[Add new encounter code for uw-retrospectives](https://trello.com/c/96zMHoKl/495-add-new-encounter-code-for-uw-retrospectives)

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
tsibley: Yes, I agree. A "preadmit" is generally a planned admission, and while there's a `planned` code, obviously the encounter has already started if we have a swab.

It's a little weird because these are encounter lifecycle statuses that don't have a place in our data model. We don't use these anyway though, right? Just producing them for spec compliance?

kairstenfay: Right, we don't currently use these Encounter.status codes downstream. We are adding them primarily for spec compliance, but also because the clinic team or modelers may eventually ask for an Encounter's dischargeDisposition. Previously we marked all Encounter.status codes as "finished", but once we started ingesting dischargeDisposition from UW retros, I started dynamically adding Encounter.status to make sure the codes aligned properly.

PR #91, opened 2020-04-02
views: Deduplicate hCoV-19 results consistently

kairstenfay:
Instead of alternating between deduplicate potentially multiple hCoV-19
tests results by latest presence_absence_id or prioritizing presence
as t > f > null, choose one method (presence_absence_id) and stick with
it.

The latest results from Samplify (BBI, NWGC) should be the final result.
Note we are also following this logic for UW clinical tests. However, in
our UW-Retrospectives REDCap-DET ETL, we're not ingesting multiple test
results for an encounter.


kairstenfay:
[Deduplicate multiple hCoV-19 results consistently](https://trello.com/c/7tovhzfX/484-deduplicate-multiple-hcov-19-results-consistently)

tsibley:
I rebased this onto master so it would be deployable on the current database. Doing my diff test again, the only changes are the minor date shift and corresponding minor Ct shifts (though fewer of these than dates).

@kairstenfay If the rebase looks good to you in the morning, please merge and deploy!

PR #92, opened 2020-04-03
etl redcap-det scan: Ingest follow-up symptoms

kairstenfay:
We're now asking SCAN study participants in a new, follow-up
questionnaire what their symptoms are and when the symptoms started. I
think this is only supposed to be used for participants in the
asymptomatic study arm who have started showing symptoms since
enrollment, but there are symptomatic study arm participants who have
also been asked these questions.

Ingest these new questions: `symptoms_2` and `symptom_duration_2`, by
reworking the existing symptom traversal code to take an optional
variable suffix (e.g. '_2').


kairstenfay:
[Update REDCap ETL to ingest new SCAN variables](https://trello.com/c/IshP2jRE/501-update-redcap-etl-to-ingest-new-scan-variables)

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -308,8 +318,12 @@ def create_resource_condition(record: dict, symptom_name: str, patient_reference
...
     }
 
-    if record["symptom_duration"]:
-        condition["onsetDateTime"] = record["symptom_duration"]
+    symptom_duration = record.get(f'symptom_duration{suffix}')
+
+    # Defend against known data quality issue in REDCap where symptom duration
+    # dates can come from the future.
+    if symptom_duration and datetime.strptime(symptom_duration, '%Y-%m-%d') < datetime.now():
+        condition["onsetDateTime"] = symptom_duration
```
kairstenfay, line 326: on second thought, this may be better as a separate commit, since it will affect `symptoms` and `symptom_duration` (not just `*_2`).

happy to split it out if there's a preference

kairstenfay: we're no longer doing this.

kairstenfay:
> LGTM! Have you tested this all the way through to make sure it updates the encounter details as expected?

Yep! It adds to the existing conditions rather than replacing them.

PR #93, opened 2020-04-03
etl redcap-det scan-en: include Specimen.note for `yes`/`can-test`

joverlee521:
When we were only including Specimen.note for `no` answers, this prevented updates for when it gets changed from `no` to `yes` (since we concatenate `sample.details`)

Noticed this because currently we have more `never-tested` samples within `warehouse.sample` than records marked `able_to_test: no` in REDCap.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -336,11 +336,9 @@ def create_specimen(record: dict, patient_reference: dict) -> tuple:
 
     if record['able_to_test'] == 'no':
         note = 'never-tested'
-    # This is based on values currently in the SCAN REDCap project,
-    # I'm guessing `can-test` was filled in at some point before the answer
-    # code was changed.
-    #   -Jover, 3 April 2020
-    elif record['able_to_test'] in {'yes', 'can-test'}:
+    # Assumes that all samples can be tested unless explicitly marked "no".
+    #   - Jover 09 April 2020
+    else:
         note = 'can-test'
 
     specimen_type = 'NSECR'  # Nasal swab.  TODO we may want shared mapping function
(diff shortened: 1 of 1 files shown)


joverlee521:
[REDCap ETL SCAN `never-tested` bug](https://trello.com/c/0JqcHjkT/504-redcap-etl-scan-never-tested-bug)

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -332,7 +332,16 @@ def create_specimen(record: dict, patient_reference: dict) -> tuple:
...
-    note = 'never-tested' if record['able_to_test'] == 'no' else None
+    note = None
+
+    if record['able_to_test'] == 'no':
+        note = 'never-tested'
+    # This is based on values currently in the SCAN REDCap project,
+    # I'm guessing `can-test` was filled in at some point before the answer
+    # code was changed.
+    #   -Jover, 3 April 2020
+    elif record['able_to_test'] in {'yes', 'can-test'}:
```
kairstenfay, line 343: Ah, unrestricted text variables in REDCap...

tsibley: Hrm. Something else is maybe going on here too, which would be good to understand I think. I currently see this distribution of `able_to_test` values as of right now:

image


(The last two are the difference between the JSON values `""` and `null` as generated by REDCap. I have no idea if that's a meaningful distinction or not, but for our purposes I think it's the same.)

Ahh! While writing this I just realized. The values I was looking at are the answer _labels_ not the answer _codes_. If I look at codes I do indeed see what Jover saw.

image

The More You Know

tsibley: I think maybe this should be an if/else instead of an if/elif though, or at least an if/elif/else, to explicitly handle the "no value" or "no known value" cases. Otherwise the same issue this PR is addressing can occur if a record goes from yes/no → no value.

joverlee521: I think I'll go with if/else with the assumption is that a sample can be tested unless explicitly marked "no". (all samples are innocent until proven guilty)

PR #94, opened 2020-04-08
add `location.identifier` to `shipping.hcvo19_observation_v1`

joverlee521:
The `location.identifier` is a hash of the canonicalized address, so it
can act as an address identifier that can be used to identify
participants of the same household.


joverlee521:
[Add a SCAN household identifier to flag addresses from the same house in shipping views](https://trello.com/c/7dDyQckR/507-add-a-scan-household-identifier-to-flag-addresses-from-the-same-house-in-shipping-views)

schema/deploy/shipping/views.sql
```diff
@@ -1135,6 +1135,7 @@ create or replace view shipping.hcov19_observation_v1 as
         best_available_site_type as site_type,
 
         location.hierarchy->'puma' as puma,
+        location.identifier as address_identifier,
```
tsibley, line 1138: This assumes that the location row is of `scale = 'address'`, which I don't think is guaranteed for all our data sources? (It definitely isn't guaranteed by the schema and joins in this query.)

I think we could make this more defensive if/when `location.scale != 'tract'` by changing this select expression to something like

```
case location.scale
    when 'address' then location.identifier
end as address_identifier
```


This relies on the fact that we save the most specific scale, and so if we have address, it'll be the row `location` refers to in this query.

joverlee521: ohh, yup. Thanks for catching this!

PR #95, opened 2020-04-09
views: Include only CLIA samples in SCAN RoR

kairstenfay:
The SCAN return of results (RoR) view should only include
presence-absence results from CLIA-certified samples. Recent work has
been done in the presence-absence ETL to mark CLIA-certified samples
with an `assay_type` equal to 'CLIA'. Use that new key to filter results
in this view.
   
Adding this extra filter does not affect the count of status codes
currently produced in this view. However, note that if we happen to
process a SCAN sample that isn't marked as `assay_type` = 'Clia', then
the `status_code` in this view will forever read "pending".


kairstenfay:
[Make sure we're only ingesting CLIA (4 crt values) in SCAN RoR](https://trello.com/c/yXENc7An/483-make-sure-were-only-ingesting-clia-4-crt-values-in-scan-ror)

kairstenfay:
Rebased onto master

PR #98, opened 2020-04-17
Add ace to views

joverlee521:
This is a new question added in the SCAN illness questionnaire that
asks participants if they take an ACE inhibitor.

Edit: Also added `never_tested` column to the `scan_encounters_v1` view. 
(This field is only available within the SCAN REDCap project, so it made more sense to add this column to the SCAN-only view rather than add it to the general `hcov19-observation` view).


joverlee521:
[Add 'ace' to Roy's views](https://trello.com/c/SP7EHE1t/536-add-ace-to-roys-views)

joverlee521:
[Add `never-tested` to Roy's shipping views](https://trello.com/c/z7XILNvX/538-add-never-tested-to-roys-shipping-views)

PR #99, opened 2020-04-21
REDCap DET: Support multiple SCAN projects, one per language

tsibley:
Adds the following commands:

```
id3c etl redcap-det scan-es
id3c etl redcap-det scan-vi
id3c etl redcap-det scan-zh-Hans
```

Details in the commit messages. Requires https://github.com/seattleflu/id3c/pull/147.

~I've not yet fully tested this locally on a fresh copy of the database, but that's in progress. However, type checks and our (minimal) tests pass and it works as expected on an older database copy.~ Tested locally and seems good to me. Travis checks are failing because I haven't yet released the core ID3C change.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -494,7 +494,7 @@ def canonicalize_name(*parts: Iterable[str]) -> str:
 
     >>> canonicalize_name("`1234567890-=~!@#$%^&*()_+")
     '1234567890'
-    >>> canonicalize_name("qwertyuiop[]\\QWERTYUIOP{}|")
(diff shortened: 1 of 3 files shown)


tsibley:
> (I'm going to need to do a more in-depth study session of decorators nerd_face )

:-) I pushed some additional commentary in a6c8c08.

tsibley:
> One thing I want to highlight: For future changes to the SCAN ETL, we need to be aware of whether new fields to ingest are available in all projects. I expect new fields being added to the English project first then translations following on a later date.

Nod. Really good point. I expect we'll be able to conditionalize on the new field's presence, so that it starts working automatically for each project.

tsibley:
Just tested with the new Spanish enrollments and it works locally. Going to merge and deploy.

PR #100, opened 2020-04-23
Scan priority code

joverlee521:
Ingest new field `priority_code` from SCAN REDCap projects and add this field to the shipping views. 

This field is currently only available in the English project, but will be translated and added to the other projects later on.


joverlee521:
[Ingest new variable `priority_code` from SCAN REDCap](https://trello.com/c/DH4rzsVG/549-ingest-new-variable-prioritycode-from-scan-redcap)

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -24,7 +24,7 @@
 LOG = logging.getLogger(__name__)
 
 
-REVISION = 7
+REVISION = 8
```
kairstenfay, line 27: I think this revision change belongs in the previous commit.

joverlee521: Ah, yes. Thank you. Also need to bump it up to 9.

PR #102, opened 2020-04-30
Scan follow up survey

joverlee521:
Ingest follow-up survey as a second Encounter and QuestionnaireResponse resource. 
Depends on https://github.com/seattleflu/id3c/pull/149. 
See commit messages for more details.


joverlee521:
[Ingest follow up survey for SCAN REDCap](https://trello.com/c/KqwnAbeQ/530-ingest-follow-up-survey-for-scan-redcap)

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -81,35 +81,48 @@ def redcap_det_scan(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_r
...
 
     if specimen_received and not specimen_entry:
         LOG.warning("Skipping enrollment with insufficent information to construct a specimen")
         return None
 
+    follow_up_encounter_entry = None
+    follow_up_complete = is_complete('day_7_follow_up', redcap_record)
+
+    if follow_up_complete:
+        # Follow-up encounter for 7 day follow-up survey
```
kairstenfay, line 115: It doesn't look like `follow_up_complete` is used again, so you could just check for `if is_complete('day_7_follow_up', redcap_record)` without creating a new variable.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -593,3 +609,33 @@ def build_response_answers(response: Union[str, List]) -> List:
...
+        value = f"{REDCAP_URL}/{record.project.id}/{record['record_id']}_follow_up"
+    )
+    encounter_class_coding = create_coding(
+        system = "http://terminology.hl7.org/CodeSystem/v3-ActCode",
+        code = "HH"
+    )
+
+    if not record.get('fu_timestamp'):
+        return None, None
+
```
kairstenfay, line 656: This check can be done at the top of the `create_follow_up_encounter` function before creating an `encounter_identifier` or `encounter_class_coding`.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -533,6 +537,9 @@ def combine_checkbox_answers(record: dict, coded_question: str) -> Optional[List
     if coded_question == 'race':
         return race(answers)
 
+    if re.match(r'fu_[1-4]{1}_symptoms$', coded_question):
```
kairstenfay, line 540: FWIW the `{1}` is unnecessary here because by default the regex will only look for a single match for a character in the square brackets.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -653,3 +660,103 @@ def create_follow_up_encounter(record: REDCapRecord, patient_reference: dict, si
...
     return create_entry_and_reference(encounter_resource, "Encounter")
+
+
+def create_follow_up_questionnaire_response(record: dict, patient_reference: dict,
+                                            encounter_reference: dict) -> Optional[dict]:
+    """
+    Returns a FHIR Questionnaire Response resource entry for the follow-up
+    encounter.
+
+    Note: `fu_which_activites` and `fu_missed_activites` are misspelled on
```
kairstenfay, line 677: Just wanted to say I giggle every time I see the prefix on these fields is `fu`.

kairstenfay:
General comment on naming symmetry here:
If we're calling the enrollment encounter "primary", should we call the follow-up questionnaire the "secondary" encounter?
Alternatively, we could have "initial" encounter and "follow-up" encounter. 
I think it would be wise to use balanced names.
What do you think?

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
joverlee521: Thanks, that's good to know 😄

joverlee521:
> General comment on naming symmetry here:
> If we're calling the enrollment encounter "primary", should we call the follow-up questionnaire the "secondary" encounter?
> Alternatively, we could have "initial" encounter and "follow-up" encounter.
> I think it would be wise to use balanced names.
> What do you think?

I don't have any particular preference for naming here. We can go with initial and follow-up encounter.

joverlee521:
[Create shipping view for follow up encounters](https://trello.com/c/Mj6pMPa0/561-create-shipping-view-for-follow-up-encounters)

tsibley:
Looks like this and #104 both make changes to views, so whichever merges second will need their sqitch changes rebased.

tsibley:
Ah, I just raised a possible issue with a related PR, https://github.com/seattleflu/id3c/pull/149#pullrequestreview-408548083, so to avoid more work than necessary, would be good to hold off on merging this until we resolve that.

joverlee521:
> That new view, though, my god. 😳 I assume you thought about refactoring it—it looks like a combination of the existing encounters + fhir details views, with tweaks for link ids—but punted due to time/effort constraints?

Yeah, I cringed when I added this new view. I'm not entirely sure how to refactor it...

joverlee521:
I've been testing locally with the latest database dump from last Friday. This new view is not useable because of the number of fields (locally takes ~20 minutes to run select all query on the new view). This also shows that our current `shipping.fhir_encounter_details_v2` is not sustainable if the number of fields increases dramatically. 

I'm not sure if there's more advanced SQL optimization that can be done here or if we need to think about parsing the QuestionnaireResponse into a more useable format in the FHIR ETL instead of just ingesting the whole JSON into `encounter.details`.

schema/deploy/shipping/views.sql
```diff
@@ -1436,4 +1438,576 @@ grant select
...
+      left join fu_3_symptoms using (encounter_id)
+      left join fu_3_test using (encounter_id)
+      left join fu_3_result using (encounter_id)
+      left join fu_4_date using (encounter_id)
+      left join fu_4_symptoms using (encounter_id)
+      left join fu_4_test using (encounter_id)
+      left join fu_4_result using (encounter_id)
+      left join fu_healthy_test using (encounter_id)
+      left join fu_healthy_result using (encounter_id)
+      left join warehouse.encounter initial on initial.identifier = encounter.details ->> 'part_of'
```
tsibley, line 1996: I believe `encounter.details @> jsonb_build_object('part_of', initial.identifier)` would use the index on `details`, but only a query plan would tell for sure.

tsibley:
> I've been testing locally with the latest database dump from last Friday. This new view is not useable because of the number of fields (locally takes ~20 minutes to run select all query on the new view). This also shows that our current `shipping.fhir_encounter_details_v2` is not sustainable if the number of fields increases dramatically.
> 
> I'm not sure if there's more advanced SQL optimization that can be done here or if we need to think about parsing the QuestionnaireResponse into a more useable format in the FHIR ETL instead of just ingesting the whole JSON into `encounter.details`.

As discussed elsewhere, the plan is to materialize this view, or at least the longest CTE.

schema/deploy/shipping/views.sql
joverlee521: In this case `initial.identifier = encounter.details ->> 'part_of'` is much faster than `encounter.details @> jsonb_build_object('part_of', initial.identifier)`. I uploaded the query plans [here](https://gist.github.com/joverlee521/bba66984f74f6970cf11439bd2afd51c) if you're curious.

joverlee521:
Using a materialized view *significantly* improved query times 😄 

I only created index on `encounter_id` and `link_id` for the materialized view `shipping.fhir_questionnaire_responses_v1`. Not sure if we need indexes on the other columns.

joverlee521:
Ah, also just realized we will need to create specific role in ID3C to refresh the materialized view, keeping in mind that a materialized view can only be refreshed by the view owner. I found a way around this on [StackExchange](https://dba.stackexchange.com/questions/171932/postgresql-9-3-13-how-do-i-refresh-materialised-views-with-different-users)

schema/deploy/roles/materialized-view-refresher/grants.sql
```diff
@@ -0,0 +1,44 @@
...
+    returns void
+    security definer
+    language plpgsql as $$
+        declare
+            sql text;
+            message text;
+        begin
+            sql := 'refresh materialized view concurrently ' || view_name;
+            execute sql;
+            raise info 'Refreshed materialzied view %', view_name;
```
tsibley, line 26: Typo :point_up:

schema/deploy/roles/materialized-view-refresher/grants.sql
```diff
@@ -0,0 +1,44 @@
...
+create or replace function public.refresh_materialized_view(view_name text)
+    returns void
+    security definer
+    language plpgsql as $$
+        declare
+            sql text;
+            message text;
+        begin
+            sql := 'refresh materialized view concurrently ' || view_name;
+            execute sql;
```
tsibley, line 25: Because `execute` allows multiple statements to be run and the view name is not quoted, there's SQL injection vulnerability here which would let anyone with `materialized-view-refresher` run SQL as the superuser (since `postgres` is the function owner). For example, consider:

    select refresh_materialized_view('shipping.fhir_questionnaire_responses_v1; drop database production;');

:grinning: 

To fix this, we can quote `view_name` as an identifier, but will first have to change the function signature to take the schema and name separately:

    execute format('refresh materialized view concurrently %I.%I', view_schema, view_name);

joverlee521: 😬 I have to work on thinking more defensively, but cool I didn't know about `%I`

joverlee521:
Moved the materialized-view-refresher role to core ID3C in https://github.com/seattleflu/id3c/pull/151

schema/deploy/shipping/views.sql
tsibley: Thanks!

PR #106, opened 2020-05-12
etl redcap-det scan: Add `zipcode_map`

joverlee521:
The coded value for zipcodes in REDCap are arbitrary numbers instead of
the actual zipcode. This is due to branching logic causing problems
when the code is the actual zipcode. Add `zipcode_map` to convert these
codes to the real zipcode.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -38,7 +38,7 @@ PROJECTS = [
     ScanProject(21808, "ru"),
 ]
 
-REVISION = 9
+REVISION = 10
 
 REDCAP_URL = 'https://redcap.iths.org/'
 INTERNAL_SYSTEM = "https://seattleflu.org"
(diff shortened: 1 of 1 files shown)


joverlee521:
[Update SCAN REDCap DET ETLs to use coded zip codes.](https://trello.com/c/L7P6Bsfj/579-update-scan-redcap-det-etls-to-use-coded-zip-codes)

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -170,6 +170,96 @@ def locations(db: DatabaseSession, cache: TTLCache, record: dict) -> list:
...
+        '84': '98045',
+        '85': '98051',
+        '86': '98065',
+        '87': '98027',
+        '88': '98029',
+        '89': '98052',
+        '90': '98074',
+        '91': '98075',
+        '92': '98092',
+        '93': '98070'
```
tsibley, line 253: I think it's good practice to sort this sort of static list of data.

Sorting also reveals that there are gaps, which seems somewhat suspicious to me. The missing numbers are:

9
11..14
16
23..24
27
29..30
37..43
74

joverlee521: I copied this over in the exact order from the SCAN REDCap codebook. Those values just seem to not be used as codes 🤷

tsibley: Alright! :-)

PR #107, opened 2020-06-05
Add date cutoff to SCAN results

joverlee521:
The addition of the materialized view made the view dependencies more complicated, so I reorganized the view definition file to drop all views at the very top. This required the view definitions to be re-ordered and created in the order of dependency as well. Since I had to re-order them, I also organized them into broad categories to hopefully make them easier to understand. (Sorry the file diffs are a little crazy because of this...)

And of course, the actual point of this PR was to add a date cutoff to the SCAN return results view so that we only return results to participants in the SCAN _research_ study.

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -1886,6 +1886,9 @@ create or replace view shipping.scan_return_results_v1 as
         left join warehouse.encounter using (encounter_id)
       where
         identifier_set.name in('collections-scan')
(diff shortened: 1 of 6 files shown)


joverlee521:
[Update shipping.scan_return_results view](https://trello.com/c/x0PA22xT/609-update-shippingscanreturnresults-view)

joverlee521:
Holding off on deploying this because we will likely add another where clause to include "re-consented" samples for participants that had results in limbo.

PR #115, opened 2020-06-29
etl redcap-det scan: skip "false" enrollments

joverlee521:
The study only wants us to ingest "true" enrollments where a participant
was mailed a swab kit. "True" enrollments can be identified by the
participant completing the illness_questionnaire. However, there is
a bug in REDCap that sometimes leaves the instrument marked as
incomplete/unverified. To get around this, we are checking that the
`illness_q_date` is filled in to verify that the instrument is complete.
However, there is yet another bug in REDCap that sometimes leaves the
`illness_q_date` blank. To catch these records, we are also checking
that the back_end_mail_scans instrument is complete.

This leaves us with a majority of records being ingested the day of
enrollment due to the completion of the `illness_q_date` and some
straggling records ingested the day after enrollment where the
back_end_mail_scans instrument is complete.

---
TODO:
- [x] test locally after latest database dump
- [ ] delete "false" encounters currently in ID3C before deployment

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -102,7 +102,7 @@ def redcap_det_scan(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_r
     # By verifying illness_questionnaire is complete first, we minimize the
     # delay in data ingestion since the back_end_mail_scans is completed the day after enrollment.
     #   -Jover, 29 June 2020
-    if not redcap_record['illness_q_date'] and redcap_record['back_end_mail_scans_complete'] != '2':
+    if not redcap_record['illness_q_date'] and not is_complete('back_end_mail_scans', redcap_record):
         LOG.debug("Skipping incomplete enrollment")
         return None
 
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -46,7 +46,7 @@ PROJECTS = [
 
 ]
 
-REVISION = 12
+REVISION = 13
 
 REDCAP_URL = 'https://redcap.iths.org/'
 INTERNAL_SYSTEM = "https://seattleflu.org"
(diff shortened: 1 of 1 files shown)


joverlee521:
[Add `back_end_mail_scan` to required instruments?](https://trello.com/c/jzvUYTun/642-add-backendmailscan-to-required-instruments)

joverlee521:
I believe this should delete all SCAN encounters without samples:
```sql
create temp table scan_encounters_without_sample as (
    select
        encounter_id
    from warehouse.encounter
    join warehouse.site using (site_id)
    left join warehouse.sample using (encounter_id)
    where site.identifier = 'SCAN'
    and sample_id is null
);

delete
from warehouse.encounter_location
where encounter_id in (select * from scan_encounters_without_sample)
;

delete
from warehouse.encounter
where encounter_id in (select * from scan_encounters_without_sample)
;
```
Deploying the latest REVISION of this PR after running the above script should fill in all the "true" encounters that do not have a sample yet.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -93,6 +93,19 @@ def command_for_each_project(function):
...
+    # Skip record if the illness_questionnaire is not complete, because this is
+    # a "false" enrollment where the participant was not mailed a swab kit.
+    # We must verify illness_questionnaire with the `illness_q_date` field
+    # since there is a bug in REDCap that sometimes leaves the questionnaire marked incomplete/unverified.
+    # We must have another check of the back_end_mail_scans because sometimes
+    # the `illness_q_date` field is not filled in due to a bug in REDCap.
+    # By verifying illness_questionnaire is complete first, we minimize the
+    # delay in data ingestion since the back_end_mail_scans is completed the day after enrollment.
+    #   -Jover, 29 June 2020
+    if not redcap_record['illness_q_date'] and redcap_record['back_end_mail_scans_complete'] != '2':
```
tsibley, line 105: Instead of using the magic value of `2`, let's use `id3c.cli.redcap.is_complete()`?

https://github.com/seattleflu/id3c/blob/f401e64480424821d39988940f7c1cd71ea94209/lib/id3c/cli/redcap.py#L244-L265

tsibley:
> I believe this should delete all SCAN encounters without samples:

For safety, surround that in a database transaction so that if something goes wrong nothing is deleted (rather than say, the encounter locations but not encounters)?

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
joverlee521: Ah, yup. Thank you!

joverlee521:
> For safety, surround that in a database transaction so that if something goes wrong nothing is deleted (rather than say, the encounter locations but not encounters)?

So adding `begin;` at the front and running `commit;` if the changes look reasonable?

tsibley:
> So adding `begin;` at the front and running `commit;` if the changes look reasonable?

Exactly. (Sorry I missed this last week!)

PR #117, opened 2020-07-14
Views for Power BI Dashboard

joverlee521:
Changes to set up the views for the public Power BI dashboard:
1. Ingest `illness_q_date` field from REDCap - this represents the "completed enrollment" date
2. Add a `power-bi-exporter` role to the database.
3. Add new views for the dashboard:
      - shipping.scan_demographics_v1
      - shipping.scan_hcov19_positives_v1
      - shipping.seattle_neighborhood_districts_v1
      - shipping.scan_enrollments_v1

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -1153,6 +1153,7 @@ create or replace view shipping.scan_encounters_v1 as
 
         encountered,
         to_char(encountered, 'IYYY-"W"IW') as encountered_week,
+        illness_questionnaire_date,
 
         site.identifier as site,
         site.details ->> 'type' as site_type,
@@ -1172,6 +1173,7 @@ create or replace view shipping.scan_encounters_v1 as
 
         location.hierarchy -> 'puma' as puma,
         location.hierarchy -> 'tract' as census_tract,
+        location.hierarchy -> 'neighborhood_district' as neighborhood_district,
 
         symptoms,
         symptom_onset,
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -565,6 +565,13 @@ create or replace view shipping.fhir_encounter_details_v2 as
                  string_response as industry
             from shipping.fhir_questionnaire_responses_v1
           where link_id = 'industry'
+        ),
+
+        illness_questionnaire_date as (
+          select encounter_id,
+                 date_response[1] as illness_questionnaire_date
+            from shipping.fhir_questionnaire_responses_v1
+          where link_id = 'illness_q_date'
         )
 
     select
@@ -614,7 +621,8 @@ create or replace view shipping.fhir_encounter_details_v2 as
         distance,
         attend_event,
         wfh,
-        industry
+        industry,
+        illness_questionnaire_date
 
       from warehouse.encounter
       left join scan_study_arm using (encounter_id)
@@ -660,6 +668,7 @@ create or replace view shipping.fhir_encounter_details_v2 as
       left join attend_event using (encounter_id)
       left join wfh using (encounter_id)
       left join industry using (encounter_id)
+      left join illness_questionnaire_date using (encounter_id)
   ;
 comment on view shipping.fhir_encounter_details_v2 is
   'A v2 view of encounter details that are in FHIR format that includes all SCAN questionnaire answers';
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -15,6 +15,8 @@ begin;
 
 -- Drop all views at the top in order of dependency so we don't have to
 -- worry about view dependencies when reworking view definitions.
+drop view if exists shipping.scan_demographics_v1;
+
 drop view if exists shipping.scan_return_results_v1;
 drop view if exists shipping.return_results_v2;
 drop view if exists shipping.return_results_v1;
@@ -2130,5 +2132,34 @@ grant select
 
 comment on view shipping.scan_return_results_v1 is
   'View of barcodes and presence/absence results for SCAN return of results on the UW Lab Med site';
+
+
+/******************** VIEWS FOR POWER BI DASHBOARDS ********************/
+create or replace view shipping.scan_demographics_v1 as
+
+    select
+        encountered_week,
+        sex,
+        age_range_fine,
+        age_range_fine_lower,
+        age_range_fine_upper,
+        race,
+        hispanic_or_latino,
+        income,
+        puma
+    from shipping.scan_encounters_v1
+;
+
+comment on view shipping.scan_demographics_v1 is
+  'A view of basic demographic data from the SCAN project for Power BI dashboards.';
+
+revoke all
+    on shipping.scan_demographics_v1
+  from "scan-dashboard-exporter";
+
+grant select
+    on shipping.scan_demographics_v1
+    to "scan-dashboard-exporter";
+
 
 commit;
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -15,6 +15,7 @@ begin;
 
 -- Drop all views at the top in order of dependency so we don't have to
 -- worry about view dependencies when reworking view definitions.
+drop view if exists shipping.scan_hcov19_positives_v1;
 drop view if exists shipping.scan_demographics_v1;
 
 drop view if exists shipping.scan_return_results_v1;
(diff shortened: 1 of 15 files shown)


schema/deploy/shipping/views.sql
```diff
@@ -2131,4 +2147,170 @@ grant select
...
+    on shipping.scan_hcov19_positives_v1
+    to "power-bi-exporter";
+
+
+create or replace view shipping.seattle_neighborhood_districts_v1 as
+
+    select
+        identifier,
+        st_assvg(point) as centroid,
+        st_assvg(polygon) as svg_path
```
joverlee521, line 2273: I am not 100% certain that [ST_AsSVG](http://postgis.net/docs/ST_AsSVG.html) is what we want to use here.

joverlee521:
[Formative data request](https://trello.com/c/57EgipiF/658-formative-data-request)

schema/deploy/shipping/views.sql
```diff
@@ -2131,4 +2147,170 @@ grant select
...
+            and target.identifier not in ('COVID-19_Orf1b', 'COVID-19-S_gene')
+        /*
+          Keep only the most recent push. According to Lea, samples are only
+          retested if there is a failed result. A positive, negative, or
+          indeterminate result would not be retested.
+
+          https://seattle-flu-study.slack.com/archives/CV1E2BC8N/p1584570226450500?thread_ts=1584569401.449800&cid=CV1E2BC8N
+        */
+        order by
+            sample_id, presence_absence_id desc
```
jccraft-uw, line 2209: ordering by sample_id, presence_absence_id desc  ... this probably means that the IDs increase with each new record.
How about ordering by sample_id, pa.created::date or some other "result date" instead to make it super obvious that we're taking the newest record? That is, use a date instead of an ID to make it clearer that we're ordering by a date.

schema/deploy/shipping/views.sql
```diff
@@ -2131,4 +2147,170 @@ grant select
...
+          hcov19_result_release_date,
+          case
+              when puma in ('5310901', '5310902') then 'yakima'
+              when puma in ('5311701', '5311702', '5311703',
+                            '5311704', '5311705', '5311706') then 'snohomish'
+              when puma in ('5311601', '5311602', '5311603', '5311604',
+                            '5311605', '5311606', '5311607', '5311608',
+                            '5311609', '5311610', '5311611', '5311612',
+                            '5311613', '5311614', '5311615', '5311616') then 'king'
+          end as county
```
jccraft-uw, line 2228: You could add an explicit "else null" to show that null is possible. You'll still get nulls without the line, of course, but adding it could make the SQL easier to grock.

schema/deploy/shipping/views.sql
jccraft-uw: Can Gernot give any input? Will he be able to say if this isn't right?

schema/revert/roles/power-bi-exporter/create.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert seattleflu/id3c-customizations:roles/power-bi-exporter/create from pg
+
+begin;
+
+drop role "power-bi-exporter";
+
```
jccraft-uw, line 6: Should there be a 'create role' in here since this file is named create.sql?

schema/deploy/shipping/views.sql
joverlee521: Ordering by date is definitely clearer, but I don't think it's reliable in our case. Each transaction of our presence/absence ETL creates multiple presence/absence records that will all have the same `created` timestamp. If there are duplicates within the same transaction, then we would have to order by `presence_absence_id` to get the newest record.

schema/revert/roles/power-bi-exporter/create.sql
joverlee521: This is a [sqitch](https://sqitch.org/) change so there are three files created: `deploy`, `revert` and `verify`. 

In this case the [deploy script](https://github.com/seattleflu/id3c-customizations/pull/117/files#diff-b168aa17ba8e63a83a7c192c28602274) creates the role, the [revert script](https://github.com/seattleflu/id3c-customizations/pull/117/files#diff-11057e04e73726cdda69d5fdfcadef1e) drops the role, and the [verify script](https://github.com/seattleflu/id3c-customizations/pull/117/files#diff-73f62be48066a6c6c8669b31ccb2760d) is empty because the database would automatically throw an error if the role couldn't be created.

schema/deploy/shipping/views.sql
jccraft-uw: I see. The created timestamp is not the time when the sample was resulted; it's when we imported it. Am I understanding what 'created' is?

schema/revert/roles/power-bi-exporter/create.sql
jccraft-uw: Gotcha. Thanks for explaining.

schema/deploy/shipping/views.sql
joverlee521: Yes, `created` is the timestamp of when the record was first processed by ID3C.

jccraft-uw: Ahhh.

schema/deploy/shipping/views.sql
tsibley: I suspect Gernot/Formative want a static SVG file they can then dynamically style, rather than SVG paths in the data. But we can export SVG paths from Pg to create the static file.

It'd be good to use relative movements and also less point precision in the path data:

    st_assvg(polygon, rel=>1, maxdecimaldigits=>5)

This will reduce the path size by a bit and make SVG renders faster. The [fifth decimal place in lon/lat gives ~1.1m precision](https://kottke.org/14/09/the-precision-of-latlong-coordinates), which should be fine for large scale thematic mapping.

schema/deploy/roles/power-bi-exporter/create.sql
```diff
@@ -0,0 +1,10 @@
+-- Deploy seattleflu/id3c-customizations:roles/power-bi-exporter/create to pg
+
+begin;
+
+create role "power-bi-exporter";
+
+comment on role "power-bi-exporter" is
+    'For exporting data to Power BI';
```
tsibley, line 8: This name seems overly generic to me since Power BI is just a tool, an implementation detail. What about something more specific, like `scan-dashboard-exporter`, that would still make sense even if, for example, the dashboard switches from Power BI to Tableau?

schema/verify/roles/power-bi-exporter/grants.sql
```diff
@@ -0,0 +1,10 @@
+-- Verify seattleflu/id3c-customizations:roles/power-bi-exporter/grants on pg
+-- requires: roles/power-bi-exporter/create
+-- requires: seattleflu/schema:shipping/schema
+
+begin;
+
+select 1/pg_catalog.has_database_privilege('return-results-exporter', :'DBNAME', 'connect')::int;
+select 1/pg_catalog.has_schema_privilege('return-results-exporter', 'shipping', 'usage')::int;
```
tsibley, line 8: These still say `return-results-exporter`.

schema/deploy/shipping/views.sql
```diff
@@ -2131,4 +2147,170 @@ grant select
...
+
+
+create or replace view shipping.scan_enrollments_v1 as
+
+    select
+        illness_questionnaire_date,
+        scan_study_arm,
+        priority_code,
+        puma,
+        centroid as seattle_neighborhood_district_centroid
```
tsibley, line 2298: I suspect the centroid as an SVG path isn't useful to Formative here. I think just leaving the neighborhood identifier might be better.

schema/revert/shipping/views.sql
```diff
@@ -1042,6 +1042,7 @@ create or replace view shipping.hcov19_observation_v1 as
         best_available_site as site,
         best_available_site_type as site_type,
 
+        location.hierarchy->'tract' as census_tract,
```
kairstenfay, line 1050: did this slip in here accidentally?

schema/revert/shipping/views.sql
```diff
@@ -1158,6 +1159,7 @@ create or replace view shipping.scan_encounters_v1 as
         age_in_years(upper(age_bin_coarse_v2.range)) as age_range_coarse_upper,
 
         location.hierarchy -> 'puma' as puma,
+        location.hierarchy -> 'tract' as census_tract,
```
kairstenfay, line 1167: did this slip in here x2 ?

schema/deploy/shipping/views.sql
```diff
@@ -2287,4 +2288,28 @@ grant select
...
+
+    select
+        illness_questionnaire_date,
+        scan_study_arm,
+        priority_code,
+        puma,
+        neighborhood_district
+
+    from shipping.scan_encounters_v1
+;
```
kairstenfay, line 2302: Because this view is only a subset of `shipping.scan_encounters_v1`, is it necessary? Can we just tell Formative which columns are relevant to them?

schema/revert/shipping/views.sql
joverlee521: This is part of the auto-generated revert script where sqitch copies the previous deploy script. I believe this is correct.

schema/revert/shipping/views.sql
joverlee521: Ditto auto-generated revert script.

schema/deploy/shipping/views.sql
joverlee521: I don't think we want Formative to have access to the entire `shipping.scan_encounters_v1` view.  I think it's easier to do that as a new view rather than maintaining grants to only specific columns of the view.

joverlee521:
>Do we now want to provide an API endpoint for exporting Seattle neighborhoods as SVG? Or how did we plan on sending them those data?

I believe we will have to create static SVGs for them rather than have them pull the SVG path from an API. @tsibley previously commented:
> I suspect Gernot/Formative want a static SVG file they can then dynamically style, rather than SVG paths in the data. But we can export SVG paths from Pg to create the static file.

schema/revert/shipping/views.sql
kairstenfay: Oh, right. I almost thought that but for some reason I overlooked that this is a `revert`. Makes sense

schema/revert/shipping/views.sql
kairstenfay: Yep, makes sense to me because it's `revert` not `deploy`.

schema/deploy/shipping/views.sql
kairstenfay: Ah, ok. Yes this probably is easier then.

PR #118, opened 2020-07-16
etl redcap-det scan: Add SCAN kiosk project

joverlee521:
Includes updates to the ingest to use the get() method for fields that
do not exist in the new project.

PR #119, opened 2020-07-21
Update SCAN return results views

joverlee521:
This PR includes changes from #118 

The name & values for `swab_type` and `staff_collected` columns were determined by the changes in https://github.com/seattleflu/lab-result-reports/pull/23


joverlee521:
[Update SCAN return of results view](https://trello.com/c/9DjYAp1l/676-update-scan-return-of-results-view)

schema/deploy/shipping/views.sql
```diff
@@ -2088,15 +2088,21 @@ create or replace view shipping.scan_return_results_v1 as
...
         barcode as qrcode,
         encountered::date as collect_ts,
-        sample.details @> '{"note": "never-tested"}' as never_tested
+        sample.details @> '{"note": "never-tested"}' as never_tested,
+        sample.details ->> 'swab_type' as swab_type,
+        -- In the future we will want to distinguish between staff-collected
+        -- vs self-collected. Currently all samples that we return results
+        -- on are only self-collected.
+        --  -Jover, 20 July 2020
+        '0' as staff_collected
```
tsibley, line 2097: Unless there's a good reason to use 0/1 for this flag in https://github.com/seattleflu/lab-result-reports/pull/23, then I think better to use either an actual boolean value or an enumerated/well-known string value.

kairstenfay: There's no good reason for my choice in seattleflu/lab-result-reports#23 to use 0/1. It was chosen arbitrarily.

kairstenfay: Should we name this `staff_observed` with values 0/1? 
`observed_by` with values "staff" or "self"?
What are your thoughts here? Trying to have this conversation in one place instead of both here and seattleflu/lab-result-reports#23.

tsibley: `staff_observed` → `{true,false}` seems good to me?

kairstenfay: Works for me.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -106,11 +115,18 @@ def redcap_det_scan(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_r
     # By verifying illness_questionnaire is complete first, we minimize the
     # delay in data ingestion since the back_end_mail_scans is completed the day after enrollment.
     #   -Jover, 29 June 2020
-    if not redcap_record['illness_q_date'] and not is_complete('back_end_mail_scans', redcap_record):
+
+    # Add check for `illness_questionnaire` is complete because the new
+    # SCAN In-Person Enrollment project does not have the `illness_q_date` field
+    # and it does not have the `back_end_mail_scans` instrument.
+    #   -Jover, 16 July 2020
+    if (not redcap_record.get('illness_q_date') and
```
kschwabefry, line 123: Would projects other than the new SCAN In-Person Enrollment project still be handled the same way with this change? E.g. for other projects, would records with incomplete enrollment which would have been previously caught by `not redcap_record['illness_q_date'] and not is_complete('back_end_mail_scans', redcap_record)` always have incomplete illness_questionnaire and still be caught by the new logic?

joverlee521: By suggestion from @tsibley , I've changed the structure of the If statement logic to now its `not(...or...or...)`. I think that clarifies the old projects would still get caught.

PR #120, opened 2020-07-23
Update is_complete checks

joverlee521:
Depends on https://github.com/seattleflu/id3c/pull/159

This should fix all of the extra "Could not create Specimen Resource due to lack of barcode." warnings we've been receiving from the mail-in project.

See commit messages for more info.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -146,13 +146,15 @@ def redcap_det_scan(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_r
 
     specimen_entry = None
     specimen_observation_entry = None
+    specimen_received = is_complete('post_collection_data_entry_qc', redcap_record)
+
     # Mail in SCAN projects have `post_collection_data_entry_qc` instrument to
     # indicate a specimen is received. The SCAN In-Person Enrollmen project
     # does not have that instrument. So we rely on `nasal_swab_collection`
     # instrument to know that we have sample data to ingest.
     #   -Jover, 16 July 2020
-    specimen_received = is_complete('post_collection_data_entry_qc', redcap_record) or \
-                        is_complete('nasal_swab_collection', redcap_record)
+    if specimen_received is None:
+        specimen_recieved = is_complete('nasal_swab_collection', redcap_record)
 
     if specimen_received:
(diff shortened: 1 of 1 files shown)


joverlee521:
[Update `is_complete()` to return `None` if instrument does not exist](https://trello.com/c/WCsPqnXD/693-update-iscomplete-to-return-none-if-instrument-does-not-exist)

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -147,13 +146,15 @@ def redcap_det_scan(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_r
...
+    specimen_received = is_complete('post_collection_data_entry_qc', redcap_record)
+
     # Mail in SCAN projects have `post_collection_data_entry_qc` instrument to
     # indicate a specimen is received. The SCAN In-Person Enrollmen project
     # does not have that instrument. So we rely on `nasal_swab_collection`
     # instrument to know that we have sample data to ingest.
     #   -Jover, 16 July 2020
-    specimen_received = is_complete('post_collection_data_entry_qc', redcap_record) or \
-                        is_complete('nasal_swab_collection', redcap_record)
+    if specimen_received is None:
```
jccraft-uw, line 156: This makes total sense for the SCAN in-person enrollment project, but I can imagine this logic not being correct for a future type of SCAN project. Would it make things easier to have a case statement for each project .... when project == "ABC" specimen_received = <blah>, when project == "DEF" specimen_received = <blah2>"
Calling out the specifics for each project could help us be more specific.

This could be a suggestion for future work ... to have project configuration that includes the logic for determining things like how to consider that a specimen was received.

kairstenfay: I think a project specific configuration is a good idea for a future improvement. I can hardly keep these straight in my head as-is. The next project we add will likely make this worse.

joverlee521: Totally agree that this logic is very specific for existing projects only! The hope is that future SCAN projects would have the **same** instruments/structure as existing projects 🤞

jccraft-uw: I added this as a comment to the Trello card about centralizing config. https://trello.com/c/LxnEd9fh

PR #123, opened 2020-07-29
etl redcap-det uw-retros: add discharge mapping for organ donor

kschwabefry:
Add a new discharge disposition mapping to the REDCap DET ETL for UW
retrospective data. We received a value of "deceased - o: deceased -
organ donor", and will map this to an appropriate FHIR code.


kschwabefry:
https://trello.com/c/uUdIZRrH/698-add-new-discharge-disposition-to-uw-retro-etl

jccraft-uw:
> Thanks for doing this! Your commit messages are very thorough. We see so many of these new codes pop up that overtime I think we as a team got a lot less descriptive about the changes in this command. Changes like these tend to happen a couple of times a week on average, so it _might_ be worth thinking about a more sustainable solution eventually.

This might fit under https://trello.com/c/LxnEd9fh

PR #124, opened 2020-07-29
views: Add kit received column to SCAN enrollments

kairstenfay:


schema/deploy/shipping/views.sql
```diff
@@ -2310,7 +2310,11 @@ create or replace view shipping.scan_enrollments_v1 as
         scan_study_arm,
         priority_code,
         puma,
-        neighborhood_district
+        neighborhood_district,
+        case
+            when sample is null then 'false'
+            else 'true'
+        end as kit_received
```
kairstenfay, line 2317: I also considered doing 0/1 here. Thoughts on the column value?

kairstenfay:
[formative request - SCAN kits returned (post_collection-data_entry_qc)](https://trello.com/c/mn36dLUX/699-formative-request-scan-kits-returned-postcollection-dataentryqc)

schema/deploy/shipping/views.sql
kschwabefry: I'd lean towards a boolean as being the clearest indicator of the potential values here. Does it still evaluate as a boolean if true/false is wrapped in single quotes?

kairstenfay: Ah, I was not even thinking of producing these in the native PostgreSQL boolean types -- I was only thinking of the downstream use cases. I think you're right -- we should use boolean true/false here. I fixed up this change into my commit.

kairstenfay: Actually, with that small change, it's clearer to me now that this conditional can be reduced to 
```
        sample is not null as kit_received
```

kairstenfay: I just re-pushed the commit with this simplification.

schema/deploy/shipping/views.sql
```diff
@@ -2310,7 +2310,8 @@ create or replace view shipping.scan_enrollments_v1 as
         scan_study_arm,
         priority_code,
         puma,
-        neighborhood_district
+        neighborhood_district,
+        sample is not null as kit_received
```
joverlee521, line 2314: I'm not sure using `sample is not null` will capture all kits received. The `sample` field is `sample.identifier` so it requires the lab to assign a sample barcode. If a sample cannot be tested, it will not have a `sample.identifier` and will not show as kit received. 

I think `sample is not null OR never_tested` should include all kits?

kairstenfay: Ah, good point! I will update using your suggestion.

kairstenfay: Fixed up in the commit. I modified your example slightly to set null results for `never_tested` to be false here

PR #125, opened 2020-08-05
SCAN Negative Counts

joverlee521:
Formative requested that we include negative counts by day as well so they can calculate the percent positives.

### DIFF
--- a/lib/seattleflu/id3c/api/routes.py
+++ b/lib/seattleflu/id3c/api/routes.py
@@ -109,11 +109,11 @@ def get_scan_demographics(session):
 @authenticated_datastore_session_required
 def get_scan_positives(session):
     """
-    Export aggregate numbers of hCoV-19 positives for SCAN
+    Export aggregate numbers of hCoV-19 result counts for SCAN
     """
-    LOG.debug("Exporting hCoV-19 positives for SCAN")
+    LOG.debug("Exporting hCoV-19 result counts for SCAN")
 
-    positives = datastore.fetch_rows_from_table(session, ("shipping", "scan_hcov19_positives_v1"))
+    positives = datastore.fetch_rows_from_table(session, ("shipping", "scan_hcov19_result_counts_v1"))
 
     return Response((row[0] + '\n' for row in positives), mimetype="application/x-ndjson")
 
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -17,7 +17,8 @@ begin;
 -- worry about view dependencies when reworking view definitions.
 drop view if exists shipping.scan_enrollments_v1;
 drop view if exists shipping.seattle_neighborhood_districts_v1;
-drop view if exists shipping.scan_hcov19_positives_v1;
+drop view if exists shipping.scan_hcov19_positives_v1; -- Can be deleted in next rework -Jover, 06 Aug 2020
+drop view if exists shipping.scan_hcov19_result_counts_v1;
 drop view if exists shipping.scan_demographics_v1;
 
 drop view if exists shipping.scan_return_results_v1;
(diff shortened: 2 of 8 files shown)


joverlee521:
[Add negative counts to `shipping.scan_hcov19_positives_v1`](https://trello.com/c/KM7SlilR/706-add-negative-counts-to-shippingscanhcov19positivesv1)

joverlee521:
> The view is called "hcov19_positives" but we now show negatives, too. Perhaps we ought to rename it so it's more accurate?

I wondered about the name too. Would `scan_hcov19_result_counts` be more appropriate?

kairstenfay:
> > The view is called "hcov19_positives" but we now show negatives, too. Perhaps we ought to rename it so it's more accurate?
> 
> I wondered about the name too. Would `scan_hcov19_result_counts` be more appropriate?

Yes, I think that's an ideal name.

joverlee521:
I've renamed the view and also took the opportunity to remove the unused `seattle_neighborhood_districts` view.

schema/deploy/shipping/views.sql
```diff
@@ -2250,56 +2251,38 @@ create or replace view shipping.scan_hcov19_positives_v1 as
...
+        count(*) filter (where hcov19_result = 'negative' and county = 'yakima') as yakima_county_negatives,
+        count(*) filter (where hcov19_result in ('positive', 'inconclusive') and county is null) as other_positives,
+        count(*) filter (where hcov19_result = 'negative' and county is null) as other_negatives
     from scan_hcov19_results
     group by hcov19_result_release_date
 ;
 
-comment on view shipping.scan_hcov19_positives_v1 is
+comment on view shipping.scan_hcov19_result_counts_v1 is
   'A view of counts of hcov19 positives from the SCAN project grouped by date results were released.';
```
jccraft-uw, line 2268: Nipik: Update the comment from positives to results.

joverlee521: Oh, will update! Thank you 👍

schema/sqitch.plan
```diff
@@ -206,3 +206,5 @@ shipping/views [shipping/views@2020-07-17] 2020-07-20T20:56:28Z Jover Lee <jover
 
 shipping/views [shipping/views@2020-07-20] 2020-07-29T18:37:58Z Kairsten Fay <kfay@fredhutch.org> # Add kit received to SCAN enrollments
 @2020-07-29 2020-07-29T18:42:32Z Kairsten Fay <kfay@fredhutch.org> # Schema as of 29 July 2020
+
+shipping/views [shipping/views@2020-07-29] 2020-08-05T20:40:42Z Jover Lee <joverlee@fredhutch.org> # Add negative counts to shipping.scan_hcov19_positives_v1
```
kairstenfay, line 210: I believe this comment is out of date

schema/deploy/shipping/views.sql
```diff
@@ -2250,33 +2251,38 @@ create or replace view shipping.scan_hcov19_positives_v1 as
...
 
 -- Even if it's just aggregate counts of hcov19 positives,
```
kairstenfay, line 2270: "positives" should be "results" here

But honestly a pretty minor change so it's probably fine to leave

lib/seattleflu/id3c/api/routes.py
```diff
@@ -109,11 +109,11 @@ def get_scan_demographics(session):
 @authenticated_datastore_session_required
 def get_scan_positives(session):
```
kairstenfay, line 110: Should this function be renamed?

joverlee521: Gernot had asked to add it to the existing endpoint, so I assume they don't want to change the API call on their end.

PR #126, opened 2020-08-07
Use sample.collected in SCAN return of results

tsibley:

- [x] Update 2020-08-14 `sample.created` threshold based on when this will be deployed. It can be sooner if this will be deployed sooner; it should be later if this is deployed later.
- [ ] Deploy in tandem with https://github.com/seattleflu/id3c/pull/161 
- [ ] Setup Metabase QC pulse/dashboard to identify cases where `sample.collected != sample.details->>'date'` for ongoing review of disagreements between data sources.
- [ ] Update SecureLink portal to say "Collection date" instead of "Enrollment date". (The PDFs we generate already say this.)


schema/deploy/shipping/views.sql
```diff
@@ -2089,7 +2089,10 @@ create or replace view shipping.scan_return_results_v1 as
       select
         sample_id,
         barcode as qrcode,
-        encountered::date as collect_ts,
+        case when sample.created >= '2020-08-14'
```
joverlee521, line 2092: Can we use another date field as the cutoff? I think it should be a date that we can easily point to for people outside the dev team. `sample.created` doesn't necessarily correlate with any specific date outside the database.

joverlee521:
[Ingest kit_reg_date for collection_date on lab ror pdf](https://trello.com/c/Xp2jXdtm/701-ingest-kitregdate-for-collectiondate-on-lab-ror-pdf)

schema/deploy/shipping/views.sql
tsibley: Hmm. We do expose it via the [Sample Lookup](https://backoffice.seattleflu.org/metabase/dashboard/45) dashboard in Metabase, but still, a good point. I used `sample.created` because it's more stable and reliable than other dates so the condition wouldn't shift over time. Do you have any suggestions on a better date field to use?

joverlee521: Posting here for posterity's sake:
I was thinking encountered date would be best? That way we can say any participant who enrolled after X date will have the corrected collection date.

tsibley: (And that's what we decided to do. :-)

tsibley:
:facepalm: Fixed.

PR #127, opened 2020-08-12
shipping.scan_enrollments_v1: add `geo_location_name` to match GeoJSON

joverlee521:
Include a human readable location name for neighborhood district
or PUMA so that it can be linked to the location name in the GeoJSON.
This link must be done using the human readable name because the linking
column is what shows up on hover of the PowerBI map.

Edit: Depends on #128


schema/deploy/shipping/views.sql
```diff
@@ -2288,15 +2286,31 @@ grant select
...
         illness_questionnaire_date,
         scan_study_arm,
         priority_code,
         puma,
         neighborhood_district,
+        geo_location_name,
         sample is not null or never_tested is not null as kit_received
 
     from shipping.scan_encounters_v1
+    join location_names on coalesce(neighborhood_district,puma) = lower(location_names.identifier)
```
tsibley, line 2312: I'm a little sad this is the join key, but ah well. I see why it's that way, and it's probably not worth larger reworking to make it not be. The other reason to fix it would be to provide PUMA names for outside of Washington, but I don't think that's in-scope for their dashboard right now?

schema/deploy/shipping/views.sql
```diff
@@ -2288,15 +2286,31 @@ grant select
...
+        where scale in ('neighborhood_district', 'puma')
+        and hierarchy @> 'state => washington'
+    )
+
     select
         illness_questionnaire_date,
         scan_study_arm,
         priority_code,
         puma,
         neighborhood_district,
```
tsibley, line 2307: This view is just for Formative's dashboards, right? I think with the new `geo_location_name` field, these `puma` and `neighborhood_district` fields aren't needed…?

joverlee521: I kept the fields so that nothing on their end breaks when I merge/deploy this change.

schema/revert/shipping/views.sql
```diff
@@ -2251,56 +2249,38 @@ create or replace view shipping.scan_hcov19_positives_v1 as
...
-  'A view of counts of hcov19 positives from the SCAN project grouped by date results were released.';
+comment on view shipping.scan_hcov19_result_counts_v1 is
+  'A view of counts of hcov19 results from the SCAN project grouped by date results were released.';
 
--- Even if it's just aggregate counts of hcov19 positives,
+-- Even if it's just aggregate counts of hcov19 results,
 -- we should probably restrict access to this view to only hcov19-visibility
 -- and scan-dashboard-exporter.
 --  -Jover, 9 July 2020
-revoke all on shipping.scan_return_results_v1 from reporter;
```
jccraft-uw, line 2269: How is the scan_result_results_v1 view related to this PR? How is reporter related to this PR?

schema/revert/shipping/views.sql
```diff
@@ -2251,56 +2249,38 @@ create or replace view shipping.scan_hcov19_positives_v1 as
...
+revoke all on shipping.scan_hcov19_result_counts_v1 from reporter;
 
 grant select
-    on shipping.scan_return_results_v1
+    on shipping.scan_hcov19_result_counts_v1
```
jccraft-uw, line 2275: Does hcov19-visibility no longer need to access scan_return_results_v1?

schema/revert/shipping/views.sql
joverlee521: It's not related to this PR at all. This revert script is automatically generated by sqitch. It basically copies the deploy script over, so when we run `sqitch revert` it will revert back to the previous state. 
The only actual changes I made to this revert script is dropping lines 19-20.

schema/revert/shipping/views.sql
jccraft-uw: I bet this is the same as above ... sqitch does its thing for the revert scripts.

joverlee521: Yup, same auto-generated revert script

schema/deploy/shipping/views.sql
joverlee521: Which specific part makes you sad? The `lower()` is only needed for the neighborhood districts. 
I don't think this affects the PUMAs at all since those are just strings of integers.

tsibley: Sorry, should have been more specific. Nothing in our schema guarantees that `warehouse.location.identfier` is by itself unique. It's unique in combination with `scale`, as `(scale, identifier)` forms the PK. So joining on the partial key is prone to collision. That said, I realize in practice this is pretty unlikely to occur with the identifiers we use for those scales.

I didn't even notice the `lower()` before, but that strikes me as an oversight on our part earlier while populating the location catalog. The convention is that the values in `hierarchy` match those in `identifier` (if they exist at all).

schema/revert/shipping/views.sql
tsibley: GitHub doesn't show this sort of change well in its diffs, but on the command-line, git can use rename/copy detection to show a much nicer diff when a file is renamed/copied with some changes.

PR #129, opened 2020-08-19
Formative v2 API + internal dashboard API

joverlee521:
- Change `shipping.scan_encounters_v1` to a materialized view in order to speed up downstream queries
- Add `shipping.scan_demographics_v2`
- Add `shipping.scan_redcap_enrollments_v1`

Will add a new cronjob to backoffice to refresh the `shipping.scan_encounters_v1` view every 5 minutes. See https://github.com/seattleflu/backoffice/pull/73

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -15,6 +15,7 @@ begin;
 
 -- Drop all views at the top in order of dependency so we don't have to
 -- worry about view dependencies when reworking view definitions.
+drop view if exists shipping.scan_redcap_enrollments_v1;
 drop view if exists shipping.scan_enrollments_v1;
 drop view if exists shipping.scan_hcov19_result_counts_v1;
 drop view if exists shipping.scan_demographics_v2;
@@ -2422,6 +2423,30 @@ revoke all
 
 grant select
     on shipping.scan_enrollments_v1
+    to "scan-dashboard-exporter";
+
+
+create or replace view shipping.scan_redcap_enrollments_v1 as
+
+    select
+        illness_questionnaire_date,
+        scan_study_arm,
+        puma,
+        age_range_fine,
+        age_range_fine_lower,
+        age_range_fine_upper
+    from shipping.scan_encounters_v1
+;
+
+comment on view shipping.scan_redcap_enrollments_v1 is
+  'A view of SCAN REDCap enrollments for internal Power BI dashboards';
+
+revoke all
+    on shipping.scan_redcap_enrollments_v1
+    from "scan-dashboard-exporter";
+
+grant select
+    on shipping.scan_redcap_enrollments_v1
     to "scan-dashboard-exporter";
 
 commit;
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -17,6 +17,7 @@ begin;
 -- worry about view dependencies when reworking view definitions.
 drop view if exists shipping.scan_enrollments_v1;
 drop view if exists shipping.scan_hcov19_result_counts_v1;
+drop view if exists shipping.scan_demographics_v2;
 drop view if exists shipping.scan_demographics_v1;
 
 drop view if exists shipping.scan_return_results_v1;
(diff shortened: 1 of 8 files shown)


joverlee521:
[Formative dashboard changes](https://trello.com/c/J7E4wxud/709-formative-dashboard-changes)

schema/deploy/shipping/views.sql
```diff
@@ -2195,6 +2196,96 @@ grant select
...
+
+    with hcov19_presence_absence as (
+        -- Collapse potentially multiple hCoV-19 results
+        select distinct on (sample_id)
+            sample.identifier as sample,
+            presence_absence_id,
+            case
+              when present is true then 'positive'
+              when present is false then 'negative'
+              when present is null then 'positive'
```
kairstenfay, line 2210: is this a typo?

kairstenfay: ah wait, I forgot that Formative wants to treat inconclusive results as positive.

schema/deploy/shipping/views.sql
```diff
@@ -2195,6 +2196,96 @@ grant select
...
+            income,
+            sample
+        from shipping.scan_encounters_v1
+        -- Limit to King County participants only since the dashboard
+        -- is only for King County.
+        where puma in (
+          '5311601', '5311602', '5311603', '5311604',
+          '5311605', '5311606', '5311607', '5311608',
+          '5311609', '5311610', '5311611', '5311612',
+          '5311613', '5311614', '5311615', '5311616')
```
kairstenfay, line 2255: where'd you get this list of King County PUMAs by the way?

joverlee521:
Rebased on to master to fix sqitch conflicts.

schema/deploy/shipping/views.sql
joverlee521: Found here: https://www.census.gov/geographies/reference-maps/2010/geo/2010-pumas/washington.html

I was actually wondering if we should formalize the PUMAs -> county hierarchy in the locations table.

kairstenfay: That's not a bad idea, at the very least for WA counties or King & Yakima counties.

schema/deploy/shipping/views.sql
```diff
@@ -2198,6 +2199,96 @@ grant select
...
+            and target.identifier not in ('COVID-19_Orf1b', 'COVID-19-S_gene')
+        /*
+          Keep only the most recent push. According to Lea, samples are only
+          retested if there is a failed result. A positive, negative, or
+          indeterminate result would not be retested.
+
+          https://seattle-flu-study.slack.com/archives/CV1E2BC8N/p1584570226450500?thread_ts=1584569401.449800&cid=CV1E2BC8N
+        */
+        order by
+            sample_id, presence_absence_id desc
```
kschwabefry, line 2237: In order to keep the most recent results push, I wonder if ordering by hcov19_result_release_date be more explicit than the id field

schema/deploy/shipping/views.sql
```diff
@@ -1172,6 +1173,10 @@ create or replace view shipping.scan_encounters_v1 as
         age_in_years(lower(age_bin_coarse_v2.range)) as age_range_coarse_lower,
         age_in_years(upper(age_bin_coarse_v2.range)) as age_range_coarse_upper,
 
+        age_bin_decade_v1.range as age_range_decade,
+        age_in_years(lower(age_bin_decade_v1.range)) as age_range_decade_lower,
+        age_in_years(upper(age_bin_decade_v1.range)) as age_range_decade_upper,
```
kschwabefry, line 1180: I'm still learning about postgres ranges, and so I'm wondering why we return all three of these values. If we include the range itself, does that contain the same information as pulling out the lower and upper explicitly?

schema/deploy/shipping/views.sql
joverlee521: There's an edge case of the same sample being tested twice in the same day, so ordering by release date would not work. See [previous discussions](https://github.com/seattleflu/id3c-customizations/pull/117#discussion_r454043119)

schema/deploy/shipping/views.sql
joverlee521: We've gotten feedback in the past from downstream users (I think IDM?) that the "range" itself can be tricky to use in certain cases, so it would be best to also include the int version of the ranges.

PR #130, opened 2020-08-19
etl: Create distinct code for having no symptoms

kairstenfay:
Currently in ID3C, a `null` symptoms response captures both encounters
with asymptomatic participants as well as participants who did not fill
out the Illness Questionnaire. Instead of relying on a proxy field
like `illness_questionnaire_complete` or `illness_q_date` to determine
if a `null` symptoms response is indicative of an asymptomatic
participant or simply indicates missing data, create a new code to
distinguish the condition where a participant reports having no
symptoms.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
@@ -23,7 +23,7 @@ from . import race
 LOG = logging.getLogger(__name__)
 
 
-REVISION = 1
+REVISION = 2
 
 REDCAP_URL = 'https://redcap.iths.org/'
 INTERNAL_SYSTEM = "https://seattleflu.org"
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py
@@ -37,7 +37,7 @@ REQUIRED_INSTRUMENTS = [
 # REDCap DET records lacking this revision number in their log.  If a
 # change to the ETL routine necessitates re-processing all REDCap DET records,
 # this revision number should be incremented.
-REVISION = 4
+REVISION = 5
 
 
 @redcap_det.command_for_project(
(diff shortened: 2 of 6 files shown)


kairstenfay:
[Distinguish between having no symptoms and not filling out symptoms field in REDCap](https://trello.com/c/6GSdFKuH/714-distinguish-between-having-no-symptoms-and-not-filling-out-symptoms-field-in-redcap)

kairstenfay:
> While looking to see if a "none" code violates the FHIR docs, I found this bit on the [Conditions page](https://www.hl7.org/fhir/condition.html):
> 
> > This resource is not typically used to record information about subjective and objective information that might lead to the recording of a Condition resource. Such signs and symptoms are typically captured using the Observation resource; although in some cases a persistent symptom, e.g. fever, headache may be captured as a condition before a definitive diagnosis can be discerned by a clinician. By contrast, headache may be captured as an Observation when it contributes to the establishment of a meningitis Condition.
> 
> > Use the Observation resource when a symptom is resolved without long term management, tracking, or when a symptom contributes to the establishment of a condition.
> 
> > Use Condition when a symptom requires long term management, tracking, or is used as a proxy for a diagnosis or problem that is not yet determined.
> 
> Out of the scope of this PR, but just wanted to bring it up somewhere since it seems like we are using the wrong FHIR resource for symptoms woman_facepalming

Thanks for your due diligence!

Oof, so according to those docs, we may want to switch to using Observation resources to track symptoms? While that is definitely extra work in the short term it may simplify our FHIR Bundles a bit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
```diff
@@ -23,7 +23,7 @@
 LOG = logging.getLogger(__name__)
 
 
-REVISION = 1
+REVISION = 2
```
kschwabefry, line 26: How do we decide when to bump a revision number?

joverlee521: We bump revision number when we need to re-process all records in the receiving table

kairstenfay:
> Oof, so according to those docs, we may want to switch to using Observation resources to track symptoms? While that is definitely extra work in the short term it may simplify our FHIR Bundles a bit.

I made [a Trello card](https://trello.com/c/ScSEnPUV/761-revisit-fhir-usage-of-condition-for-symptoms) for this task.

kairstenfay:
Rebased onto master.

PR #131, opened 2020-08-21
etl redcap-det scan: move age ceiling logic to fix `None` age values

joverlee521:
Move the logic for age ceiling to within the `cast_to_integer` function
of `questionnaire_item`. This prevents age from being returned as `None`
due to 0 being a falsy value.

This also consolidates the type casting of responses into a single
function for easiers future debugging.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -54,7 +54,7 @@ PROJECTS = [
 
 ]
 
-REVISION = 18
+REVISION = 19
 
 REDCAP_URL = 'https://redcap.iths.org/'
 INTERNAL_SYSTEM = "https://seattleflu.org"
@@ -948,13 +948,6 @@ def create_initial_questionnaire_response(record: dict, patient_reference: dict,
     record['country'] = combine_multiple_fields('country')
     record['state'] = combine_multiple_fields('state')
 
-    # Age Ceiling
-    try:
-        record['age'] = age_ceiling(int(record['age']))
-        record['age_months'] = age_ceiling(int(record['age_months']) / 12) * 12
-    except ValueError:
-        record['age'] = record['age_months'] = None
-
     return questionnaire_response(record, question_categories, patient_reference, encounter_reference)
 
 
@@ -1025,6 +1018,17 @@ def questionnaire_item(record: dict, question_id: str, response_type: str) -> Op
     response = record.get(question_id)
     if not response:
         return None
+
+    # Age ceiling
+    if question_id in {'age', 'age_months'}:
+        try:
+            value = int(response)
+            if question_id == 'age':
+                response = age_ceiling(value)
+            if question_id == 'age_months':
+                response = age_ceiling(value / 12) * 12
+        except ValueError:
+            return None
 
     def cast_to_coding(string: str):
         """ Currently the only QuestionnaireItem we code is race. """


lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -51,7 +51,7 @@ class ScanProject(NamedTuple):
 
 ]
 
-REVISION = 15
+REVISION = 16
```
joverlee521, line 54: Will need to bump revision again if #130 is merged before this PR.

joverlee521:
[Age < 1 month gets ingested as None](https://trello.com/c/RAsXJGpG/711-age-1-month-gets-ingested-as-none)

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -889,7 +882,13 @@ def cast_to_string(string: str) -> Optional[str]:
...
     def cast_to_integer(string: str) -> Optional[int]:
         try:
-            return int(string)
+            # Age Ceiling
+            if question_id == 'age':
+                return age_ceiling(int(string))
+            elif question_id == 'age_months':
+                return age_ceiling(int(string) / 12) * 12
+            else:
+                return int(string)
```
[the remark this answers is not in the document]
joverlee521: Do you mean the name of the function `cast_to_integer` is misleading because `age_ceiling` returns a float?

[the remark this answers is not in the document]
joverlee521: The type hint in `age_ceiling` indicates it returns a float, but that only happens if the passed `age` argument is a float. Since the age values we pass are integers, these will return int as expected.

The issue of reporting age >= 1y as a float is actually  within the FHIR ETL not here.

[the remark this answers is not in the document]
kairstenfay: > Do you mean the name of the function `cast_to_integer` is misleading because `age_ceiling` returns a float?

Yes.

> The type hint in `age_ceiling` indicates it returns a float, but that only happens if the passed age argument is a float. Since the age values we pass are integers, these will return int as expected.

> The issue of reporting age >= 1y as a float is actually within the FHIR ETL not here.

Ah, then perhaps all we need to do is update the type hint for the return value of `age_ceiling`.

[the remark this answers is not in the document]
joverlee521: I restructured the function so that everything gets returned as an int in the latest push.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -888,8 +881,17 @@ def cast_to_string(string: str) -> Optional[str]:
         return None
 
     def cast_to_integer(string: str) -> Optional[int]:
+        value: Union[str, float] = string
```
kairstenfay, line 884: Would it be better to set the parameter `string`'s type hint to `Union[str, float]` rather than setting it here? (maybe also renaming `string`)

joverlee521: We expect the argument to be a str, so i think we should keep it as such. 

I had to create a separate variable with the different type hint to avoid a typing error when I reassign the value with `age_ceiling()`:
```
error: Incompatible types in assignment (expression has type "float", variable has type "str")
```

kairstenfay: Ah, sorry. I didn't dig into the details of this enough. I believe the ideal solution is to change the expected argument in `age_ceiling()` to be `Union[float, int]`.

kairstenfay: ...and perhaps also have `age_ceiling()` return a type of `Union[float, int]`? Although not including this change doesn't seem to cause an error.

joverlee521: Ah, sorry I should have mentioned this: I get an error when I _only_ change the type in `age_ceiling()` to `Union[float, int]`
```
error: Incompatible return value type (got "Union[int, float]", expected "Optional[int]")
```

kairstenfay: That's odd.. the mypy tests pass for me when I only change the param type in `age_ceiling()` to `Union[float, int]`.

jccraft-uw: I know that @joverlee521 pipenv update 'd a few weeks ago to get the same mypy behavior that we're seeing Jenkins. Maybe we all need to update.

kairstenfay: I ran `pipenv sync` recently so I don't think that's the cause.

joverlee521: @kairstenfay Did you run `pipenv sync --dev`? Default `pipenv sync` does not include development dependencies like mypy.

kairstenfay: Good point. yes I did.

kairstenfay: I created a test branch using my recommendation and the Travis CI build didn't fail.
[See it here](https://github.com/seattleflu/id3c-customizations/runs/1027654962). 

(Ignore the unused import ;))

joverlee521: Huh, my errors were due to the return value of `age_ceiling` not having the right type, not the parameter. I'll just update both so it reflects the truth. 

It was initially just set to `float` because in [PEP 484 type hints](https://www.python.org/dev/peps/pep-0484/#the-numeric-tower) says:
> Rather than requiring that users write import numbers and then use numbers.Float etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type float, an argument of type int is acceptable; similar, for an argument annotated as having type complex, arguments of type float or int are acceptable.

kairstenfay: Hmm, I wonder if this is another mypy bug.

kairstenfay: I don't think this line is necessary anymore given your second commit, btw.

joverlee521: It is required because the variable `string` is expected to be a string and cannot be reassigned to a float/int.

kairstenfay: But do we ever actually expect to get a non-string input variable here? We cast it to `int` before passing it to `age_ceiling`. Perhaps I'm missing something.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -1040,7 +1033,16 @@ def cast_to_string(string: str) -> Optional[str]:
 
     def cast_to_integer(string: str) -> Optional[int]:
         try:
-            return int(string)
+            value = int(string)
+            # Age Ceiling
+            if question_id == 'age':
```
jccraft-uw, line 1038: It is a bit unusual to special case age and age_months inside the generic cast_to_integer function. Instead, we could have created `age_months_ceiling` or something like that.
But given that the cast_to_integer function is in this specific ETL code and is not in a library shared across ETLs, that's OK.

jccraft-uw (approved):
Just trying this out:
5 months old
value = int(age_ceiling(5 / 12) * 12)   = 5
Cool

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
joverlee521: Hmm. Now that you've pointed it out, it does seem like a weird side-effect for a function called `cast_to_integer`. I can move the age_ceiling outside of the cast function, but still stay within `questionnaire_item`.

joverlee521: Made this change in the latest commit.

jccraft-uw: Right on!

joverlee521:
Deployed

PR #132, opened 2020-08-22
Skip record with unmapped discharge value

jccraft-uw:
Before this change, an unmapped discharge disposition value would stop
the ETL process. With this change, we skip the problematic record and
continue processing.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
@@ -182,7 +182,11 @@ def create_encounter(db: DatabaseSession,
     if not encounter_location_references:
         return None, None
 
-    hospitalization = create_encounter_hospitalization(record)
+    try:
+        hospitalization = create_encounter_hospitalization(record)
+    except UnknownHospitalDischargeDisposition as e:
+        LOG.warning(e)
+        return None, None
 
     encounter_date = record["collection_date"]
     if not encounter_date:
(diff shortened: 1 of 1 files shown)


lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -182,7 +182,11 @@ def create_encounter(db: DatabaseSession,
     if not encounter_location_references:
         return None, None
 
-    hospitalization = create_encounter_hospitalization(record)
+    try:
+        hospitalization = create_encounter_hospitalization(record)
+    except UnknownHospitalDischargeDisposition as exc:
+        LOG.warning(exc)
```
kairstenfay, line 188: `exc` is a bit unconventional here; typically in python we see it called `exception` or simply, `e`.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -340,7 +344,7 @@ def discharge_disposition(redcap_record: dict) -> Optional[str]:
     standardized_disposition = standardize_whitespace(disposition.lower())
 
     if standardized_disposition not in mapper:
-        raise Exception(f"Unknown discharge disposition value «{standardized_disposition}».")
+        raise UnknownHospitalDischargeDisposition(f"Unknown discharge disposition value «{standardized_disposition}» for barcode «{redcap_record['barcode']}».")
```
kairstenfay, line 347: Can we wrap this line to keep it from exceeding 100 characters?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
jccraft-uw: I like "e" !

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
jccraft-uw: Fixed!

jccraft-uw:
> `present()` function (see the error message "Unknown test result value...")

I created a Trella card to do the same skipping for present()
https://trello.com/c/mf2fFe8C

PR #133, opened 2020-08-27
Add UW retro discharge dispo mapping: hosp-med fac

jccraft-uw:


jccraft-uw:
[UW retrospective: add mapping and generate DET](https://trello.com/c/j2BrSD9Y/753-uw-retrospective-add-mapping-and-generate-det)

PR #134, opened 2020-08-31
Support repeating REDCap

joverlee521:
See details in commit messages.

Needs to be deployed with changes in ID3C core (https://github.com/seattleflu/id3c/pull/163)

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
@@ -43,7 +43,6 @@ REQUIRED_INSTRUMENTS = [
     "asymptomatic-swab-n-send",
     redcap_url = REDCAP_URL,
     project_id = PROJECT_ID,
-    required_instruments = REQUIRED_INSTRUMENTS,
     revision = REVISION,
     help = __doc__)
 
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
@@ -17,7 +17,7 @@ from id3c.cli.command.location import location_lookup
 from seattleflu.id3c.cli.command import age_ceiling
 from .redcap_map import *
 from .fhir import *
-from . import race, first_record_instance
+from . import race, first_record_instance, required_instruments
 
 
 LOG = logging.getLogger(__name__)
@@ -47,6 +47,7 @@ REQUIRED_INSTRUMENTS = [
     help = __doc__)
 
 @first_record_instance
+@required_instruments(REQUIRED_INSTRUMENTS)
 def redcap_det_asymptomatic_swab_n_send(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_record: dict) -> Optional[dict]:
     location_resource_entries = locations(db, cache, redcap_record)
     patient_entry, patient_reference = create_patient(redcap_record)
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
@@ -17,7 +17,7 @@ from id3c.cli.command.location import location_lookup
 from seattleflu.id3c.cli.command import age_ceiling
 from .redcap_map import *
 from .fhir import *
-from . import race
+from . import race, first_record_instance
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


joverlee521:
[See what RedCAP records look like when a repeating instrument is used multiple times. How would we ingest that?](https://trello.com/c/z7eBYmud/743-see-what-redcap-records-look-like-when-a-repeating-instrument-is-used-multiple-times-how-would-we-ingest-that)

joverlee521:
> This looks really solid, Jover.
> 
> However, when I run `id3c redcap-det generate` in this repo, I get the following error:
> 
> ```
> TypeError: __init__() got an unexpected keyword argument 'required_instruments'
> ```

That's weird. I don't see this error. Did you run `pipenv sync` after checking out the two branches in both id3c and id3c-customizations?

lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
```diff
@@ -43,11 +44,15 @@
     "asymptomatic-swab-n-send",
     redcap_url = REDCAP_URL,
     project_id = PROJECT_ID,
-    required_instruments = REQUIRED_INSTRUMENTS,
     revision = REVISION,
     help = __doc__)
 
-def redcap_det_asymptomatic_swab_n_send(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_record: dict) -> Optional[dict]:
+@get_first_redcap_record_instance_only
```
kschwabefry, line 50: Does the `@get_first_redcap_record_instance_only` decorator resolve before the `@redcap_det.command_for_project`? If so, does the behavior around required instruments differ across ETLs still? If the behavior will consistently just to iterate over all the record instances that exist (since its already reduced to first instance only for non-longitudinal ETLs) to check for required instruments, I'm wondering if this could still be in core.

kairstenfay:
> That's weird. I don't see this error. Did you run `pipenv sync` after checking out the two branches in both id3c and id3c-customizations?

Yes

kairstenfay:
> > That's weird. I don't see this error. Did you run `pipenv sync` after checking out the two branches in both id3c and id3c-customizations?
> 
> Yes

Oh, I just realized this error is also occurring for me on the master branch, so it's not likely to be related to your changes. Sorry about that.

lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
joverlee521: `@get_first_redcap_record_instance_only` is injected between `@redcap_det.command_for_project` and the custom ETL function. 

The check for required instruments got moved to customizations because it wouldn't work the same way for repeating REDCap projects. In repeating projects, each record instance has all of the fields, but only has the values for that specific instance. So you can't just check the first instance for all required instruments.

lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py
```diff
@@ -44,11 +45,15 @@
     "kiosk",
     redcap_url = REDCAP_URL,
     project_id = PROJECT_ID,
-    required_instruments = REQUIRED_INSTRUMENTS,
     revision = REVISION,
     help = __doc__)
 
-def redcap_det_kisok(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_record: dict) -> Optional[dict]:
+@get_first_redcap_record_instance_only
```
tsibley, line 51: There's a convention that decorator names are declarative instead of imperative, since they're a bit of declarative, not imperative, syntax. `get_first_redcap_record_instance_only` is a bit of a mouthful too. What about something like `first_record_instance` or `first_instance_of_record`?

lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py
```diff
@@ -44,11 +45,15 @@
...
+def redcap_det_kisok(*, db: DatabaseSession, cache: TTLCache, det: dict, redcap_record: REDCapRecord) -> Optional[dict]:
+    # Only process REDCap record if all required instruments are complete
+    if not required_instruments_complete(REQUIRED_INSTRUMENTS, redcap_record):
+        return None
```
tsibley, line 55: It seems natural to me to make this required instruments check a bit of declarative syntax via a decorator too. e.g. it moves from `@command_for_project(required_instruments = …)` to `@required_instruments(…)`.

The existing routines would then look like:

```python
@redcap_det.command_for_project(…)
@first_instance_of_record
@required_instruments(…)
def redcap_det_kiosk(…):
    …
```

tsibley (changes requested):
Looks good. Some minor comments on naming.

lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py
joverlee521: Ah, good know. Will update, thanks!

lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py
joverlee521: Hmm yeah that makes sense, I'll update.

joverlee521: Updated in latest commit

tsibley (approved):

PR #140, opened 2020-09-11
etl redcap-det scan: normalize SCAN husky netids

kschwabefry:
Normalize netids and add @washington.edu domain before hashing for patient_id.
The netids coming from redcap sometimes have @uw.edu, and sometimes no suffix.
Normalizing helps standardize and also reduces potential future conflicts if we
hash other usernames.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap.py
@@ -1,0 +1,46 @@
+"""
+Functions shared by REDCap DET ETL
+"""
+from typing import Optional
+
+
+def normalize_net_id(net_id: str=None) -> Optional[str]:
+    """
+    If netid or UW email provided, return netid@washington.edu
+
+    >>> normalize_net_id('abcd')
+    'abcd@washington.edu'
+
+    >>> normalize_net_id('aBcD ')
+    'abcd@washington.edu'
+
+    >>> normalize_net_id('abcd@uw.edu')
+    'abcd@washington.edu'
+
+    >>> normalize_net_id('aBcD@u.washington.edu')
+    'abcd@washington.edu'
+
+    If missing netid or non-uw email provided, it cannot be normalized so return nothing
+    >>> normalize_net_id('notuw@gmail.com')
+    >>> normalize_net_id('nodomain@')
+    >>> normalize_net_id('multiple@at@signs')
+    >>> normalize_net_id()
+    >>> normalize_net_id('')
+    >>> normalize_net_id(' ')
+    """
+
+    if not net_id or net_id.isspace():
+        return None
+
+    net_id = net_id.strip().lower()
+
+    # if a uw email was entered, drop the domain before normalizing
+    if net_id.count('@') == 1:
+        net_id, domain = net_id.split('@')
+        if domain not in ['u.washington.edu', 'uw.edu']:
+            return None
+    elif net_id.count('@') > 1:
+        return None
+
+    username = f'{net_id}@washington.edu'
+    return username
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -17,6 +17,7 @@ from id3c.cli.command.de_identify import generate_hash
 from id3c.cli.redcap import is_complete, Record as REDCapRecord
 from seattleflu.id3c.cli.command import age_ceiling
 from .redcap_map import *
+from .redcap import *
 from .fhir import *
 from . import race, first_record_instance, required_instruments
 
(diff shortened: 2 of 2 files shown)


kschwabefry:
[Clean up `netid` field in SCAN Husky ETL](https://trello.com/c/EAXIHYGH/784-clean-up-netid-field-in-scan-husky-etl)

kschwabefry:
Should this include bumping the revision number to re-process records?

jccraft-uw (approved):

tsibley:
_recording our conversation from standup for posterity_

> Should this include bumping the revision number to re-process records?

@joverlee521 said (paraphrasing) that normally it should, but since the revision for this ETL currently applies to all SCAN REDCap projects, that would mean re-processing many more records than strictly necessary. She suggested instead generating new (synthetic) DETs (using `id3c redcap-det generate …`) just for the affected project.

tsibley:
(I'll note one potential downside to this revision-bump workaround is that the code change _does_ accidentally impact ingest of the other projects, but we don't realize that now and only find out later when the revision is bumped or new records from those projects come in. I think this is unlikely in this case, however, and worth the risk.)

tsibley (approved):
:+1: 

Since you asked about doctests yesterday, I'll add that this code seems like a perfect candidate for pulling into a pure function with doctests. :-) Would be happy to see that, but I don't consider it a blocker for merge.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -470,14 +470,16 @@ def create_patient(record: REDCapRecord) -> tuple:
...
 
-    # Use the UW NetID to generate patient id if it's available.
+    # Use the UW NetID and domain to generate patient id, if it's available.
     # This is the stable identifier that can allow us to match UW reopening
     # participants.
     #   -Jover, 04 September 2020
     net_id = record.get('netid')
-    net_id = net_id.strip().lower() if net_id else None
+    net_id = net_id.split('@')[0].strip().lower() if net_id else None
+    username = f'{net_id}@washington.edu' if net_id else None
```
tsibley, line 479: Hmm, after mentioning doctests, I just realized that this code will turn `xxx@gmail.com` → `xxx@washington.edu`. I think we want to avoid that. I know we have (had?) enrollments with @gmail.com addresses for the NetID. Maybe doctests with edge cases for this behaviour would be best after all!

tsibley (changes requested):

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
kschwabefry: This is true. The study team seems diligent in cleaning up the gmails etc (I don't see any in their reports now) but it'd be reasonable to detect a non-uw email and hash that as entered. I think the uw email domains are @u.washington.edu and @uw.edu

tsibley: Yep, I believe those are the two domains accepted. Instead of using the non-UW address, perhaps just fall back to hashing by the standard PII tuple?

kschwabefry:
Added doctests, and moved normalize() to redcap module for shared usage with non-scan husky etl

tsibley (approved):
Looks great, thanks! The doctests really help explicate and ensure the behaviour we want.

joverlee521 (approved):

lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+
+    if not net_id or net_id.isspace():
+        return None
+
+    net_id = net_id.strip().lower()
+
+    # if a uw email was entered, drop the domain before normalizing
+    if '@' in net_id:
+        domain = net_id.split('@')[1]
```
jccraft-uw, line 38: Could you split just once like net_id, domain =  net_id.split('@')

kairstenfay: That would only work if '@' is in the string. 
Otherwise, you'd get
```
a, b = 'aaaaa'.split('@')
ValueError: not enough values to unpack (expected 2, got 1)
```

kairstenfay (approved):
This looks great, Kristen, thanks!

lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -0,0 +1,45 @@
...
+
+    If missing netid or non-uw email provided, it cannot be normalized so return nothing
+    >>> normalize_net_id('notuw@gmail.com')
+    >>> normalize_net_id('nodomain@')
+    >>> normalize_net_id()
+    >>> normalize_net_id('')
+    >>> normalize_net_id(' ')
+    """
+
+    if not net_id or net_id.isspace():
```
kairstenfay, line 31: TIL about `isspace()`

lib/seattleflu/id3c/cli/command/etl/redcap.py
jccraft-uw: Line 37 ensures that:
if '@' in net_id:

kairstenfay: Ah I misunderstood. I thought you wanted to replace the conditional and the split with one line.

kschwabefry: @jccraft-uw I can parse the split list into `net_id, domain`, I'll just have to be more explicit about enforcing exactly one @ in the provided string and throw out edge case strings with multiple@at@signs

kschwabefry:
Added checking for provided netid strings with multiple @s, and condensed string splitting

lib/seattleflu/id3c/cli/command/etl/redcap.py
jccraft-uw: By no means should you spend time trying to do it this way. It just seemed like in its current state we could split once instead of twice.

tsibley: @kschwabefry Instead of checking for multiple `@`, you can provide a limit to `str.split`: 

```python
local_part, domain = net_id.split('@', 1)
```

kschwabefry:
Deployed.

PR #142, opened 2020-09-15
Presence absence to FHIR Bundles

joverlee521:
Creates a single FHIR bundle for each sample processed in the presence/absence ETL.
~Marking as draft since this needs to be deployed along with changes in the FHIR ETL in ID3C core.~
Needs to be deployed with https://github.com/seattleflu/id3c/pull/167

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/fhir_customizations.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir_customizations.py
@@ -1,0 +1,148 @@
+"""
+FHIR ETL functions that are custom for Seattle Flu Study.
+"""
+from id3c.cli.command.etl.fhir import *
+
+
+def check_specimen_for_update(db: DatabaseSession,
+                              sample: Any,
+                              specimen: Specimen):
+    """
+    Check if the given *specimen* has NWGC ID or an extension for sample
+    details and update the given *sample* details with these values if
+    they exist.
+
+    Returns the original or updated *sample*.
+    """
+    nwgc_id = identifier(specimen, f"{INTERNAL_SYSTEM}/nwgc_id").strip()
+    specimen_details = matching_extension_value(
+        extensions = specimen.extension,
+        url = f"{INTERNAL_SYSTEM}/sample_details",
+        value_type = "valueString"
+    )
+    # Update sample.details if the Specimen Resource contains a
+    # NWGC ID or an extension for Custom
+    if nwgc_id is not None or specimen_details is not None:
+        sample = update_sample_details(db, sample, specimen_details, nwgc_id)
+
+    return sample
+
+
+def update_sample_details(db: DatabaseSession,
+                          sample: Any,
+                          specimen_details: str = None,
+                          nwgc_id: str = None) -> Any:
+    """
+    Update the details for *sample* with additional details provided in
+    *specimen*.
+
+    The provided additional details are merged (at the top-level only) into
+    the existing sample details, if any.
+    """
+    additional_details = json.loads(specimen_details) if specimen_details else {}
+    if nwgc_id:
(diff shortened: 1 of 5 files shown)


lib/seattleflu/id3c/cli/command/etl/presence_absence.py
```diff
@@ -0,0 +1,449 @@
...
+
+2. If we receive a bogus "chip" value
+
+3. If we receive an unexpected value for the "controlStatus" of a target
+
+4. If we receive an unexpected value for the "targetResult" of a specific test
+
+5. If we receive an unexpected value for the "assayName"
+
+6. If we receive an unexpected value for the "assayType"
```
joverlee521, line 22: Worth noting that we have 6 conditions where the ETL will abort. These don't happen that often and seem to be pretty major things that we would immediately want to address. It would be great to hear others' opinions on whether any of these checks should be loosened to just skip with warnings instead of stopping the entire pipeline.

joverlee521:
[Re-engineer presence/absence ingest to split bundles from Samplify by sample and insert one FHIR bundle per sample](https://trello.com/c/zfZjRLic/757-re-engineer-presence-absence-ingest-to-split-bundles-from-samplify-by-sample-and-insert-one-fhir-bundle-per-sample)

kairstenfay:
Just did some local testing.
I cleared all processing_logs for `receiving.presence_absence` records with `presence_absence_id` > 1800 (arbitrarily).
I ran the presence-absence ETL :heavy_check_mark: 
Then I ran the FHIR ETL and got this error:
```
AssertionError: No SNOMED code found
```

joverlee521:
> Then I ran the FHIR ETL and got this error:
> 
> ```
> AssertionError: No SNOMED code found
> ```

Yup, this is expected. I'm still working on updating the FHIR ETL to ingest these new DiagnosticReport bundles 😅 

This particular error is due to the fact that we are using the custom NWGC target names instead of a standard SNOMED code. (convo on [slack](https://seattle-flu-study.slack.com/archives/C0173B4S1LY/p1599001869048000) from a ~while ago)

lib/seattleflu/id3c/cli/command/etl/presence_absence.py
```diff
@@ -96,283 +100,330 @@ def etl_presence_absence(*, db: DatabaseSession):
...
 
-                # Barcode must match a known identifier
-                received_sample_identifier = sample_identifier(db, received_sample_barcode)
-
-                if not received_sample_identifier:
-                    LOG.warning(f"Skipping results for sample without a known identifier «{received_sample_barcode}»")
-                    continue
+                # Guard against empty chip values
+                chip = received_sample.get("chip")
+                assert chip or "chip" not in received_sample, "Received bogus chip id"
```
kairstenfay, line 124: By "bogus", here we mean that there was a chip and it was a Falsy python value, right? e.g. `0`, `""`, `[]`, `{}`, `None`, ...

Are there any other kinds of bad values of chip we would want to guard against? Alternatively, what is the expected range of values for chip?

lib/seattleflu/id3c/cli/command/etl/fhir_customizations.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+        if extension_value:
+            try:
+                details[extension] = extension_value.as_json()
+            except AttributeError:
+                try:
+                    details[extension] = json.loads(extension_value)
+                except json.decoder.JSONDecodeError:
+                    details[extension] = extension_value
```
kairstenfay, line 139: Could you comment on why we try these three approaches in this order to parse an `extension_value`?

joverlee521:
Closing in favor of simpler unbundling method.

PR #143, opened 2020-09-17
Add ETL for UW reopening project

jccraft-uw:
This is the REDCap DET ETL for the UW reopening project. For this, I copied the code from redcap-det_scan. We should clean this up in the future by moving the code into a shared module, but for now it's a copy in here.
It's very likely that this code will have to be updated next week as REDCap fields change, but this is a working starting point.
There are many questionnaires in here!

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -1,0 +1,1233 @@
+"""
+Process DETs for the UW Reopening (Husky Coronavirus Testing) REDCap projects.
+"""
+import re
+import click
+import json
+import logging
+from uuid import uuid4
+from typing import Any, Callable, Dict, List, Mapping, Match, NamedTuple, Optional, Union, Tuple
+from cachetools import TTLCache
+from dateutil.relativedelta import relativedelta
+from decimal import Decimal
+from enum import Enum
+from id3c.db.session import DatabaseSession
+from id3c.cli.command.etl import redcap_det
+from id3c.cli.command.geocode import get_response_from_cache_or_geocoding
+from id3c.cli.command.location import location_lookup
+from id3c.cli.command.de_identify import generate_hash
+from id3c.cli.redcap import is_complete, Record as REDCapRecord
+from seattleflu.id3c.cli.command import age_ceiling
+from .redcap import normalize_net_id
+from .redcap_map import map_sex, map_symptom, UnknownVaccineResponseError
+from .fhir import *
+from . import race, first_record_instance, required_instruments
+
+
+LOG = logging.getLogger(__name__)
+
+
+class HuskyProject():
+    id: int
+    lang: str
+    api_token_env_var: str
+    command_name: str
+
+    def __init__(self, project_id: int, lang: str, api_token_env_var: str, command_name: str) -> None:
+        self.id = project_id
+        self.lang = lang
+        self.api_token_env_var = api_token_env_var
+        self.command_name = command_name
+
+PROJECTS = [
+        HuskyProject(23854, "en", "REDCAP_API_TOKEN", "uw-reopening")
+    ]
+
+LANGUAGE_CODE = {
+project.id: project.lang
+    for project in PROJECTS }
+
+class CollectionMethod(Enum):
+    SWAB_AND_SEND = 'swab_and_send'
+    KIOSK = 'kiosk'
+
+class EventType(Enum):
+    ENROLLMENT = 'enrollment'
+    ENCOUNTER = 'encounter'
+
+REVISION = 0
+
+REDCAP_URL = 'https://redcap.iths.org/'
+INTERNAL_SYSTEM = "https://seattleflu.org"
+ENROLLMENT_EVENT_NAME = "enrollment_arm_1"
+ENCOUNTER_EVENT_NAME = "encounter_arm_1"
+
+REQUIRED_ENROLLMENT_INSTRUMENTS = [
+    'eligibility_screening',
+    'consent_form',
+    'enrollment_questionnaire'
+]
+
+
+def command_for_each_project(function):
+    """
+    A decorator to register one redcap-det subcommand per REDCap project, each
+    calling the same base *function*.
+
+    Used for side-effects only; the original *function* is unmodified.
+    """
+    for project in PROJECTS:
+        help_message = "Process REDCap DETs for Husky Coronavirus Testing (UW reopening)"
+
+        redcap_det.command_for_project(
+            name = project.command_name,
+            redcap_url = REDCAP_URL,
+            project_id = project.id,
+            raw_coded_values = True,
+            revision = REVISION,
+            help = help_message,
+            include_incomplete = True)(function)
+
+    return function
+
+@command_for_each_project
+def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
+    redcap_record_instances: List[REDCapRecord]) -> Optional[dict]:
+
+    assert redcap_record_instances is not None and len(redcap_record_instances) > 0, \
+        "The redcap_record_instances list was not populated."
+
+    record_id = redcap_record_instances[0]['record_id']
+    project_id = redcap_record_instances[0].project.id
+
+    enrollments = [record for record in redcap_record_instances if record["redcap_event_name"] == ENROLLMENT_EVENT_NAME]
+    assert len(enrollments) == 1, \
+        f"Record had {len(enrollments)} enrollments."
+
+    enrollment = enrollments[0]
+
+    incomplete_enrollment_instruments = {
+                instrument
+                    for instrument
+                    in REQUIRED_ENROLLMENT_INSTRUMENTS
+                    if not is_complete(instrument, enrollment)
+            }
+
+    if incomplete_enrollment_instruments:
+        LOG.debug(f"The following required enrollment instruments «{incomplete_enrollment_instruments}» are not yet marked complete.")
+        return None
+
+    # If the participant's age < 18 ensure we have parental consent.
+    if (enrollment['core_age_years'] == "" or int(enrollment['core_age_years']) < 18) and \
+            (is_complete('parental_consent_form', enrollment) == False or enrollment['signature_parent'] == ''):
+        LOG.debug("The participant is < 18 years old and we do not have parental consent. Skipping record.")
+        return None
+
+    patient_entry, patient_reference = create_patient(enrollment)
+    birthdate = datetime.strptime(enrollment['core_birthdate'], '%Y-%m-%d')
+
+    if not patient_entry:
+        LOG.warning("Skipping record with insufficient information to construct patient")
+        return None
+
+    location_resource_entries = locations(db, cache, enrollment)
+    persisted_resource_entries = [patient_entry, *location_resource_entries]
+
+    for redcap_record_instance in redcap_record_instances:
+
+        event_type = None
+        collection_method = None
+
+        if redcap_record_instance["redcap_event_name"] == ENROLLMENT_EVENT_NAME:
+            event_type = EventType.ENROLLMENT
+        elif redcap_record_instance["redcap_event_name"] == ENCOUNTER_EVENT_NAME:
+            event_type = EventType.ENCOUNTER
+        else:
+            LOG.error(f"The record instance has an unexpected event name: {redcap_record_instance['redcap_event_name']}")
+            continue
+
+        if event_type == EventType.ENCOUNTER:
+            if is_complete('kiosk_registration_4c7f', redcap_record_instance):
+                collection_method = CollectionMethod.KIOSK
+            elif is_complete('test_fulfillment_form', redcap_record_instance):
+                collection_method = CollectionMethod.SWAB_AND_SEND
+
+        # site_reference refers to where the sample was collected
+        site_reference = create_site_reference(redcap_record_instance, collection_method, event_type)
+
+        initial_encounter_entry, initial_encounter_reference = create_encounter(
+            redcap_record_instance, patient_reference, site_reference,
+            location_resource_entries, event_type, collection_method)
+
+        # Skip the entire record if we can't create the enrollment encounter.
+        # Otherwise, just skip the record instance.
+        if not initial_encounter_entry:
+            if event_type == EventType.ENROLLMENT:
+                LOG.warning("Skipping record because we could not create the enrollment encounter")
+                return None
+            else:
+                LOG.warning("Skipping record instance with insufficient information to construct a initial encounter")
+                continue
+
+        specimen_entry = None
+        specimen_observation_entry = None
+        specimen_received = (collection_method == CollectionMethod.SWAB_AND_SEND and \
+            is_complete('post_collection_data_entry_qc', redcap_record_instance)) or \
+            (collection_method == CollectionMethod.KIOSK and \
+            is_complete('kiosk_registration_4c7f', redcap_record_instance))
+
+        if specimen_received:
+            specimen_entry, specimen_reference = create_specimen(redcap_record_instance, patient_reference, collection_method)
+            specimen_observation_entry = create_specimen_observation_entry(
+                specimen_reference, patient_reference, initial_encounter_reference)
+        else:
+            LOG.info("Creating encounter for record instance without sample")
+
+        if specimen_received and not specimen_entry:
+            LOG.warning("Skipping record instance with insufficent information to construct a specimen")
+            continue
+
+        computed_questionnaire_entry = None
+        enrollment_questionnaire_entry = None
+        daily_questionnaire_entry = None
+        testing_determination_internal_questionnaire_entry = None
+
+        computed_questionnaire_entry = create_computed_questionnaire_response(
+            redcap_record_instance, patient_reference, initial_encounter_reference,
+            birthdate, datetime.strptime(initial_encounter_entry['resource']['period']['start'], '%Y-%m-%d'))
+
+        if event_type == EventType.ENROLLMENT:
+            enrollment_questionnaire_entry = create_enrollment_questionnaire_response(
+            enrollment, patient_reference, initial_encounter_reference)
+        else:
+            testing_determination_internal_questionnaire_entry = \
+                create_testing_determination_internal_questionnaire_response(
+                redcap_record_instance, patient_reference, initial_encounter_reference)
+
+            daily_questionnaire_entry = \
+                create_daily_questionnaire_response(
+                redcap_record_instance, patient_reference, initial_encounter_reference)
+
+        follow_up_encounter_entry = None
+        follow_up_questionnaire_entry = None
+        follow_up_computed_questionnaire_entry = None
+
+        if event_type == EventType.ENCOUNTER and is_complete('week_followup', redcap_record_instance):
+            follow_up_encounter_entry, follow_up_encounter_reference = create_follow_up_encounter(
+                redcap_record_instance, patient_reference, site_reference, initial_encounter_reference)
+            follow_up_questionnaire_entry = create_follow_up_questionnaire_response(
+            redcap_record_instance, patient_reference, follow_up_encounter_reference)
+            follow_up_computed_questionnaire_entry = create_computed_questionnaire_response(
+            redcap_record_instance, patient_reference, follow_up_encounter_reference,
+            birthdate, datetime.strptime(follow_up_encounter_entry['resource']['period']['start'], '%Y-%m-%d'))
+
+        current_instance_entries = [
+            initial_encounter_entry,
+            computed_questionnaire_entry,
+            enrollment_questionnaire_entry,
+            testing_determination_internal_questionnaire_entry,
+            daily_questionnaire_entry,
+            specimen_entry,
+            specimen_observation_entry,
+            follow_up_encounter_entry,
+            follow_up_questionnaire_entry,
+            follow_up_computed_questionnaire_entry
+        ]
+
+        persisted_resource_entries.extend(list(filter(None, current_instance_entries)))
+
+
+    return create_bundle_resource(
+        bundle_id = str(uuid4()),
+        timestamp = datetime.now().astimezone().isoformat(),
+        source = f"{REDCAP_URL}{project_id}/{record_id}",
+        entries = list(filter(None, persisted_resource_entries))
+    )
+
+
+def create_site_reference(record: dict, collection_method: CollectionMethod, event_type: EventType) -> Optional[Dict[str,dict]]:
+    """
+    Create a Location reference for site of the sample collection encounter based
+    on how the sample was collected.
+    """
+    if collection_method == CollectionMethod.KIOSK:
+        record_location = record.get('location_type')
+        if record_location:
+            site = site_map(record_location)
+    else:
+        site = 'UWReopeningSwabNSend'
+
+    return {
+        "location": create_reference(
+            reference_type = "Location",
+            identifier = create_identifier(f"{INTERNAL_SYSTEM}/site", site)
+        )
+    }
+
+
+def site_map(record_location: str) -> str:
+    """
+    Maps *record_location* to the corresponding site name.
+    """
+    location_site_map = {
+        'bothell':  'UWBothell',
+        'odegaard': 'UWOdegaardLibrary',
+        'slu':      'UWSouthLakeUnion',
+        'tacoma':   'UWTacoma',
+        'uw_club':  'UWClub'
+    }
+
+    if record_location not in location_site_map:
+        raise UnknownRedcapRecordLocation(f"Found unknown location type «{record_location}»")
+
+    return location_site_map[record_location]
+
+
+def locations(db: DatabaseSession, cache: TTLCache, record: dict) -> list:
+    """ Creates a list of Location resource entries from a REDCap record. """
+    lodging_options = [
+        'shelter',
+        'afl',
+        'snf',
+        'ltc',
+        'be',
+        'pst',
+        'cf',
+        'none'
+    ]
+
+    if record.get('core_housing_type') in lodging_options:
+        housing_type = 'lodging'
+    else:
+        housing_type = 'residence'
+
+    address = {
+        'street': record['core_home_street'],
+        'secondary': record['core_apartment_number'],
+        'city': record['core_home_city'],
+        'state': record['core_home_state'],
+        'zipcode': record['core_zipcode'],
+    }
+
+    lat, lng, canonicalized_address = get_response_from_cache_or_geocoding(address, cache)
+    if not canonicalized_address:
+        return []  # TODO
+
+    tract_location = residence_census_tract(db, (lat, lng), housing_type)
+    # TODO what if tract_location is null?
+    tract_full_url = generate_full_url_uuid()
+    tract_entry = create_resource_entry(tract_location, tract_full_url)
+
+    address_hash = generate_hash(canonicalized_address)
+    address_location = create_location(
+        f"{INTERNAL_SYSTEM}/location/address",
+        address_hash,
+        housing_type,
+        tract_full_url
+    )
+    address_entry = create_resource_entry(address_location, generate_full_url_uuid())
+
+    return [tract_entry, address_entry]
+
+
+def residence_census_tract(db: DatabaseSession, lat_lng: Tuple[float, float],
+    housing_type: str) -> Optional[dict]:
+    """
+    Creates a new Location Resource for the census tract containing the given
+    *lat_lng* coordintes and associates it with the given *housing_type*.
+    """
+    location = location_lookup(db, lat_lng, 'tract')
+
+    if location and location.identifier:
+        return create_location(
+            f"{INTERNAL_SYSTEM}/location/tract", location.identifier, housing_type
+        )
+    else:
+        LOG.warning("No census tract found for given location.")
+        return None
+
+
+def create_location(system: str, value: str, location_type: str, parent: str=None) -> dict:
+    """ Returns a FHIR Location resource. """
+    location_type_system = "http://terminology.hl7.org/CodeSystem/v3-RoleCode"
+    location_type_map = {
+        "residence": "PTRES",
+        "school": "SCHOOL",
+        "work": "WORK",
+        "site": "HUSCS",
+        "lodging": "PTLDG",
+    }
+
+    location_type_cc = create_codeable_concept(location_type_system,
+                                            location_type_map[location_type])
+    location_identifier = create_identifier(system, value)
+    part_of = None
+    if parent:
+        part_of = create_reference(reference_type="Location", reference=parent)
+
+    return create_location_resource([location_type_cc], [location_identifier], part_of)
+
+
+def create_patient(record: REDCapRecord) -> tuple:
+    """ Returns a FHIR Patient resource entry and reference. """
+    gender = map_sex(record['core_sex'])
+
+    language_codeable_concept = create_codeable_concept(
+        system = 'urn:ietf:bcp:47',
+        code = LANGUAGE_CODE[record.project.id]
+    )
+    communication = [{
+        'language' : language_codeable_concept,
+        'preferred': True # Assumes that the project language is the patient's preferred language
+    }]
+
+    net_id = normalize_net_id(record.get('netid'))
+    if net_id:
+        patient_id = generate_hash(net_id)
+    else:
+        patient_id = generate_patient_hash(
+            names       = (record['core_participant_first_name'], record['core_participant_last_name']),
+            gender      = gender,
+            birth_date  = record['core_birthdate'],
+            postal_code = record['core_zipcode'])
+
+    if not patient_id:
+        # Some piece of information was missing, so we couldn't generate a
+        # hash.  Fallback to treating this individual as always unique by using
+        # the REDCap record id.
+        patient_id = generate_hash(f"{REDCAP_URL}{record.project.id}/{record['record_id']}")
+
+    LOG.debug(f"Generated individual identifier {patient_id}")
+
+    patient_identifier = create_identifier(f"{INTERNAL_SYSTEM}/individual", patient_id)
+    patient_resource = create_patient_resource([patient_identifier], gender, communication)
+
+    return create_entry_and_reference(patient_resource, "Patient")
+
+
+def create_encounter(record: REDCapRecord, patient_reference: dict,
+    site_reference: dict, locations: list, event_type: EventType,
+    collection_method: CollectionMethod) -> tuple:
+    """
+    Returns a FHIR Encounter resource entry and reference for the encounter in the study.
+    """
+
+    def grab_symptom_key(key: str, variable_name: str) -> Optional[Match[str]]:
+        if record[key] == '1':
+            return re.match(f"{variable_name}___[a-z_]+", key)
+        else:
+            return None
+
+    def build_condition(symptom: str, onset_date: str) -> dict:
+        return create_resource_condition(record, symptom, patient_reference, onset_date)
+
+    def build_diagnosis(symptom: str) -> Optional[dict]:
+        mapped_symptom = map_symptom(symptom)
+        if not mapped_symptom:
+            return None
+
+        return { "condition": { "reference": f"#{mapped_symptom}" } }
+
+    def build_locations_list(location: dict) -> dict:
+        return {
+            "location": create_reference(
+                reference_type = "Location",
+                reference = location["fullUrl"]
+            )
+        }
+
+    def non_tract_locations(resource: dict):
+        return bool(resource) \
+            and resource['resource']['identifier'][0]['system'] != f"{INTERNAL_SYSTEM}/location/tract"
+
+    # Map the various symptoms variables to their onset date.
+    # For daily_symptoms_covid_like we don't know the actual onset date. The questions asks
+    # "in the past 24 hours"
+    if event_type == EventType.ENCOUNTER:
+        symptom_onset_map = {
+            'daily_symptoms_covid_like': None,
+            'symptoms': record['symptom_onset'],
+            'symptoms_kiosk': record['symptom_duration_kiosk'],
+            'symptoms_swabsend': record['symptom_duration_swabsend']
+        }
+    elif event_type == EventType.ENROLLMENT:
+        symptom_onset_map = {'symptoms_base': record['symptom_onset_base']}
+
+    contained = []
+    diagnosis = []
+
+    for symptom_variable in symptom_onset_map:
+        symptom_keys = []
+
+        for redcap_key in record.keys():
+            symptom_key = grab_symptom_key(redcap_key, symptom_variable)
+            if symptom_key:
+                symptom_keys.append(symptom_key.string)
+
+        symptoms = list(map(lambda x: re.sub('[a-z_]+___', '', x), symptom_keys))
+        for symptom in symptoms:
+            contained.append(build_condition(symptom, symptom_onset_map[symptom_variable]))
+            diagnosis.append(build_diagnosis(symptom))
+
+    encounter_identifier = create_identifier(
+        system = f"{INTERNAL_SYSTEM}/encounter",
+        value = f"{REDCAP_URL}{record.project.id}/{record['record_id']}/{record['redcap_event_name']}/{record['redcap_repeat_instance']}"
+    )
+
+    collection_code = None
+
+    # See https://terminology.hl7.org/1.0.0/CodeSystem-v3-ActCode.html for
+    # possible collection codes.
+    # HH = 'home health'
+    # FLD = 'field'
+    if event_type == EventType.ENROLLMENT or collection_method == CollectionMethod.SWAB_AND_SEND:
+        collection_code = "HH"
+    elif collection_method == CollectionMethod.KIOSK:
+        collection_code = "FLD"
+
+    # For the encounter_date for an ENCOUNTER, first try the attestation_date
+    # from the daily attestation survey then try nasal_swab_timestamp from
+    # the kiosk registration and finally the swab-and-send order date.
+    encounter_date = None
+
+    if event_type == EventType.ENCOUNTER:
+        if record.get('attestation_date'):
+            encounter_date = record.get('attestation_date')
+        elif record.get('nasal_swab_timestamp'):
+            encounter_date = datetime.strptime(record.get('nasal_swab_timestamp'),
+                '%Y-%m-%d %H:%M:%S').strftime('%Y-%m-%d')
+        elif record.get('time_test_order'):
+            encounter_date = datetime.strptime(record.get('time_test_order'),
+                '%Y-%m-%d %H:%M:%S').strftime('%Y-%m-%d')
+
+    elif event_type == EventType.ENROLLMENT:
+        encounter_date = record.get('enrollment_date')
+
+    if not encounter_date:
+        return None, None
+
+    encounter_class_coding = create_coding(
+        system = "http://terminology.hl7.org/CodeSystem/v3-ActCode",
+        code = collection_code
+    )
+
+    non_tracts = list(filter(non_tract_locations, locations))
+    non_tract_references = list(map(build_locations_list, non_tracts))
+    # Add hard-coded site Location reference
+    if site_reference:
+        non_tract_references.append(site_reference)
+
+    encounter_resource = create_encounter_resource(
+        encounter_identifier = [encounter_identifier],
+        encounter_class = encounter_class_coding,
+        encounter_date = encounter_date,
+        patient_reference = patient_reference,
+        location_references = non_tract_references,
+        diagnosis = diagnosis,
+        contained = contained
+    )
+
+    return create_entry_and_reference(encounter_resource, "Encounter")
+
+
+def create_resource_condition(record: dict, symptom_name: str, patient_reference: dict, onset_date:str) -> Optional[dict]:
+    """ Returns a FHIR Condition resource. """
+    mapped_symptom_name = map_symptom(symptom_name)
+    if not mapped_symptom_name:
+        return None
+
+    # XXX TODO: Define this as a TypedDict when we upgrade from Python 3.6 to
+    # 3.8.  Until then, there's no reasonable way to type this data structure
+    # better than Any.
+    #   -trs, 24 Oct 2019
+    condition: Any = {
+        "resourceType": "Condition",
+        "id": f'{mapped_symptom_name}',
+        "code": {
+            "coding": [
+                {
+                    "system": f"{INTERNAL_SYSTEM}/symptom",
+                    "code": mapped_symptom_name
+                }
+            ]
+        },
+        "subject": patient_reference
+    }
+
+    if onset_date:
+        condition["onsetDateTime"] = onset_date
+
+    return condition
+
+
+def create_specimen(record: dict, patient_reference: dict, collection_method: CollectionMethod) -> tuple:
+    """ Returns a FHIR Specimen resource entry and reference """
+
+    def specimen_barcode() -> Optional[str]:
+        """
+        Return specimen barcode from REDCap record.
+        """
+         # Normalize all barcode fields upfront.
+        barcode_fields = {
+            # Kiosk related
+            "collect_barcode_kiosk", # from	kiosk_registration_4c7f
+
+            # Swab and Send related
+            "pre_scan_barcode", # from test_fulfillment_form
+            "barcode_swabsend",	# from husky_test_kit_registration
+            "return_utm_barcode", # from post_collection_data_entry_qc
+        }
+
+        for barcode_field in barcode_fields:
+            record[barcode_field] = record[barcode_field].strip().lower()
+
+        return record["collect_barcode_kiosk"] or record["return_utm_barcode"] or \
+            record["pre_scan_barcode"] or None
+
+
+    barcode = specimen_barcode()
+
+    if not barcode:
+        LOG.warning("Could not create Specimen Resource due to lack of barcode.")
+        return None, None
+
+    specimen_identifier = create_identifier(
+        system = f"{INTERNAL_SYSTEM}/sample",
+        value = barcode
+    )
+
+    # YYYY-MM-DD HH:MM:SS in REDCap
+    received_time = record['samp_process_date'].split()[0] if record.get('samp_process_date') else None
+
+    note = None
+
+    if record['able_to_test'] == 'no':
+        note = 'never-tested'
+    else:
+        note = 'can-test'
+
+    specimen_type = 'NSECR'  # Nasal swab.
+    specimen_resource = create_specimen_resource(
+        [specimen_identifier], patient_reference, specimen_type, received_time,
+        collection_date(record, collection_method), note
+    )
+
+    return create_entry_and_reference(specimen_resource, "Specimen")
+
+
+def collection_date(record: dict, collection_method: CollectionMethod) -> Optional[str]:
+    """
+    Determine sample/specimen collection date from the given REDCap *record*.
+    """
+    if collection_method == CollectionMethod.KIOSK:
+        return record["nasal_swab_q"]
+    elif collection_method == CollectionMethod.SWAB_AND_SEND:
+        return record["date_on_tube"] or record["kit_reg_date"] or record["back_end_scan_date"]
+    else:
+        return None
+
+def combine_multiple_fields(record: Dict[Any, Any], field_prefix: str, field_suffix: str = "") -> Optional[List]:
+        """
+        Handles the combining of multiple fields asking the same question such
+        as country and state traveled.
+        """
+        regex = rf'^{re.escape(field_prefix)}[0-9]+{re.escape(field_suffix)}$'
+        empty_value = ''
+        answered_fields = list(filter(lambda f: filter_fields(f, record[f], regex, empty_value), record))
+
+        if not answered_fields:
+            return None
+
+        return list(map(lambda x: record[x], answered_fields))
+
+
+
+def create_enrollment_questionnaire_response(record: dict, patient_reference: dict,
+                                            encounter_reference: dict) -> Optional[dict]:
+    """
+    Returns a FHIR Questionnaire Response resource entry for the enrollment
+    encounter (i.e. encounter of enrollment into the study)
+    """
+
+    # Do not include `core_age_years` because we calculate the age ourselves in the computed questionnaire.
+    integer_questions = [
+        'weight',
+        'height_total',
+        'tier'
+    ]
+
+    string_questions = [
+        'text_or_email',
+        'text_or_email_attestation',
+        'campus_location',
+        'affiliation',
+        'student_level',
+        'sea_employee_type',
+        'core_house_members',
+        'core_education',
+        'core_income',
+        'wfh_base',
+        'core_housing_type',
+        'core_health_risk',
+        'core_tobacco_use',
+        'sought_care_base',
+        'hospital_where_base',
+        'hospital_ed_base',
+        'prior_test_positive_base',
+        'prior_test_type_base',
+        'prior_test_result_base',
+        'contact_base',
+        'wash_hands_base',
+        'clean_surfaces_base',
+        'hide_cough_base',
+        'mask_base',
+        'distance_base',
+        'novax_reason',
+        'covid_vaccine',
+        'covid_novax_reason',
+        'countries_visited_base',
+        'states_visited_base',
+        'alerts_off',
+        'pronouns',
+        'on_campus_freq',
+        'vaccine_method',
+        'vaccine_where'
+    ]
+
+    date_questions = [
+        'today_consent',
+        'enrollment_date_time',
+        'hospital_arrive_base',
+        'hospital_leave_base',
+        'prior_test_positive_date_base'
+    ]
+
+    boolean_questions = [
+        'study_area',
+        'attend_uw',
+        'english_speaking',
+        'athlete',
+        'uw_medicine_yesno',
+        'inperson_classes',
+        'uw_job',
+        'uw_greek_member',
+        'live_other_uw',
+        'uw_apt_yesno',
+        'core_pregnant',
+        'core_latinx',
+        'mobility',
+        'vaccine_hx',
+        'hall_health',
+        'prior_test_base',
+        'travel_countries_phs_base',
+        'travel_states_phs_base',
+        'swab_and_send_calc',
+        'kiosk_calc',
+        'covid_test_week_base'
+    ]
+
+    decimal_questions = [
+        'bmi'
+    ]
+
+    coding_questions = [
+        'core_race'
+    ]
+
+    # Do some pre-processing
+    # Combine checkbox answers into one list
+    checkbox_fields = [
+        'core_race',
+        'core_health_risk',
+        'core_tobacco_use',
+        'sought_care_base',
+        'prior_test_type_base',
+        'prior_test_positive_base',
+        'contact_base'
+    ]
+
+    question_categories = {
+        'valueCoding': coding_questions,
+        'valueBoolean': boolean_questions,
+        'valueInteger': integer_questions,
+        'valueString': string_questions,
+        'valueDate': date_questions,
+        'valueDecimal': decimal_questions
+    }
+
+    for field in checkbox_fields:
+        record[field] = combine_checkbox_answers(record, field)
+
+    # Combine all fields answering the same question
+    record['countries_visited_base'] = combine_multiple_fields(record, 'country', '_base')
+    record['states_visited_base'] = combine_multiple_fields(record, 'state', '_base')
+
+    # Age Ceiling
+    try:
+        record['core_age_years'] = age_ceiling(int(record['core_age_years']))
+    except ValueError:
+        record['core_age_years'] = record['core_age_years'] = None
+
+    # Set the study tier
+    tier = None
+    if record['tier_1'] == '1':
+        tier = 1
+    elif record['tier_2'] == '1':
+        tier = 2
+    elif record['tier_3'] == '1':
+        tier = 3
+    record['tier'] = tier
+
+    return questionnaire_response(record, question_categories, patient_reference, encounter_reference, True)
+
+
+def filter_fields(field: str, field_value: str, regex: str, empty_value: str) -> bool:
+    """
+    Function that filters for *field* matching given *regex* and the
+    *field_value* must not equal the expected *empty_value.
+    """
+    if re.match(regex, field) and field_value != empty_value:
+        return True
+
+    return False
+
+
+def combine_checkbox_answers(record: dict, coded_question: str) -> Optional[List]:
+    """
+    Handles the combining "select all that apply"-type checkbox
+    responses into one list.
+
+    Uses our in-house mapping for race and symptoms
+    """
+    regex = rf'{re.escape(coded_question)}___[\w]*$'
+    empty_value = '0'
+    answered_checkboxes = list(filter(lambda f: filter_fields(f, record[f], regex, empty_value), record))
+    # REDCap checkbox fields have format of {question}___{answer}
+    answers = list(map(lambda k: k.replace(f"{coded_question}___", ""), answered_checkboxes))
+
+    if coded_question == 'race':
+        return race(answers)
+
+    if re.match(r'fu_[1-4]_symptoms$', coded_question):
+        return list(map(lambda a: map_symptom(a), answers))
+
+    return answers
+
+def map_vaccine(vaccine_response: str) -> Optional[bool]:
+    """
+    Maps a vaccine response to FHIR immunization status codes
+    (https://www.hl7.org/fhir/valueset-immunization-status.html)
+    """
+    vaccine_map = {
+        'yes': True,
+        'no': False,
+        'dont_know': None,
+        '': None
+    }
+
+    if vaccine_response not in vaccine_map:
+        raise UnknownVaccineResponseError(f"Unknown vaccine response «{vaccine_response}»")
+
+    return vaccine_map[vaccine_response]
+
+def vaccine(record: Any) -> Optional[dict]:
+    """
+    For a given *record*, return a questionnaire response item with the vaccine
+    response(s) encoded.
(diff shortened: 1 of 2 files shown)


joverlee521 (changes requested):
On first pass, the new ETL needs to be imported in `/etl/__init__.py` to get registered as an ETL command.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+
+    if record_location not in location_site_map:
+        raise UnknownRedcapRecordLocation(f"Found unknown location type «{record_location}»")
+
+    return location_site_map[record_location]
+
+
+def locations(db: DatabaseSession, cache: TTLCache, record: dict) -> list:
+    """ Creates a list of Location resource entries from a REDCap record. """
+    lodging_options = [
```
kschwabefry, line 277: Adding dorm into this location list could be a nice-to-have

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        system = f"{INTERNAL_SYSTEM}/encounter",
+        value = f"{REDCAP_URL}{record.project.id}/{record['record_id']}/{record['redcap_repeat_instance']}"
+    )
+
+    collection_code = None
+
+    if event_type == EventType.ENROLLMENT or collection_method == CollectionMethod.SWAB_AND_SEND:
+        collection_code = "HH"
+    elif collection_method == CollectionMethod.KIOSK:
+        collection_code = "FLD"
```
kschwabefry, line 621: What do HH and FLD stand for here? HH=Household?

kschwabefry (commented):
I think its very reasonable to copy in the code from the scan etl for now given deadlines :)
While its fresh in your mind, it may be useful to quickly go through and annotate refactoring todos on methods that should be shared in common codebase. Ymmv, so no pressure if you feel like its not worth the time now.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+def create_enrollment_questionnaire_response(record: dict, patient_reference: dict,
+                                            encounter_reference: dict) -> Optional[dict]:
+    """
+    Returns a FHIR Questionnaire Response resource entry for the enrollment
+    encounter (i.e. encounter of enrollment into the study)
+    """
+
+    # Do not include `core_age_years` because we calculate the age ourselves in the computed questionnaire.
+    integer_questions = [
+        'core_zipcode',
```
joverlee521, line 770: Zipcode is considered PII, since our mapping of code -> zipcode is public, we shouldn't ingest this.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        'weight',
+        'height_ft',
+        'height_in',
+        'height_ft_calc',
+        'height_total',
```
joverlee521, line 775: These fields seem redundant, we can ingest one field (`height_total`?) and do conversions to ft/in when we need it.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        'vaccine_month',
+        'vaccine_year'
```
joverlee521, line 777: Vaccine questionnaire responses are a bit weird and have been handled specially in the past. See: [swab-and-send etl](https://github.com/seattleflu/id3c-customizations/blob/master/lib/seattleflu/id3c/cli/command/etl/redcap_det_swab_n_send.py#L529)

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+    string_questions = [
+        'study_area',
+        'text_or_email',
+        'text_or_email_attestation',
+        'campus_location',
+        'student_or_employee',
+        'student_level',
+        'sea_employee_type',
+        'uw_greek_house',
+        'core_home_state',
```
joverlee521, line 789: I don't think we don't need to ingest state since we can get this information from their location record in `warehouse.location`.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        'core_house_members',
+        'dorm',
+        'uw_apt_names',
+        'core_sex',
```
joverlee521, line 793: Gender is already part of the the patient resource so no need to ingest this field again.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        'country1_base',
+        'country2_base',
+        'country3_base',
+        'country4_base',
+        'country5_base',
+        'state1_base',
+        'state2_base',
+        'state3_base',
+        'state4_base',
+        'state5_base',
```
joverlee521, line 822: These fields are redundant since you are already combining them below.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        'alerts_off',
+        'bmi',
+        'novax_reason',
+        'covid_vacine',
```
joverlee521, line 826: Field has been updated to `covid_vaccine`

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+    return questionnaire_response(record, question_categories, patient_reference, encounter_reference)
+
+def create_daily_questionnaire_response(record: dict, patient_reference: dict,
+                                            encounter_reference: dict) -> Optional[dict]:
+    """
+    Returns a FHIR Questionnaire Response resource entry for the daily
+    attestation questionnaire.
+    """
+    boolean_questions = [
+        'daily_symptoms',
```
joverlee521, line 1224: No need to ingest symptoms here since it's already part of encounter.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        'state4_base',
+        'state5_base',
+        'alerts_off',
+        'bmi',
+        'novax_reason',
+        'covid_vacine',
+        'vaccine_where',
+        'on_campus_freq',
+        'covid_novax_reason',
+        'symptoms_base',
```
joverlee521, line 830: No need to ingest symptoms here since it's already part of encounter.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        'states_visited_base'
+    ]
+
+    date_questions = [
+        'today_consent',
+        'enrollment_date',
+        'hospital_arrive_base',
+        'hospital_leave_base',
+        'prior_test_positive_date_base',
+        'symptom_onset_base'
```
joverlee521, line 841: No need to ingest symptoms here since it's already part of encounter.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        'country1',
+        'country2',
+        'country3',
+        'country4',
+        'country5',
+        'state1',
+        'state2',
+        'state3',
+        'state4',
+        'state5',
```
joverlee521, line 1261: Ditto redundant fields since they are getting combined below.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+    question_categories = {
+        'valueBoolean': boolean_questions,
+        'valueString': string_questions,
+        'valueDate': date_questions
+    }
+
+    # Combine checkbox answers into one list
+    checkbox_fields = [
+        'daily_symptoms_covid_like',
+        'symptoms',
```
joverlee521, line 1283: No need to ingest symptoms here since it's already part of encounter.

joverlee521 (changes requested):
Most of my comments are regarding questionnaire response fields that I think are redundant, feel free to tackle those comments are a later date when we aren't in a crunch 😄

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        return bool(resource) \
+            and resource['resource']['identifier'][0]['system'] != f"{INTERNAL_SYSTEM}/location/tract"
+
+    # Map the various symptoms variables to their onset date.
+    # For daily_symptoms_covid_like use the attestation_date - 24 hours because the question asks
+    # "in the past 24 hours"
+    if event_type == EventType.ENCOUNTER:
+        symptom_onset_map = {
+            'daily_symptoms_covid_like': (datetime.strptime(record['attestation_date'], '%Y-%m-%d') \
+                - timedelta(hours=24)).strftime('%Y-%m-%d'),
```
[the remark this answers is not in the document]
jccraft-uw: I thought about this and went with ingesting the daily_symptoms_covid_like as questionnaire items even though they do also get captured as symptoms just for ease of use.
@joverlee521 We need an onset date for these symptoms, I believe. Or can we omit one for these?
If we do have to supply an onset date, what do you suggest for these since we don't ask the participant?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: As with the previous set, I chose to ingest these as questionnaire items even though we also capture them as symptoms for convenience.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Same as above.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Same as above. Ingesting them as questionnaire items for convenience.

jccraft-uw:
> On first pass, the new ETL needs to be imported in `/etl/__init__.py` to get registered as an ETL command.

Added!

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Good question. I added a comment about those. HH = Household and FLD = Field

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        # site_reference refers to where the sample was collected
+        site_reference = create_site_reference(redcap_record_instance, collection_method, event_type)
+
+        initial_encounter_entry, initial_encounter_reference = create_encounter(
+            redcap_record_instance, patient_reference, site_reference,
+            location_resource_entries, event_type, collection_method)
+
+        if not initial_encounter_entry:
+            LOG.warning("Skipping record instance with insufficient information to construct a initial encounter")
+            return None
```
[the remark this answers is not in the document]
jccraft-uw: Sheeze. Thanks for catching that. That was carry over from before there was a loop.
Replaced with 'continue'.
I took you suggestion and stopped the whole record if we can't create the enrollment encounter.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+        if specimen_received:
+            specimen_entry, specimen_reference = create_specimen(redcap_record_instance, patient_reference, collection_method)
+            specimen_observation_entry = create_specimen_observation_entry(
+                specimen_reference, patient_reference, initial_encounter_reference)
+        else:
+            LOG.info("Creating encounter for record instance without sample")
+
+        if specimen_received and not specimen_entry:
+            LOG.warning("Skipping record instance with insufficent information to construct a specimen")
+            return None
```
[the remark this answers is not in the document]
jccraft-uw: Fixed!

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I left these intentionally, but now that I see that they get combined into an array there's no reason to keep them! Fixed.

jccraft-uw: I removed the individual country/state entries for the enrollment questionnaire and the daily questionnaire.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+    def non_tract_locations(resource: dict):
+        return bool(resource) \
+            and resource['resource']['identifier'][0]['system'] != f"{INTERNAL_SYSTEM}/location/tract"
+
+    # Map the various symptoms variables to their onset date.
+    # For daily_symptoms_covid_like use the attestation_date - 24 hours because the question asks
+    # "in the past 24 hours"
+    if event_type == EventType.ENCOUNTER:
+        symptom_onset_map = {
+            'daily_symptoms_covid_like': (datetime.strptime(record['attestation_date'], '%Y-%m-%d') \
```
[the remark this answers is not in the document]
jccraft-uw: Removed since we're not using an onset date for these anymore.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Likewise, I want to ingest them for convenience.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Removed!

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Fair point. I kept height_total.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: The theme will my responses to all these is that just it's someplace else related to the encounter, but for convenience I want to ingest it here too.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: The theme will my responses to all these is that just it's someplace else related to the encounter, but for convenience I want to ingest it here too.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Removed!

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Thanks so much for seeing that. Fixed.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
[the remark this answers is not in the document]
jccraft-uw: @joverlee521 Using None works throughout the entire pipeline. The FHIR ETL just doesn't set the onset date. 
So I removed this and am using None for the daily_symptoms_covid_like onset date.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Thanks. I did it a similar way. The REDCap values don't match those, so I created a map in this ETL.

jccraft-uw:
> I think its very reasonable to copy in the code from the scan etl for now given deadlines :)
> While its fresh in your mind, it may be useful to quickly go through and annotate refactoring todos on methods that should be shared in common codebase. Ymmv, so no pressure if you feel like its not worth the time now.

It's most of the code unfortunately. In a lot of functions I had to tweak logic slightly because variable names or values were different, though.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: @seattleflu/id3c-devs , would dorm name fit into locations? Right now we're using housing_type (residence or lodging) and the physical address.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1312 @@
...
+
+def site_map(record_location: str) -> str:
+    """
+    Maps *record_location* to the corresponding site name.
+    """
+    location_site_map = {
+        'bothell':  'UWBothell',
+        'odegaard': 'UWOdegaardLibrary',
+        'slu':      'UWSouthLakeUnion',
+        'tacoma':   'UWTacoma',
```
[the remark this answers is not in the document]
jccraft-uw: Thanks. I will create them in a separate PR for the SQL.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: On second thought, it's weird to have only state. I removed it from the questionnaire parsing.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I just went through all questionnaires with the latest data dictionary and fixed any discrepancies. I suspect I'll be doing this again before launch.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
joverlee521: What do you mean by convenience?

jccraft-uw: Convenience when someone wants to look at the questionnaire responses. The symptoms question is a question in the questionnaire. Yes, it happens to be symptoms and we process those elsewhere, but the symptoms question is a question in the questionnaire.

joverlee521: All data we collect comes from the questionnaire, but we add them to the appropriate FHIR resources. We use the QuestionnaireResponse resource to hold data that does not fit into any other FHIR resource.

jccraft-uw: You seem passionate that I should not include the symptom questions as questionnaire items because we store them as symptoms. My point is that they are questions in the surveys, so why not include them as questionnaire items? I'll take them out if this if this really bothers you.

kairstenfay: > We use the QuestionnaireResponse resource to hold data that does not fit into any other FHIR resource.

Jover's right. We want our FHIR bundles to be as clean as possible. The definition of [data cleansing](https://en.wikipedia.org/wiki/Data_cleansing) includes removing duplicate representations of the same entity. Therefore, we should only represent gender, symptoms, etc. once in a Bundle for each encounter.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: That's a good question. I remember on a recent call that we were saying dorm probably counts as PII, so it might not be a good idea to ingest it directly. Perhaps I'm not remembering correctly, though.

jccraft-uw: On a call last week we talked about not ingesting dorm_room (the actual dorm room number). I am currently ingesting the dorm name, the greek house name, and the UW apartment building name. It would be handy to do analysis of that data in ID3C (for surges, for example).  But if we don't ingest things like zipcode out of PII concern, it would seem that I should not bring in data that would identify an actual building. @kairstenfay @joverlee521 , do you agree with me *not* bringing in building names like those above?

We could still use the dorm to create a location even if we didn't store the dorm name itself, so that's still an open question.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+            continue
+
+        computed_questionnaire_entry = None
+        enrollment_questionnaire_entry = None
+        daily_questionnaire_entry = None
+        testing_determination_internal_questionnaire_entry = None
+
+        computed_questionnaire_entry = create_computed_questionnaire_response(
+            redcap_record_instance, patient_reference, initial_encounter_reference,
+            birthdate, datetime.strptime(initial_encounter_entry['resource']['period']['start'], '%Y-%m-%d'))
```
kairstenfay, line 195: `datetime.strptime(initial_encounter_entry['resource']['period']['start'], '%Y-%m-%d')` seems like something the `create_computed_questionnaire_response()` function can know how to do rather than performing a complicated parse on the fly when passing this value as an argument.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+        if event_type == EventType.ENCOUNTER and is_complete('week_followup', redcap_record_instance):
+            follow_up_encounter_entry, follow_up_encounter_reference = create_follow_up_encounter(
+                redcap_record_instance, patient_reference, site_reference, initial_encounter_reference)
+            follow_up_questionnaire_entry = create_follow_up_questionnaire_response(
+            redcap_record_instance, patient_reference, follow_up_encounter_reference)
+            follow_up_computed_questionnaire_entry = create_computed_questionnaire_response(
+            redcap_record_instance, patient_reference, follow_up_encounter_reference,
+            birthdate, datetime.strptime(follow_up_encounter_entry['resource']['period']['start'], '%Y-%m-%d'))
+
+        current_entries = [
```
kairstenfay, line 222: The `resource_entries` / `current_entries` variable lists here are a little difficult for me to understand. Perhaps there's some clearer language we can use to distinguish between them? Assuming I have the right idea here, some suggestions that come to mind are `persistent_resource_entries` versus `instance_specific_entries`.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+        computed_questionnaire_entry = create_computed_questionnaire_response(
+            redcap_record_instance, patient_reference, initial_encounter_reference,
+            birthdate, datetime.strptime(initial_encounter_entry['resource']['period']['start'], '%Y-%m-%d'))
+
+        if event_type == EventType.ENROLLMENT:
+            enrollment_questionnaire_entry = create_enrollment_questionnaire_response(
+            enrollment, patient_reference, initial_encounter_reference)
+        else:
+            testing_determination_internal_questionnaire_entry = \
+                create_testing_determination_internal_questionnaire_response(
```
kairstenfay, line 202: I'm not sure I understand why this testing determination field is being handled differently and put into a separate QuestionnaireResponse Resource than the others.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+                redcap_record_instance, patient_reference, initial_encounter_reference)
+
+            daily_questionnaire_entry = \
+                create_daily_questionnaire_response(
```
kairstenfay, line 206: Same here: it's not clear to me the motivation for processing these QuestionnaireResponse items separately from the others.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+        encounter_class = encounter_class_coding,
+        encounter_date = encounter_date,
+        patient_reference = patient_reference,
+        location_references = [site_reference],
+        reason_code = [encounter_reason_code],
+        part_of = initial_encounter_reference
+    )
+
+    return create_entry_and_reference(encounter_resource, "Encounter")
+
```
kairstenfay, line 985: This function is an exact duplicate of the one in SCAN and should be imported from a shared module.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+        'fu_care',
+        'fu_hospital_ed',
+        'fu_work_school',
+        'fu_activities',
+        'fu_which_activites',
+        'fu_test_result',
+        'fu_behaviors_no',
+        'fu_behaviors_inconclusive',
+        'fu_behaviors',
+        'fu_1_symptoms',
```
kairstenfay, line 1035: This function (I can't highlight the entire thing in GitHub UI) is an exact duplicate of the one in SCAN and should be imported from a shared module.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+
+    return questionnaire_response(record, question_categories, patient_reference, encounter_reference, True)
+
+def create_computed_questionnaire_response(record: dict, patient_reference: dict,
+                                            encounter_reference: dict,
+                                            birthdate: datetime, encounter_date: datetime) -> Optional[dict]:
+    """
+    Returns a FHIR Questionnaire Response resource entry for a
+    computed questionnaire.
+    """
```
kairstenfay, line 1111: it's not clear to me why this QuestionnaireResponse needs to created separately from other QuestionnaireResources, and furthermore, why this field needs to be calculated at all. Age should exist already as a calculated field .

kairstenfay (changes requested):
There are a lot of checks in this ETL for what type of event an instance is -- either enrollment or not (an "encounter"). I almost feel like we should just completely split the two workflows by checking only in one place for what type of event a record is instead of checking in 8 places.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: I agree with Jover here that we should not be representing the same exact data in two separate ways within the same FHIR Bundle. Rather, we should continue our practice of choosing the best available FHIR Resource and representing the data that way.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Here's what the group discussion in Slack came up with:
Do not ingest building names: Greek house, dorm, UW apartment name. Those are PII.
For creating locations:
1. Create the location for home address provided.
2. Look up a geocode for the Greek house, dorm, or UW apartment name. If we get one and if it's not the same as their home address, add it as a second location.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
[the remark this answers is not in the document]
jccraft-uw: I'm no longer ingesting the symptom questions as questionnaire items.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I'm no longer ingesting the symptom questions or core_sex as questionnaire items.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I'm no longer ingesting the symptom questions as questionnaire items.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I'm no longer ingesting the symptom questions as questionnaire items.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I'm no longer ingesting the symptom questions as questionnaire items.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I'm no longer ingesting the symptom questions as questionnaire items.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+        redcap_det.command_for_project(
+            name = project.command_name,
+            redcap_url = REDCAP_URL,
+            project_id = project.id,
+            raw_coded_values = True,
+            revision = REVISION,
+            help = help_message,
+            include_incomplete = True)(function)
+
+    return function
```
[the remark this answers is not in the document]
jccraft-uw: I did this intentionally because I suspect that we'll end up having more than one REDCap project for the UW Reopening. We may never have more than one project, I get that.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+    """
+    Create a Location reference for site of the sample collection encounter based
+    on how the sample was collected.
+    """
+    if collection_method == CollectionMethod.KIOSK:
+        record_location = record.get('location_type')
+        if record_location:
+            site = site_map(record_location)
+    else:
+        site = 'uwReopeningSwabNSend'
```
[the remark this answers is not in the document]
jccraft-uw: Ha ha. I really was trying to be consistent. I was trying to be consistent with the casing on "swabNSend" used in other ETL jobs. But that one wasn't consistent with the UW ones! I just changed this one to "UWReopeningSwabNSend".

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+    """
+    pass
+
+
+class UnknownRedcapRecordLocation(ValueError):
+    """
+    Raised by :function: `site_map` if a provided *redcap_location* is not
+    among a set of expected values.
+    """
+    pass
```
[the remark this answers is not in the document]
jccraft-uw: I noted this in the PR description and created a Slack conversation with the team about this the other day. A large part of the ETL code is copied from the SCAN ETL. Not everything is exactly the same, but a whole lot of it is. The right thing to do would be to move common code to a shared module, as you said, and refactor both ETLs. I couldn't do that within the timeframe. That would be good to refactor as a future project.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+        return {
+            "location": create_reference(
+                reference_type = "Location",
+                reference = location["fullUrl"]
+            )
+        }
+
+    def non_tract_locations(resource: dict):
+        return bool(resource) \
+            and resource['resource']['identifier'][0]['system'] != f"{INTERNAL_SYSTEM}/location/tract"
```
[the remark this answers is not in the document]
jccraft-uw: I noted this in the PR description and created a Slack conversation with the team about this the other day. A large part of the ETL code is copied from the SCAN ETL. Not everything is exactly the same, but a whole lot of it is. The right thing to do would be to move common code to a shared module, as you said, and refactor both ETLs. I couldn't do that within the timeframe. That would be good to refactor as a future project.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: The parameters needed for this bit of create_computed_questionnaire_response are:
birthdate: datetime
encounter_date: datetime
The function needs the two dates so it can compute the participant's age at the time of the encounter.

The logic that you pointed out is needed so the caller can pass in the encounter_date.

create_computed_questionnaire_response does take a encounter_reference: dict parameter , but I did not want to have it create the encounter date itself.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: OK. I switched to current_instance_entries and persisted_resource_entries.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: This questionnaire is the internal questionnaire that is needed for the priority queue/quota. We need it for every encounter.
I'm treating it differently from the daily survey because it's used for differently purposes. The two questionnaires are conceptually not the same.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I've separated the questionnaires conceptually.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Yes. The two ETLs should get refactored to use common code.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Yes. Future work is to refactor the two ETLs.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I separated the questionnaires conceptually.
Age in REDCap (core_age_years) is the participant's age when s/he enrolled. That may not be the age at the time of the encounter.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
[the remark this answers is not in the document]
kairstenfay: Thanks. Relatedly, the `swabNSend` name has been bothering me for some time, even though I'm pretty sure I was the one who came up with that name.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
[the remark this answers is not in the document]
kairstenfay: I understand. we're under serious time constraints and the REDCap project is infuriatingly not even in production mode yet, so it keeps changing. Would you please create a Trello card that refers to this PR so we can loop back soon to reduce the duplication across ETLs?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: Thanks. Could you add a description to the doc string of this function to elucidate the motivation for it?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: Thanks, could you please add more information to this function's docstring to elucidate its purpose?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
joverlee521: FYI, on the [FHIR ETL side ](https://github.com/seattleflu/id3c/blob/master/lib/id3c/cli/command/etl/fhir.py#L452) all of the QuestionnaireResponse Resources get combined into one.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+
+    assert redcap_record_instances is not None and len(redcap_record_instances) > 0, \
+        "The redcap_record_instances list was not populated."
+
+    record_id = redcap_record_instances[0]['record_id']
+    project_id = redcap_record_instances[0].project.id
+
+    enrollments = [record for record in redcap_record_instances if record["redcap_event_name"] == ENROLLMENT_EVENT_NAME]
+    assert len(enrollments) == 1, \
+        f"Record had {len(enrollments)} enrollments."
```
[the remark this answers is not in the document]
jccraft-uw: I don't know if it's possible in REDCap as a whole, but for this Husky Coronavirus Testing project we will not. The encounter event happens one time and then the encounter events happen daily.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I created Trello card https://trello.com/c/RWeUF0DA/814-in-uw-reopening-etl-add-a-location-for-dorm-greek-house-or-uw-apartment for this. This won't be in for go-live but could be added later.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1210 @@
...
+
+        event_type = None
+        collection_method = None
+
+        if redcap_record_instance["redcap_event_name"] == ENROLLMENT_EVENT_NAME:
+            event_type = EventType.ENROLLMENT
+        elif redcap_record_instance["redcap_event_name"] == ENCOUNTER_EVENT_NAME:
+            event_type = EventType.ENCOUNTER
+        else:
+            LOG.error(f"The record has an unexpected event name: {redcap_record_instance['redcap_event_name']}")
```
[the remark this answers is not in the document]
jccraft-uw: Yes indeed! Thanks for the catch. Fixed.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Good point. This is a bizarre instrument. I added this to the doc string:
This instrument is used
    to communicate with the Priority Queue application. Note that for this
    questionnaire we do want to return answers for all questions even if they are
    not answered.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I've added this to the docstring:
This "computed questionnaire" produces
    answers that do not appear directly in an actual REDCap survey.
    For example, a computed question captures the participant's age
    on the date of the encounter.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+        'core_latinx',
+        'mobility',
+        'vaccine_hx',
+        'hall_health',
+        'prior_test_base',
+        'travel_countries_phs_base',
+        'travel_states_phs_base',
+        'tier_1',
+        'tier_2',
+        'tier_3',
```
kairstenfay, line 707: Can a participant be part of multiple tiers? I think it's better to treat this as a combined question and create a questionnaireresponse item for Tier with the answer an integer equal to 1, 2, or 3.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+        'prior_test_type_base',
+        'prior_test_result_base',
+        'contact_base',
+        'wash_hands_base',
+        'clean_surfaces_base',
+        'hide_cough_base',
+        'mask_base',
+        'distance_base',
+        'bmi',
+        'novax_reason',
```
kairstenfay, line 668: This is being ingested as "8" (e.g. test record ID 48) instead of the actual value.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+        'covid_vaccine',
```
kairstenfay, line 669: This would be better ingested as boolean with "Unknown" being dropped/treated as null

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+        'vaccine_where',
+        'on_campus_freq',
```
kairstenfay, line 671: Can this be ingested as an integer instead of a string?
e.g. 0, 1, 2

kairstenfay (changes requested):
Looking good, Chris! I just ran through the entire pipeline (REDCap DET ETL + FHIR ETL). here are my observations after spot checking one record.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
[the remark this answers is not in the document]
jccraft-uw: @kairstenfay @jover On the daily attestation questionnaire the question is "Which of the following symptoms have you experienced in the last 24 hours? Select all that apply.". There is no question for onset date for these symptoms.
Originally, I was using calculating the onset date as the attestation_date minus 24 hours. Jover questioned that, so I removed it and left no onset date.
Kairsten just pointed out that having no onset date violates FHIR. 
I see two options: Don't use these as symptoms or calculate the onset date.

As Jover points out, there is another very similar question on the same questionnaire:
"Which new or worsening symptoms have you experienced in the last 7 days? Pick any that answer the question."
It has a corresponding onset date field. I am already using that one for symptoms.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+        'text_or_email',
+        'text_or_email_attestation',
+        'campus_location',
+        'affiliation',
+        'student_level',
+        'sea_employee_type',
+        'core_house_members',
+        'core_education',
+        'core_income',
+        'vaccine_method',
```
kairstenfay, line 650: These responses are being ingested as an integer, e.g. "2" in test record ID 61 instead of the expected value

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+        'hospital_ed_base',
+        'prior_test_positive_base',
+        'prior_test_type_base',
+        'prior_test_result_base',
+        'contact_base',
+        'wash_hands_base',
+        'clean_surfaces_base',
+        'hide_cough_base',
+        'mask_base',
+        'distance_base',
```
kairstenfay, line 666: These questions, when the answer is "na", should not be ingested into FHIR Bundles.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+        'bmi',
```
kairstenfay, line 667: This is a float value and should not be ingested as a string.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+        'novax_reason',
+        'covid_vaccine',
+        'vaccine_where',
```
kairstenfay, line 670: This is being ingested as "3" (e.g. record 61) instead of the expected string value

kairstenfay (commented):
I just finished spot checking another record. Here are some more observations.

In my opinion, the vaccine questions are low priority and more trouble than they're worth. They could just be dropped for now.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
joverlee521: Ah, these sound like they are coding that needs to be changed in REDCap.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: The choices in REDCap for novax_reason are integers:
8, I plan to get the flu vaccine
1, Do not have time to get vaccinated
2, Not required for work or school
3, Not recommended by a doctor or healthcare worker
4, Not covered by health insurance
5, Not offered at a convenient location
6, I am not worried about getting sick with the flu
7, Concerns about vaccine safety or effectiveness
000, None of the above
999, Prefer not to say

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: The options in REDCap for covid_vaccine are:
Yes, Yes
No, No
Undecided, Undecided

"Undecided" seems more meaningful than a general "Unknown" would be. 
I see value in capturing the undecided. What do you think?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: In REDCap vaccine_method has these options:
1, Injection (flu shot)
2, Nasal spray (flu mist)
Thus, I'm ingesting the value as an integer.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: In REDCap the options for vaccine_where are:
1, Medical clinic or hospital
2, Pharmacy or drugstore
3, Workplace
4, School
5, Other

I changed the ETL so that this gets brought in as an integer. (It was a string.)

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Ah, but I had it ingested as a string. I made it an integer now.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+        'on_campus_freq',
+        'covid_novax_reason',
+        'countries_visited_base',
+        'states_visited_base',
+        'alerts_off',
+        'pronouns'
```
kschwabefry, line 676: May want to include the pronouns_other field as well

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: No, we want it to be ingested as a string. This is a REDCap side thing that needs to be fixed.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: No, we want it to be ingested as a string. This is a REDCap side thing that needs to be fixed.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: Good question. Now I'm no longer sure. This is probably a question for the folks in #informatics.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Our rule is not to ingest any free text fields because they could potentially contain PII. In the transfer project I'm grabbing the pronouns_other straight from REDCap.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1206 @@
...
+                {
+                    "system": f"{INTERNAL_SYSTEM}/symptom",
+                    "code": mapped_symptom_name
+                }
+            ]
+        },
+        "subject": patient_reference
+    }
+
+    condition["onsetDateTime"] = onset_date
```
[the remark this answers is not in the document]
kairstenfay: We can either not ingest symptoms without an `onset_date` or make onsetDateTime conditional by writing:
```
if onset_date:
        condition["onsetDateTime"] = onset_date
```

If we only ingest with `onset_date` is available, then we should just move `onsetDateTime` up into the dictionary definition above.

kairstenfay:
To summarize from our recent video call:
Most of the fields I've commented on are not required for launch. They represent data that are "nice to have" that can be ingested at a future date. If ironing out the QuestionnaireResponse item details is too pesky right now (which I suspect it might be), then I support a new commit dropping them for now.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kschwabefry: Oh yes, that's right.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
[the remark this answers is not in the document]
jccraft-uw: Thanks for talking about this today. I'm keeping both sets of symptoms. For the daily_covid ones (past 24 hours) I'm simply not setting the onset_date if there isn't one. @kairstenfay thanks for your note. I'm sorry I was grocking what you meant earlier.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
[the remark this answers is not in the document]
jccraft-uw: Sorry I didn't grock this earlier. Thanks for spelling out the line of code to not set onset_date if we don't have one. I added that.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Great suggestion. I've made that change to have a single 'tier' item now.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: This variable was in the list whose values Misja fixed from integer to string in REDCap today. I'm ingesting this as a string.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I kept this as a string. In a view, we can always convert to boolean and drop 'Undecided' if our consumers want it that way.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: This variable was in the list whose values Misja fixed from integer to string in REDCap today. I'm ingesting this as a string.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: This variable was in the list whose values Misja fixed from integer to string in REDCap today. I'm ingesting this as a string.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Thanks for raising this point. When I wrote this I thought that the types used in SCAN were the only ones we had available.
'valueCoding', 'valueBoolean, 'valueInteger' 'valueString' and 'valueDate'
Given those I went with valueString to capture the float.
@joverlee521 (thank you) today told me that these are actually FHIR concepts, not things we came up with ourselves.
So, for this one I'm using FHIR's valueDecimal. (They don't have a valueFloat). For the value itself, I'm setting it as a float.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: This variable was in the list whose values Misja fixed from integer to string in REDCap today. I'm ingesting this as a string.

jccraft-uw:
> To summarize from our recent video call:
> Most of the fields I've commented on are not required for launch. They represent data that are "nice to have" that can be ingested at a future date. If ironing out the QuestionnaireResponse item details is too pesky right now (which I suspect it might be), then I support a new commit dropping them for now.

Thanks. Misja made the variable option changes in REDCap, and I made the corresponding code changes in here.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1237 @@
...
+
+        initial_encounter_entry, initial_encounter_reference = create_encounter(
+            redcap_record_instance, patient_reference, site_reference,
+            location_resource_entries, event_type, collection_method)
+
+        # Skip the entire record if we can't create the enrollment encounter.
+        # Otherwise, just skip the record instance.
+        if not initial_encounter_entry:
+            if event_type == EventType.ENROLLMENT:
+                return None
```
joverlee521, line 166: I think we should also log a warning here if we have to skip an enrollment

jccraft-uw: Added.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1238 @@
...
+    items: List[dict] = []
+    for category in question_categories:
+        category_items = list(map(build_questionnaire_items, question_categories[category]))
+        for item in category_items:
+            if item:
+                items.append(item)
+
+    vaccine_item = vaccine(record)
+    if vaccine_item:
+        items.append(vaccine_item)
```
joverlee521, line 889: This is adding the vaccine question item to both the computed questionnaire and the enrollment questionnaire. I think we just want it the enrollment one?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: That's not quite what I meant here. I was wondering if we could convert the string values to integers. But it was just a thought and completely optional / I'm not even sure if that's what we want to do .

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1238 @@
...
+
+def create_testing_determination_internal_questionnaire_response(record: dict, patient_reference: dict,
+                                            encounter_reference: dict) -> Optional[dict]:
+    """
+    Returns a FHIR Questionnaire Response resource entry for the
+    testing_determination_internal instrument. This instrument is used
+    to communicate with the Priority Queue application. Note that for this
+    questionnaire we do want to return answers for all questions even if they are
+    not answered.
+    """
```
joverlee521, line 1121: Can you explain why we want to ingest these fields even if they are not answered? Like we discussed earlier today, empty string values violates FHIR guidelines.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: That was a really great embedded catch. We do only want the vaccine item for the enrollment question. I've added a parameter and set it accordingly in the latest commit. I tested this locally and confirmed that I get vaccine only for the enrollment questionnaire. The `questionnaire_response` function's parameters are not very elegant, but they work.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: These fields will be used by the quota/priority queue application. I thought providing them always would make it easier on the quota query side. Unfortunately, the only way to grab data from REDCap is into a FHIR bundle. If we don't want to violate its guidelines and the quote app can do just fine without the fields when they're empty I can remove this functionality.

joverlee521: When we don't ingest them because they are empty, they should just show up as null values in the SQL query. I think that matches our expectations.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: The options in REDCap are now:
one_or_less, One day a week or less
two_or_more, Two days a week or more
not_on_campus, I do not come to campus
The first two would not convert cleanly into an integer. We could do ranges
0 to 1
2 to 7 (or 5 if we consider only workdays)
0
That doesn't seem super helpful.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: OK. I'll do the surgery to remove the related code. Look for a new commit coming in a bit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1234 @@
...
+    # Do not include `core_age_years` because we calculate the age ourselves in the computed questionnaire.
+    integer_questions = [
+        'weight',
+        'height_total',
+        'tier'
+
+    ]
+
+    string_questions = [
+        'study_area',
```
kairstenfay, line 661: I believe this should be coded as a boolean True/False instead of a string yes/no

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I've removed the code that kept items that had an empty string value.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1234 @@
...
+        'english_speaking',
+        'athlete',
+        'uw_medicine_yesno',
+        'inperson_classes',
+        'uw_job',
+        'uw_greek_member',
+        'live_other_uw',
+        'uw_apt_yesno',
+        'core_pregnant',
+        'core_latinx',
```
kairstenfay, line 718: Thoughts for the future:
In the past we've ingested this as "ethnicity", and I can't currently recall whether we stored that in a FHIR resource as a primary Resource attribute or a QuestionnaireResponse Item

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -0,0 +1,1234 @@
...
+        'prior_test_positive',
+        'prior_test_type',
+        'prior_test_result',
+        'contact',
+        'wash_hands',
+        'clean_surfaces',
+        'hide_cough',
+        'mask',
+        'distance',
+        'attend_event',
```
kairstenfay, line 1185: This should be a boolean, like the question on pregnancy

kairstenfay (changes requested):
Almost there! We have a couple of questions coded as string that need to be converted to boolean. 

Also, are we not ingesting `illness_kiosk` or `illness_swabsend`? Perhaps there's some context I've missed here.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: attend_event choices are:
yes, Yes
no, No
dont_say, Prefer not to say
That's why I made it a string.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: Thanks!

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: I lost this argument on a call this morning where we had a majority opinion that "NA" is distinct from null and therefore should be ingested. Marking this as resolved.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: No you're right. Forget I mentioned this.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: These are the same values we see for `core_pregnant`, but we're ingesting `core_pregnant` as boolean. Therefore we should be consistent.

jccraft-uw:
> Almost there! We have a couple of questions coded as string that need to be converted to boolean.
> 
> Also, are we not ingesting `illness_kiosk` or `illness_swabsend`? Perhaps there's some context I've missed here.

I'm not currently ingesting any questions from the Husky Test Kit Registration or Kiosk Registration instruments. I'm getting the symptoms from both and not getting any of the actual questionnaire items from either. Is having the symptoms enough?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: > Is having the symptoms enough?

Probably.

kairstenfay:
> Is having the symptoms enough?

Probably.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Changed in latest commit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: That is a good argument! Changed in latest commit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: In redcap_det_scan it's a boolean withe the name 'ethnicity;

kairstenfay (approved):
:smile:

joverlee521 (approved):
🎉  Thank you for all you're work @jccraft-uw

PR #144, opened 2020-09-18
Quotas

tsibley:
_WIP, not ready for review at all. Making a PR for visibility only._


tsibley:
A bunch of description here: https://github.com/seattleflu/id3c-customizations/pull/144/files#diff-275ef3202ccd63065aff184bf5c7f3b4R2-R72.

May be worth doing the views described in detail there first, then hooking it up to the quota framework implemented here.

lib/seattleflu/id3c/cli/command/triage_uw_test_queue.py
```diff
@@ -0,0 +1,181 @@
...
+    offers = db.fetch_rows("""
+        select
+            redcap_record_id,
+            redcap_event_name,
+            redcap_repeat_instance_id,
+            reason,
+            priority
+        from
+            shipping.uw_priority_queue_v1
+        limit
```
jccraft-uw, line 142: order by priority ?

tsibley: Yes, that's important! But the intent was for `shipping.uw_priority_queue_v1` to already be ordered by priority, and that's how Jover implemented it in #151. So I think the above is good as-is.

jccraft-uw: Since priority is key it would be safer to specify the order by here even if the view does it.

tsibley: The idea was that this command would treat the queue as a black box. It just pulls things off the top N at a time, and whatever the queue says is on top is on top. If we don't treat the queue as a black box, then when we update the queue later to break ties (which I think we need to), then we're going to have to update this command too.

But if you think the balance favors an explicit ordering here, then fair enough. I checked and Pg's planner does appear smart enough to realize the view is already sorted by the requested column and not resort.

tsibley: ([Further comments](https://github.com/seattleflu/id3c-customizations/pull/151#discussion_r504165879) on the tie breaking we'd have to duplicate here if we think an explicit ordering here is advised.)

jccraft-uw: > The idea was that this command would treat the queue as a black box. It just pulls things off the top N at a time, and whatever the queue says is on top is on top. If we don't treat the queue as a black box, then when we update the queue later to break ties (which I think we need to), then we're going to have to update this command too.
> 
> But if you think the balance favors an explicit ordering here, then fair enough. I checked and Pg's planner does appear smart enough to realize the view is already sorted by the requested column and not resort.

I'll drop this. It's not a big deal as long as the data gets to the code sorted.

PR #145, opened 2020-09-22
Add SQL changes to support UW reopening project

jccraft-uw:
This PR adds some sites and adds 3 shipping views for the UW reopening.

In particular, please check that I've set the right permissions on the 3 views.


kairstenfay:
Don't forget to, in the revert script ,delete the sites you added in the deploy script.

schema/sqitch.plan
```diff
@@ -229,3 +229,4 @@ shipping/views [shipping/views@2020-09-04] 2020-09-08T18:58:09Z Kairsten Fay <kf
 
 shipping/views [shipping/views@2020-09-08] 2020-09-14T21:35:26Z Kairsten Fay <kfay@fredhutch.org> # Add UW Reopening samples to reportable conditions view
 @2020-09-14 2020-09-14T21:40:13Z Kairsten Fay <kfay@fredhutch.org> # Schema as of 14 September 2020
+warehouse/site/data [warehouse/site/data@2020-09-14] 2020-09-22T01:53:23Z Chris Craft,,, <jccraft@chriscraft-lapdance> # Add UW reopening sites
```
kairstenfay, line 232: I'm not sure that your sqitch settings are configured properly. It's ok if you don't want to share your email properly, but maybe you can use your GitHub auto-reply email instead? Also, the language you use here to refer to your laptop name makes me feel uncomfortable. I do not want to commit this to our codebase. Could you please use a more professional personal identifier?

kairstenfay (changes requested):

schema/sqitch.plan
jccraft-uw: I didn't know I had to configure sqitch! I manually updated the text in sqitch.plan with my work email address.

jccraft-uw: I'm now configured.

schema/deploy/shipping/views.sql
```diff
@@ -2576,4 +2580,309 @@ grant select
...
+;
+
+comment on view shipping.uw_reopening_ehs_reporting_v1 is
+  'For the UW reopening project, a view that combines encounter, enrollment, and hCoV-19 test result data';
+
+/* The shipping.uw_reopening_ehs_reporting_v1 view needs hCoV-19 visibility, so
+ * remains owned by postgres, but it should only be accessible by those with
+ * hcov19-visibility.  Revoke existing grants to every other role.  */
+ revoke all on shipping.uw_reopening_ehs_reporting_v1 from reporter;
+
```
joverlee521, line 2887: Need to grant select on view to `"hcov19-visibility"` role

joverlee521 (changes requested):
You will need to create a new role (`ehs-results-exporter`?) that would be granted select to the ehs reporting view. The backoffice user can then be granted this role to export the results from ID3C.

schema/deploy/warehouse/site/data.sql
```diff
@@ -52,16 +52,21 @@ insert into warehouse.site(identifier, details)
...
+        ('UWBothell',                               '{"category": "community",  "type": "collegeCampus"}'),
         ('UWClub',                                  '{"category": "community",  "type": "collegeCampus"}'),
         ('UWDaycare',                               '{"category": "community",  "type": "childcare"}'),
         ('UWGreek',                                 '{"category": "community",  "type": "collegeCampus",
             "swab_site": "uw_greek"}'),
         ('UWHallHealth',                            '{"category": "clinic",     "type": "clinic",
             "swab_site": "hall_health"}'),
+        ('UWOdegaardLibrary',                       '{"category": "community",  "type": "collegeCampus",
+            "swab_site": "odegaard"}'),
         ('UWSeaMar',                                '{"category": "clinic",     "type": "clinic",
```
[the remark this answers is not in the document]
jccraft-uw: Added!

schema/deploy/warehouse/site/data.sql
```diff
@@ -52,16 +52,21 @@ insert into warehouse.site(identifier, details)
...
             "swab_site": "swab_and_send"}'),
+        ('UWBothell',                               '{"category": "community",  "type": "collegeCampus"}'),
         ('UWClub',                                  '{"category": "community",  "type": "collegeCampus"}'),
         ('UWDaycare',                               '{"category": "community",  "type": "childcare"}'),
         ('UWGreek',                                 '{"category": "community",  "type": "collegeCampus",
             "swab_site": "uw_greek"}'),
         ('UWHallHealth',                            '{"category": "clinic",     "type": "clinic",
             "swab_site": "hall_health"}'),
+        ('UWOdegaardLibrary',                       '{"category": "community",  "type": "collegeCampus",
+            "swab_site": "odegaard"}'),
```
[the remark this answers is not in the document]
jccraft-uw: Removed!

schema/deploy/shipping/views.sql
```diff
@@ -49,6 +49,10 @@ drop materialized view if exists shipping.fhir_questionnaire_responses_v1;
 
 drop view if exists shipping.sample_with_best_available_encounter_data_v1;
 
+drop view if exists shipping.uw_reopening_ehs_reporting_v1;
+drop view if exists shipping.uw_reopening_encounters_v1;
+drop view if exists shipping.uw_reopening_enrollment_fhir_encounter_details_v1;
```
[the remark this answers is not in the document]
jccraft-uw: Moved. Boy oh boy. Why not drop view if exists blah cascade ? Then we wouldn't have to worry about the exact order.

schema/revert/shipping/views.sql
```diff
@@ -49,6 +49,10 @@ drop materialized view if exists shipping.fhir_questionnaire_responses_v1;
 
 drop view if exists shipping.sample_with_best_available_encounter_data_v1;
 
+drop view if exists shipping.uw_reopening_ehs_reporting_v1;
+drop view if exists shipping.uw_reopening_encounters_v1;
+drop view if exists shipping.uw_reopening_enrollment_fhir_encounter_details_v1;
```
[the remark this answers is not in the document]
jccraft-uw: Moved.

schema/deploy/shipping/views.sql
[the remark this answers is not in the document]
joverlee521: Hmm that's a good question. I personally think it's better to be explicit about the drop statements since they are destructive, but we can definitely have a discussion about this.

schema/deploy/shipping/views.sql
```diff
@@ -2576,4 +2580,313 @@ grant select
...
+grant select
+  on shipping.uw_reopening_encounters_v1
+  to "incidence-modeler";
+
+create or replace view shipping.uw_reopening_ehs_reporting_v1 as
+(
+  select ids.barcode
+  , results.collect_ts as sample_collection_date_time
+  , results.status_code as test_result
+  , results.result_ts as test_result_date_time
```
kschwabefry, line 2843: Non-blocker, but I'd have a preference for naming these sample_collection_date and test_result_date as the values  are dates and not datetimes.

kschwabefry (approved):
Looks good on my end. Once permissions are set and any other feedback from Kairsten/Jover is addressed, this looks good to go.

jccraft-uw:
> You will need to create a new role (`ehs-results-exporter`?) that would be granted select to the ehs reporting view. The backoffice user can then be granted this role to export the results from ID3C.

The role is in now.

schema/deploy/shipping/views.sql
jccraft-uw: Changed.

schema/deploy/roles/ehs-results-exporter/grants.sql
```diff
@@ -0,0 +1,19 @@
...
+
+grant connect on database :"DBNAME" to "ehs-results-exporter";
+
+grant usage
+    on schema shipping
+    to "ehs-results-exporter";
+
+grant select
+    on shipping.uw_reopening_ehs_reporting_v1
+    to "ehs-results-exporter";
```
joverlee521, line 17: View specific grants need to live in the views.sql. Sorry this isn't very clear! 🙏 

We made the recent switch because we are dropping all views on every rework, the view specific grants will not persist once the view has been dropped. So these view grants need to be declared under the view create (similar to how we are granting "incidence-modeler" on all the views).

schema/verify/roles/ehs-results-exporter/grants.sql
```diff
@@ -0,0 +1,12 @@
+-- Verify seattleflu/id3c-customizations:roles/ehs-results-exporter/grants on pg
+
+begin;
+
+select 1/pg_catalog.has_database_privilege('ehs-results-exporter', :'DBNAME', 'connect')::int;
+select 1/pg_catalog.has_schema_privilege('ehs-results-exporter', 'shipping', 'usage')::int;
+select 1/pg_catalog.has_table_privilege('ehs-results-exporter', 'shipping.uw_reopening_ehs_reporting_v1', 'select')::int;
+
+select 1/(not pg_catalog.has_table_privilege('ehs-results-exporter', 'shipping.uw_reopening_ehs_reporting_v1', 'insert,update,delete'))::int;
```
joverlee521, line 9: this can be removed since the view specific grant needs to be in the views rework.

schema/sqitch.plan
```diff
@@ -229,3 +229,11 @@ shipping/views [shipping/views@2020-09-04] 2020-09-08T18:58:09Z Kairsten Fay <kf
 
 shipping/views [shipping/views@2020-09-08] 2020-09-14T21:35:26Z Kairsten Fay <kfay@fredhutch.org> # Add UW Reopening samples to reportable conditions view
 @2020-09-14 2020-09-14T21:40:13Z Kairsten Fay <kfay@fredhutch.org> # Schema as of 14 September 2020
+
+warehouse/site/data [warehouse/site/data@2020-09-14] 2020-09-22T01:53:23Z Chris Craft <jccraft@uw.edu> # Add UW reopening sites
+shipping/views [shipping/views@2020-09-14] 2020-09-22T16:07:49Z Chris Craft <jccraft@uw.edu> # Add shipping views for UW reopening project
+@2020-09-22 2020-09-22T22:46:29Z Chris Craft <jccraft@uw.edu> # Schema as of 22 September 2020
+
+roles/ehs-results-exporter/create 2020-09-23T17:45:08Z Chris Craft <jccraft@uw.edu> # Create ehs-results-exporter role
+roles/ehs-results-exporter/grants 2020-09-23T17:45:53Z Chris Craft <jccraft@uw.edu> # Add grants to new ehs-results-exporter role
```
joverlee521, line 238: The role creation/grants needs to be before the view rework so the view grants can be declared in the rework.

schema/deploy/roles/ehs-results-exporter/grants.sql
jccraft-uw: Changed. This makes sense but does make things complicated. If the roles relate only to shipping views we could manage the roles in the same script as the views. That's how I did it at Xealth. Roles were at the top and the views were below them.

schema/verify/roles/ehs-results-exporter/grants.sql
jccraft-uw: I kept the non-view checks.

schema/sqitch.plan
jccraft-uw: I switched sqitch.

kairstenfay (changes requested):
This is looking really good, Chris.

Sorry I didn't catch this earlier, but we've changed the way we grant roles access to views in the past few months. 

See [Thomas's comment here](https://github.com/seattleflu/id3c-customizations/blob/5623717fe7e8039c5f42babeb4305681d3889fdf/schema/deploy/shipping/views.sql#L1126) for an explanation of what we used to do, why, and why we're switching now. 

You don't need a sqitch add for grants for the new EH&S role. Instead, we'll just grant them scoped access to the specific view they need. See the way we grant privileges to `hcov19-visibility` in `views.sql` as an example.

kairstenfay:
> This is looking really good, Chris.
> 
> Sorry I didn't catch this earlier, but we've changed the way we grant roles access to views in the past few months.
> 
> See [Thomas's comment here](https://github.com/seattleflu/id3c-customizations/blob/5623717fe7e8039c5f42babeb4305681d3889fdf/schema/deploy/shipping/views.sql#L1126) for an explanation of what we used to do, why, and why we're switching now.
> 
> You don't need a sqitch add for grants for the new EH&S role. Instead, we'll just grant them scoped access to the specific view they need. See the way we grant privileges to `hcov19-visibility` in `views.sql` as an example.

Looking again at your code, I see you do add grants in `views.sql`. I overlooked that in my initial review. I would suggest just mimicking what we've done for hcov19-visibility (of course, with differently scoped permissions) for this new role.

jccraft-uw:
> > This is looking really good, Chris.
> > Sorry I didn't catch this earlier, but we've changed the way we grant roles access to views in the past few months.
> > See [Thomas's comment here](https://github.com/seattleflu/id3c-customizations/blob/5623717fe7e8039c5f42babeb4305681d3889fdf/schema/deploy/shipping/views.sql#L1126) for an explanation of what we used to do, why, and why we're switching now.
> > You don't need a sqitch add for grants for the new EH&S role. Instead, we'll just grant them scoped access to the specific view they need. See the way we grant privileges to `hcov19-visibility` in `views.sql` as an example.
> 
> Looking again at your code, I see you do add grants in `views.sql`. I overlooked that in my initial review. I would suggest just mimicking what we've done for hcov19-visibility (of course, with differently scoped permissions) for this new role.

@joverlee521 and I went through a couple of iterations on this. As you said, granting select on the view is right under the view in views.sql. 
I like how I ended up with this. The role is created and is given very basic permissions in the grant. That lets is connect to the DB and use the shipping schema. The verify checks those things.

OK?

kairstenfay (approved):
Jover wisely pointed out that the grant handling here for ehs-results-exporter is the same as was previously done for the scan-dashboard-exporter. I overlooked that.

joverlee521 (approved):

PR #146, opened 2020-09-27
uw-reopening ETL: Check that encounter data exists

jccraft-uw:
Adds a check to the start of the record instance loop to make sure that
we have the data we need to create an ENCOUNTER encounter.
Otherwise, skip the record instance with a log entry at the DEBUG
level.

Switches instrument from test_fulfillment_form to test_order_survey
when looking for swab and send encounter to make the check
consistent with what is pulled for the encounter_date.

This commit also condenses some of the event type logic.

The 'Testing Determination - Internal' instrument does not capture a
date, so calculate encounter_date based on the repeat instance number
and the date that the first repeat instance represents for the project.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -33,6 +33,7 @@ class HuskyProject():
     api_token_env_var: str
     command_name: str
 
+
     def __init__(self, project_id: int, lang: str, api_token_env_var: str, command_name: str) -> None:
         self.id = project_id
         self.lang = lang
@@ -55,7 +56,7 @@ class EventType(Enum):
     ENROLLMENT = 'enrollment'
     ENCOUNTER = 'encounter'
 
-REVISION = 0
+REVISION = 1
 
 REDCAP_URL = 'https://redcap.iths.org/'
 INTERNAL_SYSTEM = "https://seattleflu.org"
@@ -142,15 +143,22 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
             event_type = EventType.ENROLLMENT
         elif redcap_record_instance["redcap_event_name"] == ENCOUNTER_EVENT_NAME:
             event_type = EventType.ENCOUNTER
+            if is_complete('kiosk_registration_4c7f', redcap_record_instance):
+                collection_method = CollectionMethod.KIOSK
+            elif is_complete('test_order_survey', redcap_record_instance):
+                collection_method = CollectionMethod.SWAB_AND_SEND
         else:
             LOG.error(f"The record instance has an unexpected event name: {redcap_record_instance['redcap_event_name']}")
             continue
 
-        if event_type == EventType.ENCOUNTER:
-            if is_complete('kiosk_registration_4c7f', redcap_record_instance):
-                collection_method = CollectionMethod.KIOSK
-            elif is_complete('test_fulfillment_form', redcap_record_instance):
-                collection_method = CollectionMethod.SWAB_AND_SEND
+        # Skip an ENCOUNTER instance if we don't have the data we need to
+        # create an encounter.
+        if event_type == EventType.ENCOUNTER \
+            and not is_complete('daily_attestation', redcap_record_instance) \
+                and not collection_method  \
+                and not redcap_record_instance['testing_date']: # from the 'Testing Determination - Internal' instrument
+                    LOG.debug("Skipping record instance with insufficient information to construct the initial encounter")
+                    continue
 
         # site_reference refers to where the sample was collected
         site_reference = create_site_reference(redcap_record_instance, collection_method, event_type)
@@ -166,7 +174,7 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
                 LOG.warning("Skipping record because we could not create the enrollment encounter")
                 return None
             else:
-                LOG.warning("Skipping record instance with insufficient information to construct a initial encounter")
+                LOG.warning("Skipping record instance with insufficient information to construct the initial encounter")
                 continue
 
         specimen_entry = None
@@ -191,6 +199,9 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
         enrollment_questionnaire_entry = None
         daily_questionnaire_entry = None
         testing_determination_internal_questionnaire_entry = None
+        follow_up_encounter_entry = None
+        follow_up_questionnaire_entry = None
+        follow_up_computed_questionnaire_entry = None
 
         computed_questionnaire_entry = create_computed_questionnaire_response(
             redcap_record_instance, patient_reference, initial_encounter_reference,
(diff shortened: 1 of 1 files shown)


jccraft-uw:
[Lower log level when skipping record w/ insufficient info to create an encounter - UW Reopening REDCap DET ETL](https://trello.com/c/ADBWy8jR/823-lower-log-level-when-skipping-record-w-insufficient-info-to-create-an-encounter-uw-reopening-redcap-det-etl)

kairstenfay (changes requested):
Thanks for these updates, Chris. I left a few comments.

Note that we'll need to bump the revision number if we want to reprocess all the records.

jccraft-uw:
> Thanks for these updates, Chris. I left a few comments.
> 
> Note that we'll need to bump the revision number if we want to reprocess all the records.

We won't need to re-process existing records. We skipped them before, just in a way that was too noisy.

kairstenfay:
> > Thanks for these updates, Chris. I left a few comments.
> > Note that we'll need to bump the revision number if we want to reprocess all the records.
> 
> We won't need to re-process existing records. We skipped them before, just in a way that was too noisy.

Perhaps I've misunderstood.  In your commit message, you say:

> Switches instrument from test_fulfillment_form to test_order_survey
when looking for swab and send encounter to make the check
consistent with what is pulled for the encounter_date.

Does this mean we're now pulling encounter date from a different instrument?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -166,7 +180,7 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
                 LOG.warning("Skipping record because we could not create the enrollment encounter")
                 return None
             else:
-                LOG.warning("Skipping record instance with insufficient information to construct a initial encounter")
+                LOG.warning("Skipping record instance with insufficient information to construct the initial encounter")
```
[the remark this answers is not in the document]
jccraft-uw: With the new checks in place this scenario really shouldn't happen. If it does I would like to be alerted to it, so I want to keep this a warning. It's similar to the line above, something unexpected was missing from the data.

jccraft-uw:
> > > Thanks for these updates, Chris. I left a few comments.
> > > Note that we'll need to bump the revision number if we want to reprocess all the records.
> > 
> > 
> > We won't need to re-process existing records. We skipped them before, just in a way that was too noisy.
> 
> Perhaps I've misunderstood. In your commit message, you say:
> 
> > Switches instrument from test_fulfillment_form to test_order_survey
> > when looking for swab and send encounter to make the check
> > consistent with what is pulled for the encounter_date.
> 
> Does this mean we're now pulling encounter date from a different instrument?

How we set the encounter_date hasn't changed for anything except now for an encounter instance when we have only the Testing Determination - Internal instrument and its surge is populated we would now pull a date.
Fair enough. I'll bump the version.

kairstenfay:
> How we set the encounter_date hasn't changed for anything except now for an encounter instance when we have only the Testing Determination - Internal instrument and its surge is populated we would now pull a date.
> Fair enough. I'll bump the version.

Ah. You have far more context here than I do. If this change doesn't affect any of the REDCap records, then no need to bump the revision number! Similarly, if it only affects a small subset of them, then we can just manually generate REDCap DETs for the few applicable ones and keep the current revision number the same.

jccraft-uw:
> > How we set the encounter_date hasn't changed for anything except now for an encounter instance when we have only the Testing Determination - Internal instrument and its surge is populated we would now pull a date.
> > Fair enough. I'll bump the version.
> 
> Ah. You have far more context here than I do. If this change doesn't affect any of the REDCap records, then no need to bump the revision number! Similarly, if it only affects a small subset of them, then we can just manually generate REDCap DETs for the few applicable ones and keep the current revision number the same.

There was mention of a surge last week, so to be safe I'll bump the revision.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -33,14 +33,19 @@ class HuskyProject():
...
+        repeat_instance_1_date: datetime) -> None:
         self.id = project_id
         self.lang = lang
         self.api_token_env_var = api_token_env_var
         self.command_name = command_name
+        self.repeat_instance_1_date = repeat_instance_1_date
 
 PROJECTS = [
-        HuskyProject(23854, "en", "REDCAP_API_TOKEN", "uw-reopening")
+        HuskyProject(23854, "en", "REDCAP_API_TOKEN", "uw-reopening", datetime(2020, 9, 24))
```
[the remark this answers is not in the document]
jccraft-uw: This is not global. It's for the specific project. (We have only one for now.) I added a comment there to make it clear what the value represents.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -500,6 +516,14 @@ def non_tract_locations(resource: dict):
...
                 '%Y-%m-%d %H:%M:%S').strftime('%Y-%m-%d')
 
+        # The 'Testing Determination - Internal' instrument does not capture a date,
+        # so calculate encounter_date based on the repeat instance number and the
+        # date that the first repeat instance represents for the project.
+        elif record.get('surge_selected'):
+            encounter_date = (husky_project.repeat_instance_1_date + \
+            relativedelta(days =+ (int(record.get('redcap_repeat_instance')) - 1))) \
+                .strftime('%Y-%m-%d')
+
```
tsibley, line 526: IME, datetime math used for purposes like this is sometimes prone to subtle edge cases. Instead of addressing the symptom, could we address the problem and change REDCap to capture a date? Or use one of the existing dates from the other instruments in the same repeated event instance?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -99,6 +106,8 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
 
     record_id = redcap_record_instances[0]['record_id']
     project_id = redcap_record_instances[0].project.id
+    husky_project = next((p for p in PROJECTS if p.id == project_id))
+
```
tsibley, line 110: This lookup suggests to me that `PROJECTS` would be better off as a dictionary.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -33,14 +33,21 @@ class HuskyProject():
     api_token_env_var: str
     command_name: str
 
-    def __init__(self, project_id: int, lang: str, api_token_env_var: str, command_name: str) -> None:
+    # The date represented by REDCap record repeat instance #1.
+    repeat_instance_1_date: datetime
```
tsibley, line 37: I'd expect to use a `date` instead of `datetime` object: https://docs.python.org/3.5/library/datetime.html#date-objects

Dates are much simpler than datetimes and would avoid lots of complex logic around timezones that we don't need or want in this case.

tsibley (commented):
Thanks for working to make this less noisy! I left a few comments on the approach here, but nothing immediately blocking.

One other thing that stood out to me is that `HuskyProject` could be a `typing.NamedTuple` ([like `ScanProject`](https://github.com/seattleflu/id3c-customizations/blob/6e9a1d7de13969362c6ac9dba214275b52d79644/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py#L28-L31)) and avoid some boilerplate.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: I will ask if they can add a date to the Testing Internal survey. The surge scenario could mean that the encounter event in REDCap gets created and the only thing populated is the name of a surge on the Testing Internal survey. In that case we wouldn't have another instrument to grab the date from.

tsibley: Ah, got it! Thanks for explaining. That makes sense. Hopefully a date can be added to REDCap.

jccraft-uw: Woo hoo! They're going to add the field. I'll hold this code change. I'll take out the code I added to handle the repeated instance date once that field gets in.

jccraft-uw: We got the date field, and I've updated the code to use that instead.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: We don't need this code anymore, so I'm removed it.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: We don't need this code anymore, so I'm removed it.

kairstenfay (approved):

kschwabefry (approved):

joverlee521 (approved):

tsibley (approved):

PR #147, opened 2020-09-30
etl redcap-det scan: Add PICAWA site

kairstenfay:
Add a new site for PICA-WA to the SCAN In-Person Enrollments project.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -248,6 +248,7 @@ def site_map(record_location: str) -> str:
     """
     location_site_map = {
         'greek': 'UWGreek',
+        'PICAWA': 'PICAWA',
         'uw_study': 'UWClub',
     }
 


kairstenfay:
[Add PICAWA to SCAN In Person Enrollments project](https://trello.com/c/vbtiShmG/830-add-picawa-to-scan-in-person-enrollments-project)

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -248,6 +248,7 @@ def site_map(record_location: str) -> str:
     """
     location_site_map = {
         'greek': 'UWGreek',
+        'PICAWA': 'PICAWA',
```
kschwabefry, line 251: I didn't see this value added into the redcap project yet (just childcare and greek), but as long as the value stays PICAWA, based on my read of the mapping, it would be just fine to deploy this even before its added in redcap 👍

kschwabefry (approved):

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
kairstenfay: Yep, the new site is still in the drafted changes (thanks for double-checking!). Izzy says the enrollments are happening tomorrow, so it may be best to deploy this before the REDCap changes are in production so that the ETL doesn't break.

joverlee521 (approved):

PR #148, opened 2020-09-30
Encounter Provenance

joverlee521:
~Add Encounter.meta.source to track the provenance information of the
encounter. For the REDCap DET ETLs, this is a URL with the format:~
```
https://redcap.iths.org/<project_id>/<record_id>/<event_name>/<repeat_instance>
```
~Event name and repeat instance number are only included if the
project has repeating events.~

See latest commits for details

Should be deployed along with https://github.com/seattleflu/id3c/pull/170

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/fhir.py
+++ b/lib/seattleflu/id3c/cli/command/etl/fhir.py
@@ -3,10 +3,13 @@ REDCap DET ETL shared functions to create FHIR documents
 """
 import logging
 import regex
+import json
 from itertools import filterfalse
 from typing import Iterable, NamedTuple, Optional, List, Any, Callable
 from uuid import uuid4
 from datetime import datetime
+from urllib.parse import quote
+from id3c.cli.redcap import Record as REDCapRecord
 from id3c.cli.command.de_identify import generate_hash
 
 
@@ -196,7 +199,8 @@ def create_location_resource(location_type: List[dict],
     return location_resource
 
 
-def create_encounter_resource(encounter_identifier: List[dict],
+def create_encounter_resource(encounter_source: str,
+                              encounter_identifier: List[dict],
                               encounter_class: dict,
                               encounter_date: str,
                               patient_reference: dict,
@@ -235,6 +239,7 @@ def create_encounter_resource(encounter_identifier: List[dict],
 
     encounter_resource = {
         "resourceType": "Encounter",
+        "meta": { "source": encounter_source },
         "class": encounter_class,
         "identifier": encounter_identifier,
         "status": encounter_status,
@@ -553,3 +558,29 @@ def observation_resource(device: str) -> Any:
             )
         )
     }
+
+
+def create_redcap_uri(record: REDCapRecord) -> str:
+    """
+    Create a JSON data URI representing the source of a given REDCap *record*.
+
+    Used in FHIR Encounter resources as the meta.source, which ultimately winds
+    up in ID3C's ``warehouse.encounter.details`` column.
+    """
+    data_scheme = 'data:application/json'
+
+    redcap_data = {
+        'url': record.project.base_url,
+        'project_id': record.project.id,
+        'record_id': record.id
+    }
+
+    if record.get('redcap_event_name'):
+        redcap_data['event_name'] = record.get('redcap_event_name')
+
+    if record.get('redcap_repeat_instance'):
+        redcap_data['repeat_instance'] = record.get('redcap_repeat_instance')
+
+    data = {'redcap': redcap_data}
+
+    return data_scheme + ',' + quote(json.dumps(data))
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_asymptomatic_swab_n_send.py
@@ -11,6 +11,7 @@ from typing import Any, Callable, Dict, List, Mapping, Match, Optional, Union, T
 from datetime import datetime
 from cachetools import TTLCache
 from id3c.db.session import DatabaseSession
+from id3c.cli.redcap import Record as REDCapRecord
 from id3c.cli.command.etl import redcap_det
 from id3c.cli.command.geocode import get_response_from_cache_or_geocoding
 from id3c.cli.command.location import location_lookup
@@ -23,7 +24,7 @@ from . import race, first_record_instance, required_instruments
 LOG = logging.getLogger(__name__)
 
 
-REVISION = 2
+REVISION = 3
 
 REDCAP_URL = 'https://redcap.iths.org/'
 INTERNAL_SYSTEM = "https://seattleflu.org"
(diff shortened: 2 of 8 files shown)


joverlee521:
[Add encounter provenance to `encounter.details`](https://trello.com/c/Rr3xrKQE/828-add-encounter-provenance-to-encounterdetails)

kairstenfay (approved):

jccraft-uw (approved):
This makes a lot of since and is easy to understand. The code looks good, and I tried this out locally with the the uw-reopening ETL.

lib/seattleflu/id3c/cli/command/etl/fhir.py
```diff
@@ -553,3 +558,29 @@ def observation_resource(device: str) -> Any:
...
+
+
+def create_redcap_uri(record: REDCapRecord) -> str:
+    """
+    Create a JSON data URI representing the source of a given REDCap *record*.
+
+    Used in FHIR Encounter resources as the meta.source, which ultimately winds
+    up in ID3C's ``warehouse.encounter.details`` column.
+    """
+    data_scheme = 'data:application/json'
```
jccraft-uw, line 570: Could we have some doctests for this function? Do we ever mock a REDCap record for tests?

lib/seattleflu/id3c/cli/command/etl/fhir.py
```diff
@@ -553,3 +558,29 @@ def observation_resource(device: str) -> Any:
...
+        'url': 'https://redcap.iths.org',
+        'project_id': record.project.id,
+        'record_id': record.id
+    }
+
+    if record.get('redcap_event_name'):
+        redcap_data['event_name'] = record.get('redcap_event_name')
+
+    if record.get('redcap_repeat_instance'):
+        redcap_data['repeat_instance'] = record.get('redcap_repeat_instance')
```
tsibley, line 582: I'm slightly worried that `redcap_repeat_instance` could be `0` and we'd skip it here. Maybe REDCap doesn't allow that though?

I wonder about using `'redcap_repeat_instance' in record` instead (and ditto for event name for consistency)?

lib/seattleflu/id3c/cli/command/etl/fhir.py
```diff
@@ -553,3 +558,29 @@ def observation_resource(device: str) -> Any:
...
+    """
+    Create a JSON data URI representing the source of a given REDCap *record*.
+
+    Used in FHIR Encounter resources as the meta.source, which ultimately winds
+    up in ID3C's ``warehouse.encounter.details`` column.
+    """
+    data_scheme = 'data:application/json'
+
+    redcap_data = {
+        'url': 'https://redcap.iths.org',
```
tsibley, line 573: This is a hideously tedious point (software 😑), but all the DETs we receive and our DET ETL routines use a URL with a trailing slash (`https://redcap.iths.org/`). I wonder if we should match that here to prevent future potential hair-pulling?

…

Ahhh, and looking now, we could get that value without hardcoding it here by using `record.project.base_url`.

tsibley (changes requested):
One change to REDCap URL I think would be worthwhile, and one change to optional field checking that I'm not sure how much it matters.

I think everyone realizes this, but I want to explicitly state it anyway and make sure we've considered the implications: this bumps the revisions for all our DET ETLs and will cause a very large reprocessing of essentially all DETs, thus refetching tens of thousands of records from REDCap and making associated upserts into the warehouse. I _think_ this will be fine, but I also wouldn't be surprised if something goes wrong during it that we didn't anticipate. Reprocessings of this size always make me a little nervous ("did we get everything right‽ will the code corrupt our data in a way we don't realize‽"). :grimacing:

lib/seattleflu/id3c/cli/command/etl/fhir.py
joverlee521: AH right, I'll avoid hardcoding the URL here.

lib/seattleflu/id3c/cli/command/etl/fhir.py
joverlee521: REDCap doesn't allow repeat instance of 0, it returns this error when I try to import a record with instance 0:
```
Values being imported into a repeating instrument/event cannot have the 'redcap_repeat_instance' field with a value of '0'.
It must have a numeric value greater than '0'.
```

lib/seattleflu/id3c/cli/command/etl/fhir.py
joverlee521: Hmm, I don't think we've created REDCap records for tests. Would that be too unwieldy for a doctest?

lib/seattleflu/id3c/cli/command/etl/fhir.py
joverlee521: I wanted to avoid ingesting empty strings. I believe currently the `enrollment_arm_1` event record has an empty string for `redcap_repeat_instance`

tsibley: Ah, great. Fine as-is then!

lib/seattleflu/id3c/cli/command/etl/fhir.py
tsibley: Mocking an `id3c.cli.redcap.Record` would be possible, but probably beyond the scope of a doctest. It could be done in a new file under _tests/_ though. But probably not a merge blocker?

jccraft-uw: I tried to make this so much more complicated than it really is!
Our REDCAP Record object is just a dictionary with a method for getting the project_id.
For refactoring work I just created a simple test like this, just building the dictionary.

>>> parse_date_field({'field1': '2000-2-12'}, 'field1')
    datetime.datetime(2000, 2, 12, 0, 0)

joverlee521:
> I think everyone realizes this, but I want to explicitly state it anyway and make sure we've considered the implications: this bumps the revisions for all our DET ETLs and will cause a very large reprocessing of essentially all DETs, thus refetching tens of thousands of records from REDCap and making associated upserts into the warehouse. I _think_ this will be fine, but I also wouldn't be surprised if something goes wrong during it that we didn't anticipate. Reprocessings of this size always make me a little nervous ("did we get everything right‽ will the code corrupt our data in a way we don't realize‽"). 😬

Would it be better to not bump all revisions, but backfill the `_provenance` details by splitting the `encounter.identifier`?
I think the only one we _need_ to bump it UW-retrospectives (`encounter.identifier` is hashed for these) and UW-reopening (we needed to bump the revision to ingest skipped testing_determination encounters)

kschwabefry:
> > I think everyone realizes this, but I want to explicitly state it anyway and make sure we've considered the implications: this bumps the revisions for all our DET ETLs and will cause a very large reprocessing of essentially all DETs, thus refetching tens of thousands of records from REDCap and making associated upserts into the warehouse. I _think_ this will be fine, but I also wouldn't be surprised if something goes wrong during it that we didn't anticipate. Reprocessings of this size always make me a little nervous ("did we get everything right‽ will the code corrupt our data in a way we don't realize‽"). 😬
> 
> Would it be better to not bump all revisions, but backfill the `_provenance` details by splitting the `encounter.identifier`?
> I think the only one we _need_ to bump it UW-retrospectives (`encounter.identifier` is hashed for these) and UW-reopening (we needed to bump the revision to ingest skipped testing_determination encounters)

I'm curious to watch how reprocessing the UW reopening project will go with the new batch size limitting.

It'd be nice if there was another way to spread out the reprocessing over a greater period of time. Deploying the bumped revision numbers project by project may help a bit with that?

lib/seattleflu/id3c/cli/command/etl/fhir.py
tsibley: > Our REDCAP Record object is just a dictionary with a method for getting the project_id.

Yes, though in the specific case of this PR, a plain dictionary look-a-like won't suffice as the code does access the `.project` and `.id` properties.

tsibley:
> Deploying the bumped revision numbers project by project may help a bit with that?

Hmm, yeah, maybe that's a good approach. Bump the two projects we need to do it for, because we can't backfill them (slash need updated data anyway) and do need this data in `_provenance`. Then later, bump others as necessary (if ever), or during other changes necessitating bumps.

kschwabefry (approved):

PR #149, opened 2020-10-03
etl redcap-det uw-reopening: Handle invalid birth date

kairstenfay:
If a given `core_birthdate` is invalid (e.g. year is 0000), return the
birth date as None and emit a warning.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -125,7 +125,7 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
         return None
 
     patient_entry, patient_reference = create_patient(enrollment)
-    birthdate = datetime.strptime(enrollment['core_birthdate'], '%Y-%m-%d')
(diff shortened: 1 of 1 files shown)


lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -1156,8 +1156,13 @@ def create_computed_questionnaire_response(record: dict, patient_reference: dict
...
     """
-
-    record['age'] = relativedelta(encounter_date, birthdate).years
+    # A birthdate of None will return a falsy relativedelta() object
+    delta = relativedelta(encounter_date, birthdate)
+    if not delta:
+        age = None
+    else:
+        age = delta.years
+    record['age'] = age
```
kairstenfay, line 1165: Oh second thought I think it might be better to avoid entering age altogether?

kairstenfay: My thoughts being the usual "null values is a violation of the FHIR docs". But I'm not quite sure how this would affect QuestionnaireResponse itemAnswer calculation and I don't really feel like getting too deeply into this at 8AM on a Saturday morning.

PR #150, opened 2020-10-06
setup.py: Remove id3c as a requirement

joverlee521:
The core id3c package is technically required for id3c-customizations,
but the editable install of id3c through pipenv does not resolve this
dependency. 

Remove id3c from setup.py so that we can update Pipfile.lock with the latest core id3c dependencies.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,6 +10,7 @@ pytest = "*"
 [packages]
 seattleflu-id3c-customizations = {editable = true,path = "."}
 id3c = {editable = true,path = "./../id3c"}
+fsspec = "*"
 
 [requires]
 python_version = "3.6"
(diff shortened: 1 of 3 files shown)


joverlee521:
> It looks like the tests didn't work, though.

Grrr, it's the fsspec issue again. Pushing up a fix with explicit install of fsspec.

kairstenfay:
How are you installing the latest id3c if not via `setup.py`?

joverlee521:
> How are you installing the latest id3c if not via `setup.py`?

The `Pipfile` lists the editable id3c package that points to your local id3c instance, so id3c will be installed when you run `pipenv install` or `pipenv update`. I don't think id3c was ever installed via `setup.py` because id3c is not available on pypi

kschwabefry (approved):
Thanks for digging into this behavior

kairstenfay (approved):
Thanks!

PR #151, opened 2020-10-08
Testing quota views

joverlee521:
Views for the testing quota system based on what @tsibley had outlined [here](https://github.com/seattleflu/id3c-customizations/pull/144/files#diff-275ef3202ccd63065aff184bf5c7f3b4R2-R72) and latest discussion on [slack](https://seattle-flu-study.slack.com/archives/C018V0WGGJ0/p1601664080020400) of what the priority order should be.

Depends on https://github.com/seattleflu/id3c-customizations/pull/148


joverlee521:
[Test quota views](https://trello.com/c/sNpBto9q/825-test-quota-views)

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2913,139 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+            when surge_selected is not null then surge_selected -- surge_selected is a text field with name of surge event
+            when tier = '1' then 'tier_1_surveillance'
+            when tier in ('2', '3') then 'tier_2_and_3_surveillance'
+            else null
+        end as priority_reason
+    from uw_individuals
+    join uw_enrollments using (individual)
+    left join uw_encounters using (individual)
+    -- Filter to participants who have never been invited to test or their last invite was over 7 days before encounter
+    where (latest_invite_date is null or latest_invite_date < uw_encounters.encountered - interval '7 days')
```
jccraft-uw, line 3003: I'm thinking about the scenario when a participant registers and never completes the daily attestation. That means that we won't have encounters. You're left joining to uw_encounters, so the where (latest_invite_date is null) part will capture this PT for baseline testing ... the PT has never been invited, so include them in the query.
In this scenario without daily attestations, if the PT enrolled and got baseline testing we'd have latest_invite_date populated.  Would this PT not get included in the priority queue any more because they don't have encounters?

joverlee521: Even without daily attestations, we would have an encounter if the PT did a baseline test (from kiosk or swab&send). The `latest_invite_date` would be null so they get included in the queue. We don't want them in the queue if the baseline test was within the last 7 days, so that's why I added the filter based on `latest_collection_date` below. 

Not sure if that makes senses? 🤔

jccraft-uw: Trying an example:
PT enrolls on 1/1/2021. They never do a daily attestation.
They fall into the `where (latest_invite_date is null)` and get baseline testing. Like you say, this will create an encounter.
Encounter gets created for date 1/20/2021. This is where the testing internal instrument gets marked, so 1/20/2021 is the latest_invite_date. They get in to test a couple days later, so Kiosk reg is on 1/22/2021. This now changes the encounter date to 1/22/2021. (Kiosk reg date trumps testing internal instrument date). So, latest_invite_date is now 1/22/2021.

This PT never does daily attestation, so there aren't additional encounters created.
latest_invite_date remains 1/22/2021
most recent encountered remains 1/22/2021

Now, latest_invite_date is not null and latest_invite_date is not 7 days before encountered.
It looks like this PT wouldn't be picked up for surveillance because they don't have any more encounters.

joverlee521: Ahhh thank you for the detailed example!! Hmmmm, will have to think through this more...

joverlee521: Okay, I _think_ I see a solution for this: 
Instead of trying to bundle both daily attestations and baseline/surveillance into one `select` query, I can separate them into two and combine them with `union all`. So the PT will always have a baseline/surveillance record in the queue even if they don't do daily attestations. Does that make sense?

jccraft-uw: Splitting that into two does make a lot of sense. You could even split into 3 .... baseline/surveillance, daily attestations, and surge,  if not for functionality then for making it easier to see what's going on where.
And the rules *might* be different for the different chunks. If a PT got tested six days ago but was today included in a surge test we might want to include them.

joverlee521: See if the latest fixup commit makes sense 🙏

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2913,198 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+                when coalesce(latest_invite_date, latest_collection_date) is null and tier = '1' then 'tier_1_baseline'
+                when coalesce(latest_invite_date, latest_collection_date) is null and tier in ('2', '3') then 'tier_2_and_3_baseline'
+                when tier = '1' then 'tier_1_surveillance'
+                when tier in ('2', '3') then 'tier_2_and_3_surveillance'
+                else null
+            end as priority_reason
+        from uw_enrollments
+        join uw_individuals using (individual)
+        -- Filter to enrollments that have never been invited to test or last invite was over 7 days before today
+        where (latest_invite_date is null or latest_invite_date < current_date - interval '7 days')
```
jccraft-uw, line 3032: current_date makes sense here to capture people who don't attest.

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2913,198 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+            tier,
+            latest_invite_date,
+            latest_collection_date,
+            6 as priority,
+            surge_selected as priority_reason
+        from uw_encounters
+        join uw_individuals using (individual)
+        join uw_enrollments using (individual)
+        where surge_selected is not null
+        and testing_trigger is null
```
jccraft-uw, line 3059: This does seem reasonable ... consider the person in need of a test if the surge_selected is marked and testing_trigger is null.

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2913,198 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+    )
+
+    select * from positive_daily_attestations
```
jccraft-uw, line 3062: We may want to wrap this with a big select *, min(priority) from <the union all queries> group by individual (and other columns) .... so that we return a given person only once.

tsibley: The inner view intentionally doesn't collapse by person; that's the job of the outer view below, which does:

```
select distinct on (individual)
```

tsibley (changes requested):
Thanks for working through this! I left some relatively small comments below. 

One general but untested concern I have is if the organization of the CTEs will impact performance. CTEs act as a query plan barrier and force Pg to materialize those as temp tables in either memory or disk; these CTEs may produce a lot of results. Particularly, the `uw_encounters` and `uw_enrollments` CTEs both include `all_uw_encounters.*` but then are later themselves joined together; this may bloat the row width of the temp tables Pg has to work with. Not sure. I don't think we need to pre-emptively change anything now, but let's keep an eye on how the view runs in production. There are some techniques for flattening large views like this that can dramatically improve evaluation time, but at the expense of the nice CTE factoring.

schema/deploy/roles/uw-priority-exporter/create.sql
```diff
@@ -0,0 +1,10 @@
+-- Deploy seattleflu/id3c-customizations:roles/uw-priority-exporter/create to pg
+
+begin;
+
+create role "uw-priority-exporter";
+
+comment on role "uw-priority-exporter" is
+    'Used to export data for the UW reopening priority queue';
```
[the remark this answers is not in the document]
joverlee521: Ah, that makes sense. I'll update the role created.

schema/deploy/shipping/views.sql
```diff
@@ -2672,7 +2674,7 @@ create or replace view shipping.uw_reopening_enrollment_fhir_encounter_details_v
   ,(select date_response[1] from shipping.fhir_questionnaire_responses_v1 where encounter_id = encounter.encounter_id and link_id = 'vaccine') as received_this_season_flu_vaccine_date
 
   from warehouse.encounter encounter
-  where identifier similar to 'https://redcap.iths.org/23854/\d+/enrollment_arm_1/'
+  where encounter.details @> '{"_provenance": {"redcap": {"project_id":23854, "event_name":"enrollment_arm_1"}}}'
```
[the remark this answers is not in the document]
joverlee521: Right, we have the REDCap URL for a reason 😄

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2913,198 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+            end as priority_reason
+        from uw_encounters
+        join uw_individuals using (individual)
+        join uw_enrollments using (individual)
+        -- Filter to encounters for participants whose last invite was over 7 days before encounter
+        where (latest_invite_date is null or latest_invite_date < uw_encounters.encountered - interval '7 days')
+        -- Filter to encounters for participants who have never had a sample collected or their last sample collection was over 7 days before encounter
+        and (latest_collection_date is null or latest_collection_date < uw_encounters.encountered - interval '7 days')
+        -- Filter for instances taht do no already have testing_trigger filled
+        and testing_trigger is null
```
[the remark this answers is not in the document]
joverlee521: I had assumed that we can rely on the REDCap/FHIR ETL to not ingest empty string values. 
We could specifically check `testing_trigger is null or trim(testing_trigger) = ''`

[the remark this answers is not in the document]
jccraft-uw: Right, the REDCap ETL won't write a questionnaire item if its answer is ''. 
But to be safe if would make sense to check.

schema/deploy/shipping/views.sql
jccraft-uw: Right.

joverlee521:
> One general but untested concern I have is if the organization of the CTEs will impact performance. CTEs act as a query plan barrier and force Pg to materialize those as temp tables in either memory or disk; these CTEs may produce a lot of results. Particularly, the uw_encounters and uw_enrollments CTEs both include all_uw_encounters.* but then are later themselves joined together; this may bloat the row width of the temp tables Pg has to work with. Not sure. I don't think we need to pre-emptively change anything now, but let's keep an eye on how the view runs in production. There are some techniques for flattening large views like this that can dramatically improve evaluation time, but at the expense of the nice CTE factoring.

This is a general concern that I have as well. I'm not entirely happy with the splitting & re-joining of the encounters and enrollments. What techniques are there for flattening large view like this? It would be helpful to talk through those 🙏

jccraft-uw:
> > One general but untested concern I have is if the organization of the CTEs will impact performance. CTEs act as a query plan barrier and force Pg to materialize those as temp tables in either memory or disk; these CTEs may produce a lot of results. Particularly, the uw_encounters and uw_enrollments CTEs both include all_uw_encounters.* but then are later themselves joined together; this may bloat the row width of the temp tables Pg has to work with. Not sure. I don't think we need to pre-emptively change anything now, but let's keep an eye on how the view runs in production. There are some techniques for flattening large views like this that can dramatically improve evaluation time, but at the expense of the nice CTE factoring.
> 
> This is a general concern that I have as well. I'm not entirely happy with the splitting & re-joining of the encounters and enrollments. What techniques are there for flattening large view like this? It would be helpful to talk through those

SQL Server has similar behavior with CTEs: the optimizer may choose to create temp tables behind the scenes. We found it often more efficient to go ahead and create the temp tables ourselves in some cases. We had the luxury of being inside stored procedures, so we could do that.

schema/deploy/shipping/views.sql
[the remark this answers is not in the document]
tsibley: Instead of sprinkling that empty string check in a few places, it might be a cleaner separation of concerns to use `nullif(…, '') as testing_trigger` in the view which provides the `testing_trigger` column. (And ditto with `surge_selected`.)

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2912,163 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+            sample.collected::date as sample_collection_date
+        from all_uw_instances
+        left join warehouse.sample using (encounter_id)
+        where redcap_event_name = 'encounter_arm_1'
+    ),
+
+    uw_individuals as (
+        select
+            individual,
+            max(uw_encounters.encountered) filter (where testing_trigger = '1') as latest_invite_date,
```
kairstenfay, line 2959: I saw this and realized that testing_trigger should be a boolean. See https://github.com/seattleflu/id3c-customizations/pull/153

tsibley:
> What techniques are there for flattening large view like this? It would be helpful to talk through those pray

Three techniques I've used in the past for queries of similar structure:

1. Pulling stuff out of CTEs and into sub-selects. This lets the query planner cut across the whole query more effectively and push conditions down when its more efficient to do so. Alternatively, you can sometimes pull CTEs into separate, additional "inner" views.

2. Restructuring the query to combine (potentially even partially overlapping) sets with unions instead of joins. You've done this already a bit in the final query, but the actual query plans might inform other places to do it.

If we [upgrade from Pg 10 → 12](https://www.postgresql.org/docs/12/sql-select.html#SQL-WITH), we could use `with … as not materialized (…)`. :-)

> We found it often more efficient to go ahead and create the temp tables ourselves in some cases.

Interesting! I wonder why that was. Do you know what was driving the difference between the implicit temp tables vs. your explicitly-created ones? One reason I can imagine is that you could also create (temporary) indexes on the temp tables, which you can't do with the implicit ones. I've used that technique for large alter tables requiring data transformations in a migration.

tsibley:
@joverlee521 Have you looked at performance of these views on production? Locally they're quite slow right now. (Though I'm not too worried as we haven't tried to optimize them yet.)

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2912,163 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+            end as priority_reason
+        from uw_encounters
+        join uw_individuals using (individual)
+        join uw_enrollments using (individual)
+        -- Filter to encounters for participants whose last invite was over 7 days before encounter
+        where (latest_invite_date is null or latest_invite_date < uw_encounters.encountered - interval '7 days')
+        -- Filter to encounters for participants who have never had a sample collected or their last sample collection was over 7 days before encounter
+        and (latest_collection_date is null or latest_collection_date < uw_encounters.encountered - interval '7 days')
+        -- Filter for instances that do no already have testing_trigger filled (null or empty string)
+        and (testing_trigger is null or testing_trigger = '')
```
kairstenfay, line 2998: This will change with #153

kairstenfay (commented):
This is looking great, Jover!

I think it would be nice for me and for future us to have a comment above either of the new views (internal or public facing one) to describe the range of priority numbers and what they map to in English. Could we add some documentation there?

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2913,200 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+            priority,
+            priority_reason
+        from shipping.__uw_priority_queue_v1
+        -- Use the highest priority encounter that is the latest instance
+        order by individual, priority asc nulls last, redcap_repeat_instance desc
+    )
+
+    select *
+    from distinct_individuals
+    order by priority asc
```
tsibley, line 3097: We should be explicit about how we want to break priority ties by specifying fallback columns to order by. Options I can think of (none mutually exclusive):

- Newest or oldest first by encountered (which one is better?)
- `age(latest_collection_date) desc nulls first` — Time since last collection (longest first)
- `age(latest_invite_date) desc nulls first` — Time since last invite (longest first)
- Randomly!
- Record id

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2913,200 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+        join uw_individuals using (individual)
+        join uw_enrollments using (individual)
+        -- Filter to encounters for participants whose last invite was over 7 days before encounter
+        where (latest_invite_date is null or latest_invite_date < uw_encounters.encountered - interval '7 days')
+        -- Filter to encounters for participants who have never had a sample collected or their last sample collection was over 7 days before encounter
+        and (latest_collection_date is null or latest_collection_date < uw_encounters.encountered - interval '7 days')
+        -- Filter for instances that do no already have testing_trigger filled (null or empty string)
+        and (testing_trigger is null or testing_trigger = '')
+        -- Filter for postive daily attestations only
+        and screen_positive
```
tsibley, line 3000: Ah, just remembering that we'll need to be more precise than this. For attestations where `daily_exposure_known_pos is true`, we only want to offer testing after a lag of X days. If we test too early after exposure, we risk negatives before there's enough virus to detect. I believe Trevor quoted 2-3 days as the optimum value for X, but we should confirm that value.

tsibley (changes requested):

joverlee521:
> @joverlee521 Have you looked at performance of these views on production? Locally they're quite slow right now. (Though I'm not too worried as we haven't tried to optimize them yet.)

When I ran the query for `shipping.__uw_priority_queue_v1` on the production database, I got:
```
 Planning time: 6.845 ms
 Execution time: 678784.889 ms
```

tsibley:
> Execution time: 678784.889 ms

Time to read some query plans and optimize!

schema/deploy/shipping/views.sql
joverlee521: I thought I was doing the fallback to the latest instance with the CTE above this:
```sql
order by individual, priority asc nulls last, redcap_repeat_instance desc
```

tsibley: Ah, that's within an individual's reasons to get tested. I'm talking about across individuals within a single priority level.

schema/deploy/shipping/views.sql
joverlee521: Ah, right. Do we need lag for both `daily_exposure_known_pos` and `daily_exposure`?
I can incoporate this into the case statement:
```sql
case
    when daily_symptoms then 1
    when daily_exposure_known_pos and current_date - uw_encounters.encountered >= 3  then 2
    when daily_exposure and current_date - uw_encounters.encountered >= 3 then 3
    else null
 end as priority,
```

schema/deploy/shipping/views.sql
joverlee521: Ah, I see. I thought that would be decided by the id3c command, but I can see it being easier to manage if just defined within the view itself.

I think the first fallback is tier (tier 1 get higher priority), then last collection date and last invite date?

jccraft-uw:
> > What techniques are there for flattening large view like this? It would be helpful to talk through those pray
> 
> Three techniques I've used in the past for queries of similar structure:
> 
> 1. Pulling stuff out of CTEs and into sub-selects. This lets the query planner cut across the whole query more effectively and push conditions down when its more efficient to do so. Alternatively, you can sometimes pull CTEs into separate, additional "inner" views.
> 2. Restructuring the query to combine (potentially even partially overlapping) sets with unions instead of joins. You've done this already a bit in the final query, but the actual query plans might inform other places to do it.
> 
> If we [upgrade from Pg 10 → 12](https://www.postgresql.org/docs/12/sql-select.html#SQL-WITH), we could use `with … as not materialized (…)`. :-)
> 
> > We found it often more efficient to go ahead and create the temp tables ourselves in some cases.
> 
> Interesting! I wonder why that was. Do you know what was driving the difference between the implicit temp tables vs. your explicitly-created ones? One reason I can imagine is that you could also create (temporary) indexes on the temp tables, which you can't do with the implicit ones. I've used that technique for large alter tables requiring data transformations in a migration.

I don't know for sure what was driving the difference. One thought is that SQL spent some time trying to use the CTE in memory and then switched to writing it to a temp table. We avoiding that "wasted time" by making the temp table upfront. We didn't index it.

jccraft-uw:
> > Execution time: 678784.889 ms
> 
> Time to read some query plans and optimize!

Oh dear.
Can we see if having permanent uw_enrollments and uw_encounters views helps at all?

joverlee521:
> > > Execution time: 678784.889 ms
> > 
> > 
> > Time to read some query plans and optimize!
> 
> Oh dear.
> Can we see if having permanent uw_enrollments and uw_encounters views helps at all?

I'm making progress doing some optimization by using subqueries (so far got it down to 85541.642 ms). Will push the updates soon.

schema/deploy/shipping/views.sql
tsibley: > I thought that would be decided by the id3c command, but I can see it being easier to manage if just defined within the view itself.

Yeah, I think the view should provide the authoritative state of the queue. This will help keep the command simpler by separating concerns, as well as make it much simpler when manually inspecting or writing reports on the queue.

> I think the first fallback is tier (tier 1 get higher priority), then last collection date and last invite date?

Sounds reasonable to me, but we should confirm with Trevor. Confirming doesn't need to block initial implementation/deploy. We may also want a final arbiter that can't (or is very unlikely to) be a tie, like individual identifier. This is better than record id because it's both stable (unlike random num) and (mostly) evenly distributed (like random num).

schema/deploy/shipping/views.sql
tsibley: Yes, both daily exposure questions. That case looks good to me, although I think interval conditions like this read easier:

```
age(uw_encounters.encountered) >= '3 days'
```

tsibley:
> I don't know for sure what was driving the difference. One thought is that SQL spent some time trying to use the CTE in memory and then switched to writing it to a temp table. We avoiding that "wasted time" by making the temp table upfront. We didn't index it.

Ah, this makes sense. I know that Pg will start using in memory tables and then spill over to disk, for which there is a penalty of course. There are some configuration knobs to tune this behaviour, too. Thanks for the insight!

schema/deploy/shipping/views.sql
tsibley: I've asked Trevor in #informatics.

schema/deploy/shipping/views.sql
tsibley: I've [asked Trevor in #informatics](https://seattle-flu-study.slack.com/archives/CCAA9RBFS/p1602628191080200).

joverlee521:
The latest fixup commit depends on https://github.com/seattleflu/id3c-customizations/pull/153

schema/deploy/shipping/views.sql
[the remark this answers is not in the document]
joverlee521: Updated `testing_trigger` and `surge_selected_flag` to be boolean values.

schema/deploy/shipping/views.sql
joverlee521: Updated with the fallback for priority in the order proposed on slack.

schema/deploy/shipping/views.sql
joverlee521: Updated case statement to only include exposure for `age(uw_encounters.encountered) >= '2 days'`

joverlee521:
I think I'm caught up with all the changes requested. 

This is the final query plan when running the `shipping.uw_priority_queue_v1` as a temp view on production:
```                                                                                                                                                                                                
 Subquery Scan on uw_priority_queue_v1  (cost=3543.41..3543.45 rows=3 width=212) (actual time=42889.101..42895.598 rows=8416 loops=1)
   ->  Sort  (cost=3543.41..3543.42 rows=3 width=244) (actual time=42889.097..42891.152 rows=8416 loops=1)
         Sort Key: distinct_individuals.priority, distinct_individuals.tier, (age((CURRENT_DATE)::timestamp with time zone, (distinct_individuals.latest_collection_date)::timestamp with time zone)) DESC, (age((CURRENT_DATE)::timestamp with time zone, (distinct_individuals.latest_invite_date)::timestamp with time zone)) DESC, distinct_individuals.individual
         Sort Method: quicksort  Memory: 2568kB
         CTE distinct_individuals
           ->  Unique  (cost=3543.25..3543.26 rows=3 width=212) (actual time=42831.779..42840.118 rows=8416 loops=1)
                 ->  Sort  (cost=3543.25..3543.26 rows=3 width=212) (actual time=42831.776..42834.436 rows=8639 loops=1)
                       Sort Key: positive_daily_attestations.individual, positive_daily_attestations.priority, positive_daily_attestations.redcap_repeat_instance DESC
                       Sort Method: quicksort  Memory: 2627kB
                       ->  Append  (cost=3543.14..3543.20 rows=3 width=212) (actual time=28068.023..42807.649 rows=8639 loops=1)
                             CTE all_uw_instances
                               ->  Nested Loop  (cost=926.03..3457.37 rows=208 width=200) (actual time=578.429..2728.443 rows=113567 loops=1)
                                     ->  Bitmap Heap Scan on encounter  (cost=925.61..1736.13 rows=208 width=1251) (actual time=577.244..1338.618 rows=113567 loops=1)
                                           Recheck Cond: (details @> '{"_provenance": {"redcap": {"url": "https://redcap.iths.org/", "project_id": 23854}}}'::jsonb)
                                           Heap Blocks: exact=36448
                                           ->  Bitmap Index Scan on encounter_details_idx  (cost=0.00..925.56 rows=208 width=0) (actual time=567.849..567.849 rows=116642 loops=1)
                                                 Index Cond: (details @> '{"_provenance": {"redcap": {"url": "https://redcap.iths.org/", "project_id": 23854}}}'::jsonb)
                                     ->  Index Scan using individual_pkey on individual  (cost=0.42..8.24 rows=1 width=68) (actual time=0.002..0.002 rows=1 loops=113567)
                                           Index Cond: (individual_id = encounter.individual_id)
                             CTE uw_encounters
                               ->  Nested Loop Left Join  (cost=0.42..63.83 rows=1 width=178) (actual time=578.703..6290.251 rows=102174 loops=1)
                                     ->  CTE Scan on all_uw_instances  (cost=0.00..4.68 rows=1 width=168) (actual time=578.640..2899.412 rows=102171 loops=1)
                                           Filter: (redcap_event_name = 'encounter_arm_1'::text)
                                           Rows Removed by Filter: 11396
                                     ->  Index Scan using sample_encounter_id_idx on sample  (cost=0.42..8.44 rows=1 width=8) (actual time=0.005..0.005 rows=0 loops=102171)
                                           Index Cond: (all_uw_instances.encounter_id = encounter_id)
                                     SubPlan 2
                                       ->  Index Scan using fhir_questionnaire_responses_unique_link_id_per_encounter on fhir_questionnaire_responses_v1  (cost=0.43..8.45 rows=1 width=1) (actual time=0.006..0.006 rows=1 loops=102174)
                                             Index Cond: ((encounter_id = all_uw_instances.encounter_id) AND (link_id = 'screen_positive'::text))
                                     SubPlan 3
                                       ->  Index Scan using fhir_questionnaire_responses_unique_link_id_per_encounter on fhir_questionnaire_responses_v1 fhir_questionnaire_responses_v1_1  (cost=0.43..8.45 rows=1 width=1) (actual time=0.004..0.004 rows=1 loops=102174)
                                             Index Cond: ((encounter_id = all_uw_instances.encounter_id) AND (link_id = 'daily_symptoms'::text))
                                     SubPlan 4
                                       ->  Index Scan using fhir_questionnaire_responses_unique_link_id_per_encounter on fhir_questionnaire_responses_v1 fhir_questionnaire_responses_v1_2  (cost=0.43..8.45 rows=1 width=1) (actual time=0.004..0.004 rows=1 loops=102174)
                                             Index Cond: ((encounter_id = all_uw_instances.encounter_id) AND (link_id = 'daily_exposure'::text))
                                     SubPlan 5
                                       ->  Index Scan using fhir_questionnaire_responses_unique_link_id_per_encounter on fhir_questionnaire_responses_v1 fhir_questionnaire_responses_v1_3  (cost=0.43..8.45 rows=1 width=1) (actual time=0.004..0.004 rows=1 loops=102174)
                                             Index Cond: ((encounter_id = all_uw_instances.encounter_id) AND (link_id = 'daily_exposure_known_pos'::text))
                                     SubPlan 6
                                       ->  Index Scan using fhir_questionnaire_responses_unique_link_id_per_encounter on fhir_questionnaire_responses_v1 fhir_questionnaire_responses_v1_4  (cost=0.43..8.45 rows=1 width=1) (actual time=0.002..0.002 rows=0 loops=102174)
                                             Index Cond: ((encounter_id = all_uw_instances.encounter_id) AND (link_id = 'testing_trigger'::text))
                                     SubPlan 7
                                       ->  Index Scan using fhir_questionnaire_responses_unique_link_id_per_encounter on fhir_questionnaire_responses_v1 fhir_questionnaire_responses_v1_5  (cost=0.43..8.45 rows=1 width=1) (actual time=0.002..0.002 rows=0 loops=102174)
                                             Index Cond: ((encounter_id = all_uw_instances.encounter_id) AND (link_id = 'surge_selected_flag'::text))
                             CTE uw_enrollments
                               ->  Nested Loop Left Join  (cost=0.45..21.71 rows=1 width=180) (actual time=5872.859..33279.045 rows=11396 loops=1)
                                     Join Filter: (all_uw_instances_1.individual = uw_encounters.individual)
                                     Rows Removed by Join Filter: 45023815
                                     ->  Nested Loop  (cost=0.42..13.20 rows=1 width=172) (actual time=0.031..367.812 rows=11396 loops=1)
                                           ->  CTE Scan on all_uw_instances all_uw_instances_1  (cost=0.00..4.68 rows=1 width=168) (actual time=0.002..39.438 rows=11396 loops=1)
                                                 Filter: (redcap_event_name = 'enrollment_arm_1'::text)
                                                 Rows Removed by Filter: 102171
                                           ->  Index Scan using encounter_pkey on encounter encounter_1  (cost=0.42..8.44 rows=1 width=4) (actual time=0.027..0.027 rows=1 loops=11396)
                                                 Index Cond: (encounter_id = all_uw_instances_1.encounter_id)
                                                 Filter: (details @> '{"_provenance": {"redcap": {"url": "https://redcap.iths.org/", "event_name": "enrollment_arm_1", "project_id": 23854}}}'::jsonb)
                                     ->  HashAggregate  (cost=0.03..0.04 rows=1 width=40) (actual time=0.516..1.877 rows=3951 loops=11396)
                                           Group Key: uw_encounters.individual
                                           ->  CTE Scan on uw_encounters  (cost=0.00..0.02 rows=1 width=41) (actual time=0.002..5852.103 rows=9867 loops=1)
                                                 Filter: ((testing_trigger IS TRUE) OR (sample_collection_date IS NOT NULL))
                                                 Rows Removed by Filter: 92307
                                     SubPlan 9
                                       ->  Index Scan using fhir_questionnaire_responses_unique_link_id_per_encounter on fhir_questionnaire_responses_v1 fhir_questionnaire_responses_v1_6  (cost=0.43..8.45 rows=1 width=4) (actual time=0.011..0.011 rows=1 loops=11396)
                                             Index Cond: ((encounter_id = encounter_1.encounter_id) AND (link_id = 'tier'::text))
                             CTE positive_daily_attestations
                               ->  Nested Loop  (cost=0.00..0.12 rows=1 width=212) (actual time=28068.020..42769.560 rows=223 loops=1)
                                     Join Filter: ((uw_encounters_1.individual = uw_enrollments.individual) AND ((uw_enrollments.latest_invite_date IS NULL) OR (uw_enrollments.latest_invite_date < (uw_encounters_1.encountered - '7 days'::interval))) AND ((uw_enrollments.latest_collection_date IS NULL) OR (uw_enrollments.latest_collection_date < (uw_encounters_1.encountered - '7 days'::interval))))
                                     Rows Removed by Join Filter: 15088081
                                     ->  CTE Scan on uw_encounters uw_encounters_1  (cost=0.00..0.03 rows=1 width=167) (actual time=584.918..615.407 rows=1324 loops=1)
                                           Filter: ((testing_trigger IS NULL) AND screen_positive AND (age((CURRENT_DATE)::timestamp with time zone, (encountered)::timestamp with time zone) <= '7 days'::interval))
                                           Rows Removed by Filter: 100850
                                     ->  CTE Scan on uw_enrollments  (cost=0.00..0.02 rows=1 width=44) (actual time=4.436..28.315 rows=11396 loops=1324)
                             CTE baseline_and_surveillance
                               ->  CTE Scan on uw_enrollments uw_enrollments_1  (cost=0.00..0.06 rows=1 width=212) (actual time=0.012..7.167 rows=8416 loops=1)
                                     Filter: (((latest_invite_date IS NULL) OR (latest_invite_date < (CURRENT_DATE - '7 days'::interval))) AND ((latest_collection_date IS NULL) OR (latest_collection_date < (CURRENT_DATE - '7 days'::interval))))
                                     Rows Removed by Filter: 2980
                             CTE surge_testing
                               ->  Nested Loop  (cost=0.00..0.05 rows=1 width=212) (actual time=18.527..18.527 rows=0 loops=1)
                                     Join Filter: (uw_encounters_2.individual = uw_enrollments_2.individual)
                                     ->  CTE Scan on uw_encounters uw_encounters_2  (cost=0.00..0.02 rows=1 width=164) (actual time=18.525..18.525 rows=0 loops=1)
                                           Filter: ((surge_selected_flag IS TRUE) AND (testing_trigger IS NULL))
                                           Rows Removed by Filter: 102174
                                     ->  CTE Scan on uw_enrollments uw_enrollments_2  (cost=0.00..0.02 rows=1 width=44) (never executed)
                             ->  CTE Scan on positive_daily_attestations  (cost=0.00..0.02 rows=1 width=212) (actual time=28068.023..42770.454 rows=223 loops=1)
                             ->  CTE Scan on baseline_and_surveillance  (cost=0.00..0.02 rows=1 width=212) (actual time=0.016..15.093 rows=8416 loops=1)
                             ->  CTE Scan on surge_testing  (cost=0.00..0.02 rows=1 width=212) (actual time=18.528..18.528 rows=0 loops=1)
         ->  CTE Scan on distinct_individuals  (cost=0.00..0.12 rows=3 width=244) (actual time=42831.799..42850.529 rows=8416 loops=1)
 Planning time: 21.586 ms
 Execution time: 42903.220 ms
(88 rows)
```

tsibley (approved):
LGTM. I added a few comments in d26b5e1 to explain some particularly non-obvious bits. I think our future selves (or others) will appreciate it.

On our call this morning I suggested we hold off on merging and deploying this until the other half of quotas is ready, in case the need to make other changes to the views arises from that work. However, I realize that this means the sqitch changes may need rebasing later onto other development in the meantime. To avoid that annoying situation, I'd be ok with merging+deploying this sooner than later. If other changes are necessary, I can rework the views at that point. Not a big deal, really.

schema/deploy/shipping/views.sql
```diff
@@ -2911,4 +2913,221 @@ comment on view shipping.uw_reopening_ehs_reporting_v1 is
...
+            redcap_event_name,
+            redcap_repeat_instance,
+            encountered,
+            individual,
+            tier,
+            latest_invite_date,
+            latest_collection_date,
+            case
+                when coalesce(latest_invite_date, latest_collection_date) is null and tier = '1' then 4
+                when coalesce(latest_invite_date, latest_collection_date) is null and tier in ('2', '3') then 5
```
kschwabefry, line 3035: Is baseline > surge, or vice versa? I think I've read it both ways in slack 😞 
e.g. https://seattle-flu-study.slack.com/archives/C018V0WGGJ0/p1601664080020400 and https://seattle-flu-study.slack.com/archives/C018V0WGGJ0/p1601941736078300?thread_ts=1601941315.074900&cid=C018V0WGGJ0

PR #152, opened 2020-10-13
Move common REDCap ETL code to redcap.py

jccraft-uw:
This change moves common code from redcap_det_uw_reopening.py to redcap.py so that the code can be reused for other REDCap ETL projects.

The work was primarily cut-and-paste, but some elements were parameterized to accommodate differences between REDCap projects and future needs. For example, in redcap.py, the internal system ("https://seattleflu.org" in redcap_det_uw_reopening.py) is passed in as a parameter.

For testing, I processed the same test record through the code in the master branch and through this feature branch. In a diff I see that the only differences are the reference and fullUrl uuids and the timestamp at the head of the document.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap.py
@@ -1,7 +1,32 @@
 """
 Functions shared by REDCap DET ETL
 """
-from typing import Optional
+from datetime import datetime
+from enum import Enum
+import re
+from typing import Dict, List, Mapping, Match, Optional, Tuple, Union
+
+from cachetools import TTLCache
+
+from . import race
+from .fhir import *
+from .redcap_map import map_sex, map_symptom, UnknownVaccineResponseError
+from id3c.cli.command.geocode import get_response_from_cache_or_geocoding
+from id3c.cli.command.location import location_lookup
+from id3c.cli.redcap import Record as REDCapRecord
+from id3c.db.session import DatabaseSession
+import logging
+
+
+LOG = logging.getLogger(__name__)
+
+# See https://terminology.hl7.org/1.0.0/CodeSystem-v3-ActCode.html for
+    # possible collection codes.
+    # HH = 'home health'
+    # FLD = 'field'
+class CollectionCode(Enum):
+    HOME_HEALTH = "HH"
+    FIELD = "FLD"
 
 
 def normalize_net_id(net_id: str=None) -> Optional[str]:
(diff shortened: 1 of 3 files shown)


jccraft-uw:
[Refactor uw-reopening ETL to share common code](https://trello.com/c/NPGmCuVN/813-refactor-uw-reopening-etl-to-share-common-code)

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -124,14 +111,49 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+            redcap_record_id = record_id,
+            system_identifier = INTERNAL_SYSTEM)
+    else:
+        patient_entry, patient_reference = create_patient_using_demographics(
+            sex = enrollment['core_sex'],
+            preferred_language = LANGUAGE_CODE[project_id],
+            first_name = enrollment['core_participant_first_name'],
+            last_name = enrollment['core_participant_last_name'],
+            birth_date = enrollment['core_birthdate'],
+            zipcode = enrollment['core_zipcode'],
```
kairstenfay, line 134: It's a shame we can't just pass the `enrollment` object and allow `create_patient_using_demographics()` to parse it. 
One day, all of the core REDCap fields will be the same. One day.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -161,11 +183,74 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+            parent_encounter_reference = None,
+            encounter_reason_code = None,
+            encounter_identifier_suffix = None,
+            system_identifier = INTERNAL_SYSTEM,
+            redcap_uri = create_redcap_uri(redcap_record_instance),
+            redcap_url = REDCAP_URL,
+            redcap_project_id = project_id,
+            redcap_record_id = record_id,
+            redcap_event_name = redcap_record_instance['redcap_event_name'],
+            redcap_repeat_instance = redcap_record_instance['redcap_repeat_instance'])
```
kairstenfay, line 253: This is a long list of arguments!!
I wonder if there's a data structure or a class we can use to improve this eventually.

kairstenfay (commented):
Thanks for doing this work, @jccraft-uw. 

I'm looking at the code you've repurposed and parametrized in `redcap.py`. We now have a few functions with many, many arguments. To me, this is a red flag that makes this function extremely difficult to invoke and maintain in the future. 

I'm now wondering if our (mostly) purely functional approach to ingesting REDCap projects is beginning to fail us. Perhaps there are better approaches (via data structures or classes) we can utilize to minimize the number of arguments we're passing around?

joverlee521:
> Perhaps there are better approaches (via data structures or classes) we can utilize to minimize the number of arguments we're passing around? I don't have any immediate ideas here, but I'd like to get this conversation started.

I agree. I'm seeing a lot of arguments that are `redcap_*`. I believe these are all properties of the `REDCapRecord` returned by the core redcap_det command. So instead of passing each property individually, these can be accessed from a single `REDCapRecord` argument.

jccraft-uw:
> > Perhaps there are better approaches (via data structures or classes) we can utilize to minimize the number of arguments we're passing around? I don't have any immediate ideas here, but I'd like to get this conversation started.
> 
> I agree. I'm seeing a lot of arguments that are `redcap_*`. I believe these are all properties of the `REDCapRecord` returned by the core redcap_det command. So instead of passing each property individually, these can be accessed from a single `REDCapRecord` argument.

Totally, but if we pass the whole REDCAPRecord the consumer has to know the keys in order to pull out most fields. project_id hangs as an attribute right there, but record id, instance id, etc. have to be read from the dictionary. 
How about if we add more properties to our Record object, to "promote" certain fields as attributes?

@kairstenfay I was thinking the same thing but didn't make the change because I don't see a lot of object oriented code in our codebase. I was thinking that it would be nice to create an instance of this redcap utility, passing in the system_identifier and redcap info, for example, and not have to provide those values to every function. Could we go that approach?

jccraft-uw:
> > > Perhaps there are better approaches (via data structures or classes) we can utilize to minimize the number of arguments we're passing around? I don't have any immediate ideas here, but I'd like to get this conversation started.
> > 
> > 
> > I agree. I'm seeing a lot of arguments that are `redcap_*`. I believe these are all properties of the `REDCapRecord` returned by the core redcap_det command. So instead of passing each property individually, these can be accessed from a single `REDCapRecord` argument.
> 
> Totally, but if we pass the whole REDCAPRecord the consumer has to know the keys in order to pull out most fields. project_id hangs as an attribute right there, but record id, instance id, etc. have to be read from the dictionary.
> How about if we add more properties to our Record object, to "promote" certain fields as attributes?
> 
> @kairstenfay I was thinking the same thing but didn't make the change because I don't see a lot of object oriented code in our codebase. I was thinking that it would be nice to create an instance of this redcap utility, passing in the system_identifier and redcap info, for example, and not have to provide those values to every function. Could we go that approach?

But looking at redcap.py again, there aren't a lot of functions that require the redcap_* pieces. Most functions require the caller to have pulled from the REDCap record the values (like first_name) that they need to pass in. Because the REDCap projects aren't standardized I couldn't put that parsing into redcap.py.

joverlee521:
> How about if we add more properties to our Record object, to "promote" certain fields as attributes?

This seems worth doing! `Record` already has it's `record_id` field as [a property](https://github.com/seattleflu/id3c/blob/master/lib/id3c/cli/redcap.py#L264). We can promote other fields such as `event_name` and `repeat_instance` in core. 

> Because the REDCap projects aren't standardized I couldn't put that parsing into redcap.py.

@jccraft-uw Does this seem like a good place start the config idea you've shared in the past? We can have a config file that maps the specific REDCap field names to our "core" field names and rename the fields before processing? I think things like the `QuestionnaireResponse` might have to stay custom within each individual ETL, but shared things like `Encounter`, `Patient` or `Specimen` can get standardized.

jccraft-uw:
> > How about if we add more properties to our Record object, to "promote" certain fields as attributes?
> 
> This seems worth doing! `Record` already has it's `record_id` field as [a property](https://github.com/seattleflu/id3c/blob/master/lib/id3c/cli/redcap.py#L264). We can promote other fields such as `event_name` and `repeat_instance` in core.
> 
> > Because the REDCap projects aren't standardized I couldn't put that parsing into redcap.py.
> 
> @jccraft-uw Does this seem like a good place start the config idea you've shared in the past? We can have a config file that maps the specific REDCap field names to our "core" field names and rename the fields before processing? I think things like the `QuestionnaireResponse` might have to stay custom within each individual ETL, but shared things like `Encounter`, `Patient` or `Specimen` can get standardized.

See what you think about the latest commit. It's a stop-gap. Ideally, we'd add the properties to Record in ID3C. 
Hm. Now that I've done that I think the better idea is to put it in Record, but I was hesitant to touch too much core stuff.
Travis CI just failed, so maybe that's Fate telling me I should put the properties onto Record instead of this stop-gap.

@joverlee521 I like the idea a lot about having a config that maps field names. I may not have time to implement this for Childcare, but I'll see. It would be *so* nice to write functions that can assume field names.

kairstenfay:
>> @kairstenfay I was thinking the same thing but didn't make the change because I don't see a lot of object oriented code in our codebase. I was thinking that it would be nice to create an instance of this redcap utility, passing in the system_identifier and redcap info, for example, and not have to provide those values to every function. Could we go that approach?
> But looking at redcap.py again, there aren't a lot of functions that require the redcap_* pieces. Most functions require the caller to have pulled from the REDCap record the values (like first_name) that they need to pass in. Because the REDCap projects aren't standardized I couldn't put that parsing into redcap.py.

@jccraft-uw I'm not sure if you're still looking for an answer here or if you answered yourself. But, the new commit looks great -- that looks to me like we're heading in the right direction.

> @jccraft-uw Does this seem like a good place start the config idea you've shared in the past? We can have a config file that maps the specific REDCap field names to our "core" field names and rename the fields before processing? I think things like the QuestionnaireResponse might have to stay custom within each individual ETL, but shared things like Encounter, Patient or Specimen can get standardized.

I **love** this idea! I agree that we'll probably run out of time before childcare, though.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -248,263 +351,12 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
     return create_bundle_resource(
         bundle_id = str(uuid4()),
         timestamp = datetime.now().astimezone().isoformat(),
-        source = f"{REDCAP_URL}{project_id}/{record_id}",
+        source = f"{REDCAP_URL}{str(redcap_record_instance.project.id)}/{redcap_record_instance.id}",
```
joverlee521, line 354: This is outside the for loop so `redcap_record_instance` does not exist, but you can use `enrollment` here to get project id and record id. 

Also side question, I thought the `str` conversion isn't needed for f-strings?

jccraft-uw: Nice catch, there. When I saw your comment I was all "but it worked". Ah, python's scoping ... redcap_record_instance is available outside the loop. I switched to enrollment because that is saner.

In the second commit I'm removing the str() from the f'strings.

lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -44,3 +69,628 @@ def normalize_net_id(net_id: str=None) -> Optional[str]:
...
+    return _create_patient(
+        sex = sex,
+        preferred_language = preferred_language,
+        first_name = None,
+        last_name = None,
+        birth_date = None,
+        zipcode = None,
+        record = record,
+        system_identifier = system_identifier,
+        unique_identifier = unique_identifier)
```
joverlee521, line 288: I'm not sure I understand why this needs to be two separate functions. With the `unique_identifier` being an optional argument, you can call the inner function directly without the two wrappers. The inner `_create_patient` function already does the right thing of checking whether or not `unique_identifier` is available.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -124,31 +108,62 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+    else:
+        patient_entry, patient_reference = create_patient_using_demographics(
+            sex = enrollment['core_sex'],
+            preferred_language = LANGUAGE_CODE[enrollment.project.id],
+            first_name = enrollment['core_participant_first_name'],
+            last_name = enrollment['core_participant_last_name'],
+            birth_date = enrollment['core_birthdate'],
+            zipcode = enrollment['core_zipcode'],
+            record = enrollment,
+            system_identifier = INTERNAL_SYSTEM)
```
joverlee521, line 131: Partnered with my other comment, this can be single function call with `_create_patient`. If `netid` is None, then the `_create_patient` function will do the right thing of falling back to using the demographic data.

joverlee521 (commented):
This is looking good. Would be great to refactor the SCAN ETL as well to use the shared functions to see any pitfalls that might have been missed from just refactoring the UW reopening ETL.

jccraft-uw:
> This is looking good. Would be great to refactor the SCAN ETL as well to use the shared functions to see any pitfalls that might have been missed from just refactoring the UW reopening ETL.

That will be a good task. So far the refactor is working for childcare.

lib/seattleflu/id3c/cli/command/etl/redcap.py
jccraft-uw: It's for better usability. I wanted explicit functions for creating a patient with a unique identifier or with demographics. This prevents the caller from having to pass in None values for first_name, last_name, birth_date, and zip_code if they want to use a unique identifier.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Same note as above--for usability. The caller doesn't need to think about first_name, last_name, birth_date, or zipcode if they're creating with a unique identifier. No need to pass in that bunch of parameters with None.

lib/seattleflu/id3c/cli/command/etl/redcap.py
joverlee521: Hmm, then why not make the demographic fields optional as well. So `create_patient` can do the right thing without the caller having to pass `None` values?

jccraft-uw: All the optional parameters have to come at the end of the function signature, right? I didn't really like sticking those all on the far right.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: @joverlee521 makes a good point here (caveat: I haven't tested this locally).
We can keep the block written here:
```py
if netid:
    ...
else:
    ...
```

and have each block call `_create_patient`. The function call can still look the same (we don't need to pass `first_name`, `last_name` etc in the first block if we rewrite `_create_patient` to have default values of `None` for these arguments. That might be nicer instead of explicitly passing `None` values in each situation, as `create_patient_using_unique_identifier` and `create_patient_using_demographics` currently do.

jccraft-uw: Thanks for the suggestion. I prefer having the two specific overloads.

jccraft-uw:
[Refactor uw-reopening ETL to share common code](https://trello.com/c/NPGmCuVN/813-refactor-uw-reopening-etl-to-share-common-code)

lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -44,3 +69,628 @@ def normalize_net_id(net_id: str=None) -> Optional[str]:
...
+        housing_type,
+        tract_full_url
+    )
+    address_entry = create_resource_entry(address_location, generate_full_url_uuid())
+
+    return [tract_entry, address_entry]
+
+
+def create_site_reference(location: str, site_map: dict, default_site: str,
+    system_identifier: str) -> Optional[Dict[str,dict]]:
```
joverlee521, line 184: Seeing how this is used in the [Childcare ETL](https://github.com/seattleflu/id3c-customizations/pull/154/files#diff-3360a443a510e893404249bad41afeca5eefb61e2d9f7e684e6c9369e2777226R181), the `location` and `site_map` parameters should be optional so the caller doesn't have to pass the `None` values.

jccraft-uw: For sure. I did that in the new branch for the redo.

PR #153, opened 2020-10-13
Testing trigger

kairstenfay:
Convert `testing_trigger` to a boolean, and stop ingesting the free text field `surge_selected`.

Depends on the `surge_selected_flag` to be in the production REDCap.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -56,7 +56,7 @@ class EventType(Enum):
     ENROLLMENT = 'enrollment'
     ENCOUNTER = 'encounter'
 
-REVISION = 1
+REVISION = 2
 
 REDCAP_URL = 'https://redcap.iths.org/'
 INTERNAL_SYSTEM = "https://seattleflu.org"
@@ -1137,14 +1137,17 @@ def create_testing_determination_internal_questionnaire_response(record: dict, p
     questionnaire we do want to return answers for all questions even if they are
     not answered.
     """
+    boolean_questions = [
+        'testing_trigger',
+    ]
 
     string_questions = [
-        'testing_trigger',
         'testing_type',
     ]
 
     question_categories = {
-        'valueString': string_questions
+        'valueBoolean': boolean_questions,
+        'valueString': string_questions,
     }
 
     return questionnaire_response(record, question_categories, patient_reference, encounter_reference, False)
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -1141,7 +1141,6 @@ def create_testing_determination_internal_questionnaire_response(record: dict, p
     string_questions = [
         'testing_trigger',
         'testing_type',
-        'surge_selected'
     ]
 
     question_categories = {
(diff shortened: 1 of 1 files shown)


tsibley:
+1 for boolean, but I think we still need surge selected. Why remove it?

kairstenfay:
> +1 for boolean, but I think we still need surge selected. Why remove it?

In REDCap it says:
>The name of the surge event (for example, "Lander Hall surge on 11/04/2020") that the participant has been selected for

That feels like sensitive information to me. But I also understand the need to keep it or some form of it.

kairstenfay:
> That feels like sensitive information to me. But I also understand the need to keep it or some form of it.

I don't have the context of quotas in mind here. How important is the actual value of `surge_selected` versus just an evaluation of whether it's an empty string or not?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -1137,15 +1137,17 @@ def create_testing_determination_internal_questionnaire_response(record: dict, p
     questionnaire we do want to return answers for all questions even if they are
     not answered.
     """
+    boolean_questions = [
+        'testing_trigger',
+    ]
 
     string_questions = [
-        'testing_trigger',
         'testing_type',
```
jccraft-uw, line 1145: It looks like testing_type turned out to have numeric values instead of text. You could make this an integer question if you wanted.

kairstenfay: Hmm, we should be ingesting the label of testing_type, not the raw integer code for it.

joverlee521: These don't have human readable codes for them, so we would have to map the integer code to the labels.
(Sorry, I must have missed this when I asked them to update codings)

kairstenfay: How can you tell @joverlee521 ? In REDCap, the data dictionary looks like this: 
image

joverlee521: The first column is the coding that we pull in for the ETL. Compare to other fields that do have human readable codes:

Screen Shot 2020-10-13 at 1 54 18 PM

kairstenfay: Got it, thanks.

kairstenfay: Ah, in that case I'm going to leave `testing_type` as a string for now because we'll probably want to convert it to a human readable code at some point. Does that sound OK?

joverlee521: Yeah, not even sure we truly need to ingest `testing_type`. I'm not currently using this field in the quota query.

jccraft-uw: The quota system should set a value in testing_type. If we ingest it it could compare to what's already there.

kairstenfay:
I just pushed a commit that ingests `surge_selected_flag`. I'll wait until Jessica confirms this field is live in REDCap before merging.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
kairstenfay: I'll leave it alone for now.

jccraft-uw (approved):

joverlee521 (approved):

kschwabefry (approved):

joverlee521:
> I just pushed a commit that ingests `surge_selected_flag`. I'll wait until Jessica confirms this field is live in REDCap before merging.

Looks like the `surge_selected_flag` is now in the REDCap project. Good to merge 👍

PR #155, opened 2020-10-20
Uw reopening refactor take 2

jccraft-uw:
This is a re-do of the UW reopening ETL refactoring to move reusable code into redcap.py. 
I created a commit for each related chunk of code being refactored to make it easier follow the changes.
The "move to redcap.py" commit cut the code from redcap_det_uw_reopening.py and pasted it into redcap.py.
The "cleanup" commit afterwards cleaned up some things I saw while refactoring. This does bump the revision.

For testing I created the FHIR document from the master branch and diff'ed it against that which I created from this branch. I also ingested that FHIR bundle just to make sure that works.


jccraft-uw:
[Refactor uw-reopening ETL to share common code](https://trello.com/c/NPGmCuVN/813-refactor-uw-reopening-etl-to-share-common-code)

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -125,7 +125,7 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
         return None
 
     patient_entry, patient_reference = create_patient(enrollment)
-    birthdate = parse_birth_date(enrollment)
+    birthdate = parse_date_from_string(enrollment.get('core_birthdate'))
```
kschwabefry: Since the log level was downgraded from warning > debug inside the method to make it more general purpose, should we check here that birthdate is not null and otherwise warn? I wonder if having a birthdate we couldn't parse is a case we would need to take action on to manually debug/fix.

lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -44,3 +69,630 @@ def normalize_net_id(net_id: str=None) -> Optional[str]:
...
+    """
+    Returns a FHIR Patient resource entry and reference.
+    Uses a unique identifier to create the patient identifier
+    """
+    return _create_patient(
+        sex = sex,
+        preferred_language = preferred_language,
+        record = record,
+        system_identifier = system_identifier,
+        unique_identifier = unique_identifier)
```
joverlee521, line 289: I want to continue the conversation that I had started in the [previous PR](https://github.com/seattleflu/id3c-customizations/pull/152#discussion_r506711947). This is definitely _not_ a blocker, I just want to understand the reasoning of using two overloads vs using the inner function with different arguments. (Would love to learn if this is a coding practice that I'm not aware of).

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -689,10 +681,16 @@ def create_computed_questionnaire_response(record: REDCapRecord, patient_referen
...
+
     if not delta:
         age = None
     else:
-        age = delta.years
+        # Age Ceiling
+        try:
+            age = age_ceiling(delta.years)
+        except ValueError:
+            age = None
```
joverlee521, line 692: Is the try/except needed now that we apply `age_ceiling` directly to the `delta.years`? It was previously added to catch the int conversion error if `core_age_years` was an empty string.

lib/seattleflu/id3c/cli/command/etl/redcap.py
jccraft-uw: The optional parameters on  _create_patient are: 
1 first_name
2 last_name
3 birth_date
4 zipcode
5 unique_identifier

But they are *conditionally* optional.

If you provide a value for 5, you don't need to provide a value for 1 through 4. But if you don't provide a value for 5, you should provide a value for 1, 2, 3, and 4. (I've added a debug statement to warn if any of the values of 1 through 4 are None when 5 is None. I'm sure there will be cases when we don't have 4 but have 1, 2, and 3, so I didn't want to error on that.)

So that the caller doesn't have to keep track of which parameters are required in which scenarios I made the two overloads in which all parameters are required.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: That's a really good idea. We would want to investigate an HCT record that didn't have its `core_birthdate`. I added a LOG.warning and am continuing to process the record.

lib/seattleflu/id3c/cli/command/etl/redcap.py
joverlee521: Ah, I see. Thank you for the explanation 🙏

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Good point! That try/except isn't needed anymore, so I've removed it in the latest commit.
If the incoming birthdate is None then relativedelta will return None, and we'll not set an age.
If the incoming birthdate is not a date then relativedelta will throw a TypeError.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -124,31 +110,64 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+    location_resource_entries = build_residential_location_resources(
+        db = db,
+        cache = cache,
+        housing_type = enrollment.get('core_housing_type'),
+        primary_street_address = enrollment['core_home_street'],
+        secondary_street_address = enrollment['core_apartment_number'],
+        city = enrollment['core_home_city'],
+        state = enrollment['core_home_state'],
+        zipcode = enrollment['core_zipcode'],
+        system_identifier = INTERNAL_SYSTEM)
```
kairstenfay, line 152: Because the `build_residential_location_resources` function accepts information that comes from `core_*` fields, it would be better to given the function the `enrollment` object and allow it to parse for the various housing_type, primary_street_address, etc fields. 
Having fewer arguments in a function is generally considered better practice, making the function easier to maintain, call, and understand.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -124,31 +110,64 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+
     persisted_resource_entries = [patient_entry, *location_resource_entries]
 
     for redcap_record_instance in redcap_record_instances:
 
         event_type = None
         collection_method = None
```
kairstenfay, line 159: There's no longer a need to initialize these variables as `None`

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -161,11 +180,68 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+            encounter_date = encounter_date,
+            patient_reference = patient_reference,
+            site_reference = site_reference,
+            locations = location_resource_entries,
+            diagnosis = diagnosis,
+            contained = contained,
+            collection_code = collection_code,
+            parent_encounter_reference = None,
+            encounter_reason_code = None,
+            encounter_identifier_suffix = None,
```
kairstenfay, line 242: These are already optional parameters, so there's no need to pass `None` to them here.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -220,13 +312,23 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+                    encounter_identifier_suffix = "_follow_up",
+                    system_identifier = INTERNAL_SYSTEM,
+                    record = redcap_record_instance)
+
                 follow_up_questionnaire_entry = create_follow_up_questionnaire_response(
                 redcap_record_instance, patient_reference, follow_up_encounter_reference)
                 follow_up_computed_questionnaire_entry = create_computed_questionnaire_response(
                 redcap_record_instance, patient_reference, follow_up_encounter_reference,
-                birthdate, datetime.strptime(follow_up_encounter_entry['resource']['period']['start'], '%Y-%m-%d'))
+                birthdate, parse_date_from_string(follow_up_encounter_entry['resource']['period']['start']))
```
kairstenfay, line 331: This line (and the ones immediately above it) are under-indented, making this part of the code harder to read.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -524,120 +376,13 @@ def non_tract_locations(resource: dict):
...
-        note = 'never-tested'
     else:
-        note = 'can-test'
-
-    specimen_type = 'NSECR'  # Nasal swab.
-    specimen_resource = create_specimen_resource(
-        [specimen_identifier], patient_reference, specimen_type, received_time,
-        collection_date(record, collection_method), note
-    )
+        return None
```
kairstenfay, line 380: This else statement is unnecessary. In the next line, we'll return `encounter_date = None` if we don't enter the first two conditional blocks.

kairstenfay (changes requested):

kairstenfay:
As I mentioned in Slack, thank you for reworking this PR. Having separate commits for changing + moving code makes this much much easier to review! I left some suggestions that would make the code style here more similar to the existing id3c code base.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -781,10 +779,3 @@ def create_daily_questionnaire_response(record: REDCapRecord, patient_reference:
         patient_reference = patient_reference,
         encounter_reference = encounter_reference,
         system_identifier = INTERNAL_SYSTEM)
-
-class UnknownRedcapZipCode(ValueError):
-    """
-    Raised by :function: `zipcode_map` if a provided *redcap_code* is not
-    among a set of expected values.
-    """
-    pass
```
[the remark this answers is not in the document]
jccraft-uw: This error class was abandoned before the refactoring. This wasn't needed after HCT starting using zipcodes as the value in REDCap. I got rid of the zipcode_map function but didn't get rid of this error class that it used.
In the 'cleanup' commit I'm getting rid of it.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -321,8 +331,47 @@ class UnknownRedcapRecordLocation(ValueError):
...
+        """
+        location = location_lookup(db, lat_lng, 'tract')
+
+        if location and location.identifier:
+            return create_location(
+                f"{system_identifier}/location/tract", location.identifier, housing_type
+            )
+        else:
+            LOG.debug("No census tract found for given location.")
+            return None
```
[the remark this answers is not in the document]
jccraft-uw: The else there is just letting us do the LOG.debug.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: In this refactor in redcap.py, I'm not having code access REDCap fields. Instead, the values are passed in as parameters, and accessing the REDCap fields is left to the caller. Once the "core" fields settle down (maybe it's me seeing them get used in another project) I'd be likely to add a build_residential_location_resources_using_core (or some name like that) type of functions to redcap.py.

lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -44,3 +69,632 @@ def normalize_net_id(net_id: str=None) -> Optional[str]:
...
+    else:
+        redcap_event_name = ""
+    if record.repeat_instance:
+        redcap_repeat_instance = str(record.repeat_instance)
+    else:
+        redcap_repeat_instance = ""
+    if not encounter_identifier_suffix:
+        encounter_identifier_suffix = ""
+    encounter_id = f"{record.project.base_url}{record.project.id}/{record.id}/{redcap_event_name}/" + \
+        f"{redcap_repeat_instance}{encounter_identifier_suffix}"
```
joverlee521, line 455: I wonder about keeping this outside of the shared `create_encounter` function and instead have an `encounter_id` parameter that gets passed in. I'm saying this because I was thinking it would be great to start off with hashed `encounter.identifier` for the Childcare study since all of the REDCap metadata is available in `encounter.details`

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Because variable values can live across for-loop instances (because the python variable scope is the function and not the loop) I like to explicitly "clear them out" so there's no risk of data left from the previous loop iteration.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -417,43 +438,87 @@ def create_location(system: str, value: str, location_type: str, parent: str=Non
...
     else:
         patient_id = generate_patient_hash(
-            names       = (record['core_participant_first_name'], record['core_participant_last_name']),
-            gender      = gender,
-            birth_date  = record['core_birthdate'],
-            postal_code = record['core_zipcode'])
+                names       = (first_name, last_name),
+                gender      = gender,
+                birth_date  = birth_date,
+                postal_code = zipcode)
```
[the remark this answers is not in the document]
jccraft-uw: Fixed in "FIXUP: Kairsten's feedback" commit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -542,6 +653,45 @@ def build_diagnosis(symptom: str) -> Optional[dict]:
...
+    return contained, diagnosis
+
+
+def follow_up_encounter_reason_code() -> dict:
+    encounter_reason_code = create_codeable_concept(
+        system = "http://snomed.info/sct",
+        code = "390906007",
+        display = "Follow-up encounter"
+    )
+    return encounter_reason_code
```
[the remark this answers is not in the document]
jccraft-uw: Fixed in "FIXUP: Kairsten's feedback" commit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -282,14 +282,30 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+                specimen_reference = specimen_reference,
+                patient_reference = patient_reference,
+                encounter_reference = initial_encounter_reference)
         else:
             LOG.info("Creating encounter for record instance without sample")
 
         if specimen_received and not specimen_entry:
-            LOG.warning("Skipping record instance with insufficent information to construct a specimen")
+            LOG.warning("Skipping record instance. We think the specimen was received," \
+                + " but we're unable to create the specimen_entry.")
```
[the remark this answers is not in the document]
jccraft-uw: Thanks for that. That is nice.

Fixed in a few places in "FIXUP: Kairsten's feedback" commit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Fixed in "FIXUP: Kairsten's feedback" commit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Fixed in "FIXUP: Kairsten's feedback" commit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Fixed in "FIXUP: Kairsten's feedback" commit.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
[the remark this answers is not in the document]
kairstenfay: My comment remains the same. We do not need the `else:` statement. It can be dropped, but the logging and returning None can be kept.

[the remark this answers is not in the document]
jccraft-uw: In the `FIXUP: Kairsten's feedback` I reversed the logic so that it reads more easily.

lib/seattleflu/id3c/cli/command/etl/redcap.py
jccraft-uw: Absolutely. Thanks for catching that. It makes sense for the caller to create the encounter_id and pass it into create_encounter. I've made that change in the "FIXUP: create_encounter : ID as param" commit.

lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -44,3 +69,616 @@ def normalize_net_id(net_id: str=None) -> Optional[str]:
...
+        "location": create_reference(
+            reference_type = "Location",
+            identifier = create_identifier(f"{system_identifier}/site", site)
+        )
+    }
+
+
+def _create_patient(sex: str, preferred_language: str, record: REDCapRecord,
+        system_identifier: str, first_name: str = None, last_name: str = None,
+        birth_date: str = None, zipcode: str = None, unique_identifier: str = None) -> tuple:
```
kschwabefry, line 215: I wonder here about including the record itself as well as all the parsed fields individually (sex, name, birthdate, etc) -it seems perhaps redundant to include both, rather than just parse out the fields from redcaprecord inside of _create_patient. Do we identify the fields differently from redcap record depending on the caller?

kschwabefry (approved):
I think this is helpful work. Thanks for restructuring the commit history here Chris.

lib/seattleflu/id3c/cli/command/etl/redcap.py
jccraft-uw: The notion of "core" REDCap fields is relatively new. HCT uses them, and we got Childcare to use them. I am having callers pass in particular values as parameters instead of having code inside redcap.py assume field names. I wanted to see the "core" fields stick before I felt comfortable assuming them in utility code.
@joverlee521 had a really neat idea about leveraging a mapping configuration outside the code that would map a project's REDCap fields to "core" fields for projects that didn't confirm to core. That would be neat because then we could safely assume field names inside the utility.

kschwabefry: Oh I like the mapping config idea. Makes sense, that will make it flexible to be used more broadly for now if we want to use this in older projects also.

joverlee521 (approved):
Thank you for catching that `age_ceiling` wasn't being used in the UW ETL! 🙏 

It was very helpful to see the refactored functions used in the new Childcare ETL.

kairstenfay (approved):
Thanks for making my suggested changes, Chris. I haven't had time to look at them in detail but I approve if others do, too.

PR #156, opened 2020-10-23
reportable_conditions notify: add notifications to new Childcare channel

joverlee521:
hCoV-19 results for the Childcare also need to be sent to the new
`#ncov-reporting-childcare` channel because some members of the
Childcare study should not have access to the general `#ncov-reporting`
channel.

Due to time restrictions, this addition is done very lazily by
verifying all responses from Slack have a 200 status code. This does
mean that if one POST request fails, all of them will be sent again
on the next run of the notify command. The chance of this happening
seems pretty small, but we can revisit this at a later date if it
becomes a common issue.

EDIT: TODO:

- [x] Add new `SLACK_WEBHOOK_REPORTING_HCOV19_CHILDCARE` env var to backoffice server


jccraft-uw (approved):
This was easy to follow.

kschwabefry (approved):

joverlee521:
FYI, notifications worked as expected when I used the webhook for my DM and modified records to Childcare sites in my local database.

kairstenfay (approved):

PR #161, opened 2020-10-28
Add valueDecimal support to a shipping view

jccraft-uw:
In recent projects we started capturing questionnaire items as the
valueDecimal FHIR value type. Today, we're capturing `bmi` as valueDecimal.
We need to expose these in the shipping.fhir_questionnaire_responses_v1
view.

In python we capture these as `float`. In the shipping view we expose them
as the postgres `double precision` type. This type allows for 15 digits
of precision. We probably don't need that much precision for `bmi` but we
may for other items that get stored as valueDecimal.

** I'll create the sqitch tag just before deploying.

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -141,6 +141,7 @@ create materialized view shipping.fhir_questionnaire_responses_v1 as
            bool_and("valueBoolean") as boolean_response,
            array_remove(array_agg("valueDate" order by "valueDate"), null) as date_response,
            array_remove(array_agg("valueInteger" order by "valueInteger"), null) as integer_response,
+           array_remove(array_agg("valueDecimal" order by "valueDecimal"), null) as double_response,
            array_remove(array_agg("code" order by "code"), null) as code_response
       from warehouse.encounter,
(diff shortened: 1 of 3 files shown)


joverlee521:
@jccraft-uw Did you also want to add/uncomment the decimal bmi fields to the UW reopening queries in this same rework?

jccraft-uw:
> @jccraft-uw Did you also want to add/uncomment the decimal bmi fields to the UW reopening queries in this same rework?

That would make a lot of sense, wouldn't it?! I did that as a separate commit. Thanks.

joverlee521 (approved):
Looks good to me 👍 

If you happen to deploy this today, the new sqitch tag should be `@2020-10-28b`. This follows the naming convention that we've been using for same day tags. (You can find examples within `sqitch.plan`).

kschwabefry (approved):
Thanks for the detailed commit message! Really useful to help me follow along.

jccraft-uw:
> Looks good to me 
> 
> If you happen to deploy this today, the new sqitch tag should be `@2020-10-28b`. This follows the naming convention that we've been using for same day tags. (You can find examples within `sqitch.plan`).

Hey, thanks. And I saw the "later on ... " part with the 'b'.

jccraft-uw:
This is deployed.

PR #162, opened 2020-10-29
Handle valueDecimal items as numeric in view

jccraft-uw:
Feedback was that numeric, not double precision, is the better choice.

I have the sqitch tag in here with the thinking that this will be approved soon. If not I will remove it.


jccraft-uw:
[shipping.fhir_questionnaire_responses_v1 : for valueDecimal use the numeric pg type instead of double precision](https://trello.com/c/PIrKCJjT/870-shippingfhirquestionnaireresponsesv1-for-valuedecimal-use-the-numeric-pg-type-instead-of-double-precision)

tsibley (approved):

jccraft-uw:
Deployed.

PR #163, opened 2020-10-29
Add missing files related to preview 2 db reworks

jccraft-uw:
In my sqitch reworks yesterday and today I forgot to check in some files,
so this commit is adding them.


kairstenfay (approved):
Thanks for tracking these down!

jccraft-uw:
Deployed.

PR #165, opened 2020-10-30
Offer testing to records in the UW HCT priority queue

tsibley:
Code is ready for review. Details in the commit messages, as normal.

Still some things to do here:

- [x] End to end testing with REDCap. I've done a bit of this, but in individual pieces. This is made more difficult by not having a REDCap dev instance.
- [x] Do we need to insert a synthetic DET for each record offered testing since REDCap API imports don't send DETs? Or do other record changes happen (from calculated fields?) that will trigger DETs?
- [x] There are some outstanding todos which I _think_ are ok to iterate on after an initial deploy, but I would appreciate people's thoughts. I'll get them written up here in a moment.
- [x] Grants to uw-priority-queue-processor role

Deploy:

- [ ] backoffice cronjob
- [ ] Grant `uw-priority-queue-processor` to backoffice Pg user
- [ ] Populate `operations.test_quota` from data provided by enrollment/lab teams
- [x] Dashboard providing visibility into the queue
- [ ] Check in re: surveillance criteria

Next iterations:

 1. Deal with the potential failure/delay of the roundtrip update of the testing flag through REDCap and back into ID3C? https://trello.com/c/HQzkV5YD/886-uw-priority-queue-deal-with-failure-delay-of-roundtrip-update

 2. Implementation of "we won't be able to offer you a test" notifications. https://trello.com/c/QkT5v73z/887-uw-priority-queue-implement-we-cant-test-you-flagging

 3. Implementation of double testing of exposures.  https://trello.com/c/yUaPNY6H/888-uw-priority-queue-implement-double-testing-of-exposures

Note that re-testing of inconclusive/failed samples, as currently planned, will not be handled by the priority queue.  These will be manually marked in REDCap by clinical staff and REDCap will notify participants.  Volume is expected to be low enough that these won't impact lab quotas and thus can be unmanaged by the priority queue.


tsibley:
(Travis won't work because it tests against ID3C master, and this depends on https://github.com/seattleflu/id3c/pull/181.)

joverlee521:
Haven't taken a deep dive into the code yet, but the `uw-priority-queue-processor` role needs grants to the new `operations` schema and `operations.test_quota` table

joverlee521 (commented):
> Do we need to insert a synthetic DET for each record offered testing since REDCap API imports don't send DETs? Or do other record changes happen (from calculated fields?) that will trigger DETs?

Hmm, if we don't, the worst case scenario is the record never gets updated via other means and the record is stuck in the quota (until it falls off after 7 days).

lib/seattleflu/id3c/cli/command/offer_uw_testing.py
```diff
@@ -0,0 +1,232 @@
...
+
+        if log_offers:
+            dump_ndjson(offers)
+
+        # Token will automatically come from the environment.  If we're doing a
+        # dry run, then Project will make sure we update_records() doesn't
+        # actually update records.
+        project = Project(url, project_id, dry_run = dry_run)
+
+        offer_count += project.update_records(offers)
```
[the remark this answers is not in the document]
tsibley: Yep. We could also schedule the offering job to run more than once per timespan, to account for transient failures the first time.

tsibley:
> Hmm, if we don't, the worst case scenario is the record never gets updated via other means and the record is stuck in the quota (until it falls off after 7 days).

Right, but if it's high enough priority, it could suck up quota limits for every timespan until it's updated or falls out.

Hmm. As an alternative to relying on the roundtrip data flow to work correctly most of the time, the offering code could double-check the record status in REDCap and skip any that were in the queue but which REDCap already indicates have been offered testing.

schema/deploy/operations/schema.sql
```diff
@@ -0,0 +1,9 @@
+-- Deploy seattleflu/id3c-customizations:operations/schema to pg
+
+begin;
+
+create schema operations;
+
+comment on schema operations is 'Data supporting opertions management processes';
```
jccraft-uw, line 7: Gasp! A typo in the comment ...supporting operations.

tsibley: 🙃 Thanks!

tsibley:
Latest commit (c83a6ce) adds generation and insertion of synthetic DETs. This helps complete the roundtrip loop, but it may still fail if the REDCap DET ETL isn't running often enough, so we may still want to consider additional measures. Relies on further refactoring in https://github.com/seattleflu/id3c/pull/181.

joverlee521:
> Hmm. As an alternative to relying on the roundtrip data flow to work correctly most of the time, the offering code could double-check the record status in REDCap and skip any that were in the queue but which REDCap already indicates have been offered testing. This wouldn't be exclusive of also synthesizing DETs or other solutions to make the roundtrip more robust.

Oof, that sounds like a plate of data spaghetti 🤯 

Since you've added the synthesized DETs, if we time the cronjob for test offer to be right after the REDCap ETL/FHIR ETL, we should be good without the additional REDCap check?

tsibley:
> Oof, that sounds like a plate of data spaghetti :exploding_head:

:-) Not as bad as it sounds, I think. One implementation might be a loop that intentionally overshoots to reduce requests to REDCap:

1. Start with an empty list of offers.
2. If we're aiming for _N_ offers, grab _M_ = _N_ × _X_ potentials from the top of the queue. Maybe _X_ = 2 or _X_ = 10, it's not super important.
3. From those _M_, filter by test offer status in REDCap and append to the list of offers.
5. If the number of offers is less than _N_, start again at step 2 with the remainder as _N_.
6. Finally, if the number of offers is greater than the total desired, truncate the list.

We'd tune _X_ so that most of the time only one fetch to REDCap is needed.

schema/deploy/operations/test-quota.sql
```diff
@@ -0,0 +1,40 @@
...
+create table operations.test_quota (
+    name text not null,
+    timespan tstzrange not null,
+
+    /* Using a reserved word is slightly annoying because we'll have to quote
+     * it everywhere… but I dunno, the other words I can think of aren't quite
+     * as descriptive.  Going with it for now!
+     *   -trs, 17 Sept 2020
+     */
+    "limit" integer check ("limit" >= 0) not null,
```
kschwabefry, line 16: Brainstorming - how about replacing limit with maximum, upper bound, or quota? I'm not a fan of reserved words as column names either

lib/seattleflu/id3c/cli/command/offer_uw_testing.py
```diff
@@ -0,0 +1,232 @@
...
+    # keep more metadata about that.  These same benefits could be realized by
+    # turning our normal logging output into structured event logs.  I think
+    # that's preferrable, so decided not to implement as a ledger right now.
+    #   -trs, 17 Sept & 13 Oct 2020
+
+    # Lookup the quota for the current time, locking it for update at the end
+    # after we make offers.
+    #
+    # XXX TODO: As a future improvement, automatically pick up any remaining
+    # quota left from _past_ timespans in the current day.
```
kschwabefry, line 75: +1 This would be nice to have operationally

kschwabefry (approved):

tsibley:
I added the grants to `uw-priority-queue-processor` in the latest commit.

lib/seattleflu/id3c/cli/command/offer_uw_testing.py
tsibley: Agreed. I left it out because it's a bit nuanced to get right and additional complexity right now that we probably don't need (assuming that our queue is never close to empty and/or our quota for a timespan is never very large).

schema/deploy/operations/test-quota.sql
tsibley: Fair! I'd love better options. "Limit" makes most sense to me and is used in other quota systems I've used, but reserved words are a little awkward (but also we're unlikely to be manually using them much).

I didn't use `quota` because "quota" conceptually encompasses the whole row/set of rows for me, not just the limit/max/upper bound, and this is why "quota" is in the table name.

I like `maximum` or `max`. That makes sense as a more specific "limit", since limit can be either a minimum or a maximum.

tsibley: Renamed to `max` in the latest repush.

tsibley:
Ran into two catches during testing:

1. For baseline/surveillance, we need to create a new `encounter_arm_1` instance when updating `testing_trigger` because the queue-reported REDCap event is `enrollment_arm_1` (and there's no repeat instance).

2. Type checks were failing.

Both addressed in the [latest repush](https://github.com/seattleflu/id3c-customizations/compare/176c8f079136fadef0d6d8522f90b5f4d6fd30e1..7f23368df6f6c559e72641749abb80cc606d8541).

joverlee521:
Latest push includes updates to the priority queue view, details in commit messages.

jccraft-uw (approved):
@joverlee521 The changes in the two commits look great. You're excluding PTs who say they've tested positive in the past 7 days, and you updated the order of priorities. 

Would it make things (in these views and troubleshooting outside the views) easier if we made some of the CTEs (like all_uw_instances, uw_encounters, and uw_enrollments) actual persisted views?

joverlee521:
@jccraft-uw  From your commit message it seems this change is mostly useful for the Metabase dashboard? Just wondering if we can add the `row_number()` when writing the Metabase query instead of adding it to the operational query. 

When I look at the new query plan, there's an additional Sort for the `row_number()` window function. However, the query run time doesn't change much, so I'm also okay with keeping it in the operational query.

jccraft-uw:
> @jccraft-uw From your commit message it seems this change is mostly useful for the Metabase dashboard? Just wondering if we can add the `row_number()` when writing the Metabase query instead of adding it to the operational query.
> 
> When I look at the new query plan, there's an additional Sort for the `row_number()` window function. However, the query run time doesn't change much, so I'm also okay with keeping it in the operational query.

Good point, @joverlee521 I can just add the row_number() in the Metabase query. I've removed that commit.

jccraft-uw:
@joverlee521 are you good with this commit I just added https://github.com/seattleflu/id3c-customizations/pull/165/commits/3783f6f21e24174aaa9cf7e5efaaa540f75c98b8
In there I grant select on operations tables to the reporter role, and I updated the timestamps in the sqitch plan.

joverlee521:
> @joverlee521 are you good with this commit I just added [3783f6f](https://github.com/seattleflu/id3c-customizations/commit/3783f6f21e24174aaa9cf7e5efaaa540f75c98b8)
> In there I grant select on operations tables to the reporter role, and I updated the timestamps in the sqitch plan.

I thought we didn't need to update the reporter grants because it has [default privileges](https://github.com/seattleflu/id3c/blob/master/schema/deploy/roles/reporter.sql#L15-L16). When I deploy the sqitch changes locally (without your latest commit), the `reporter` role has select access to `operations.test_quota`.

jccraft-uw:
> > @joverlee521 are you good with this commit I just added [3783f6f](https://github.com/seattleflu/id3c-customizations/commit/3783f6f21e24174aaa9cf7e5efaaa540f75c98b8)
> > In there I grant select on operations tables to the reporter role, and I updated the timestamps in the sqitch plan.
> 
> I thought we didn't need to update the reporter grants because it has [default privileges](https://github.com/seattleflu/id3c/blob/master/schema/deploy/roles/reporter.sql#L15-L16). When I deploy the sqitch changes locally (without your latest commit), the `reporter` role has select access to `operations.test_quota`.

Without my latest commit, when I try locally the reporter role does not have access to `operations.test_quota`. 
I did see this
`alter default privileges grant usage on schemas to reporter;` 
and wondered why that doesn't work. Then I found this:
https://www.postgresql.org/docs/9.4/sql-alterdefaultprivileges.html
"ALTER DEFAULT PRIVILEGES allows you to set the privileges that will be applied to objects created in the future. (It does not affect privileges assigned to already-existing objects.) Currently, only the privileges for tables (including views and foreign tables), sequences, functions, and types (including domains) can be altered."
It does not mention schema.

joverlee521:
> Without my latest commit, when I try locally the reporter role does not have access to `operations.test_quota`.
> I did see this
> `alter default privileges grant usage on schemas to reporter;`
> and wondered why that doesn't work. Then I found this:
> https://www.postgresql.org/docs/9.4/sql-alterdefaultprivileges.html
> "ALTER DEFAULT PRIVILEGES allows you to set the privileges that will be applied to objects created in the future. (It does not affect privileges assigned to already-existing objects.) Currently, only the privileges for tables (including views and foreign tables), sequences, functions, and types (including domains) can be altered."
> It does not mention schema.

The production database is using [version 10](https://www.postgresql.org/docs/10/sql-alterdefaultprivileges.html) which does include schema. 
When I deploy locally with `PGUSER=postgres`, the reporter role has the correct privileges, but it does _not_ have access when I deploy with my default superuser. I assume this is because the initial "alter default privileges" command was deployed using the postgres user?

jccraft-uw:
> > Without my latest commit, when I try locally the reporter role does not have access to `operations.test_quota`.
> > I did see this
> > `alter default privileges grant usage on schemas to reporter;`
> > and wondered why that doesn't work. Then I found this:
> > https://www.postgresql.org/docs/9.4/sql-alterdefaultprivileges.html
> > "ALTER DEFAULT PRIVILEGES allows you to set the privileges that will be applied to objects created in the future. (It does not affect privileges assigned to already-existing objects.) Currently, only the privileges for tables (including views and foreign tables), sequences, functions, and types (including domains) can be altered."
> > It does not mention schema.
> 
> The production database is using [version 10](https://www.postgresql.org/docs/10/sql-alterdefaultprivileges.html) which does include schema.
> When I deploy locally with `PGUSER=postgres`, the reporter role has the correct privileges, but it does _not_ have access when I deploy with my default superuser. I assume this is because the initial "alter default privileges" command was deployed using the postgres user?

OK doke. Yep, when I deploy locally as postgres it's there. Go figure. I've removed that commit and added one that just changes the timestamps of in the sqitch.plan. I'll tag before merging. Thanks for all that.

tsibley:
> I've removed that commit and added one that just changes the timestamps of in the sqitch.plan.

Why update the timestamps?

jccraft-uw:
> > I've removed that commit and added one that just changes the timestamps of in the sqitch.plan.
> 
> Why update the timestamps?

The team has mentioned that they want to keep sqitch entries aligned with git commits. I updated the timestamps so that these entries at the bottom of the file are newer than the ones above it.
Just trying to keep everybody happy.

tsibley:
@jccraft-uw Got it, no worries. I don't personally agree with "the team" on the necessity or usefulness of that, but that's ok! :-) Thanks for paying attention to the details.

tsibley:
@jccraft-uw I've check off the "End to end testing with REDCap" item in the top post since it's my understanding you did that last week or the week prior.

Will you manage Trello cards for the other deploy items in the list up top? (if any cards are needed)

jccraft-uw:
> @jccraft-uw I've check off the "End to end testing with REDCap" item in the top post since it's my understanding you did that last week or the week prior.
> 
> Will you manage Trello cards for the other deploy items in the list up top? (if any cards are needed)

Will do, Tom! Thanks a lot.
Yes, I did the REDCap testing, so thanks for checking that off.

joverlee521:
> > > I've removed that commit and added one that just changes the timestamps of in the sqitch.plan.
> > 
> > 
> > Why update the timestamps?
> 
> The team has mentioned that they want to keep sqitch entries aligned with git commits. I updated the timestamps so that these entries at the bottom of the file are newer than the ones above it.
> Just trying to keep everybody happy.

Ah this might have been me, sorry if I mislead you. I think we want to keep the sqitch changes and github commits in the same order, but I never paid attention to the timestamps in the sqitch plan.

jccraft-uw:
Deployed to backoffice.

PR #166, opened 2020-11-05
etl redcap-det: Ignore float values which are non-finite (NaN and ±Inf)

tsibley:
Generalize non-finite float handling and remove special-casing. Details in commits.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap.py
@@ -3,6 +3,7 @@ Functions shared by REDCap DET ETL
 """
 from datetime import datetime
 from enum import Enum
+from math import isfinite
 import re
 from typing import Dict, List, Mapping, Match, Optional, Tuple, Union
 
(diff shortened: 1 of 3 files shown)


lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -611,10 +612,12 @@ def cast_to_integer(string: str) -> Optional[int]:
 
     def cast_to_float(string: str) -> Optional[float]:
         try:
-            return float(string)
+            n = float(string)
         except ValueError:
             return None
 
+        return n if isfinite(n) else None
```
jccraft-uw, line 619: I debated whether to put this logic into this utility but didn't like the idea that the caller wouldn't know that it was happening. It's possible that the float value is expected to always have a legit value and the caller would want to know that it sometimes doesn't--for example, we might need to fix the REDCap calculation.

tsibley: Ah, hmm. For the future, that sort of reasoning and context is exactly the sort of thing I think is really valuable to include in commit messages (e.g. in e6a089d in this case).

I agree it's possible the caller might want to be able to know that a non-finite value was received… but that seems at odds with the current `cast_to_float` behaviour of converting all `ValueError`s to `None`.

jccraft-uw (approved):

PR #167, opened 2020-11-20
Accommodate unscheduled event in childcare ETL

jccraft-uw:
The unscheduled event has fewer instruments than a regular
scheduled encounter event but is similar enough that this ETL job can
treat it just like a scheduled encounter event.


joverlee521 (approved):
I didn't test locally, but the code change looks good to me. 

Small reminder to re-process the skipped records that currently have an unscheduled event, probably by manually generating DETs for those records.

jccraft-uw:
Thanks, @joverlee521  !
Deployed, and I generated and imported the DETs for record 20.

PR #168, opened 2020-11-24
etl redcap-det scan: Fix error when the zip code field is empty

joverlee521:
Update :func: `zipcode_map` to check if the provided *redcap_code* is an
emptry string to catch the edge case of when the zipcode field is empty.

This field may be an empty string if edited by internal SFS members
for participants that live outside the SCAN coverage zip codes. We've
been doing in person dropoffs for priority codes that can be outside of
the normal SCAN coverage.

There's no need to update the geocoding process because SmartyStreets
can handle address lookups without a zipcode.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -305,12 +305,18 @@ def locations(db: DatabaseSession, cache: TTLCache, record: dict) -> list:
     return [tract_entry, address_entry]
 
 
-def zipcode_map(redcap_code: str) -> str:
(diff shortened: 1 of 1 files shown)


jccraft-uw (approved):

jccraft-uw:
This simple change makes sense, @joverlee521 . I didn't test it locally to know if everything downstream works as it should, however.

kairstenfay (approved):

joverlee521:
Latest commit also removes the `UnknownRedcapZipCode` error and replaces it with a warning so our ETL doesn't get stopped by unknown zip codes.

jccraft-uw (approved):

joverlee521:
Deployed.

PR #170, opened 2020-12-04
Lower log level when no quota exists for uw offers

jccraft-uw:
We expect the job to run all the time, but we will purposefully not
configure quotas for hours outside "waking" hours as not to annoy
participants with alerts from REDCap. This change lowers the log
level from WARNING to INFO for two cases when the job doesn't find
a quota to work with.


jccraft-uw:
Deployed.

PR #171, opened 2020-12-07
Create new versions of our modeling views

kairstenfay:
Create a new version of the incidence modeling view and observation with presence absence modeling view.

Also refactor many of our existing hCoV-19 presence-absence CTE queries into a new view for reduced duplication across our view definitions.

~I'm waiting on one confirmation from Roy as to whether he wants all samples _and_ all encounters in these new views, or just all samples.~

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -40,6 +40,7 @@ drop view if exists shipping.flu_assembly_jobs_v1;
 
 drop view if exists shipping.scan_follow_up_encounters_v1;
 drop materialized view if exists shipping.scan_encounters_v1;
+drop view if exists shipping.hcov19_presence_absence_result_v1;
 drop view if exists shipping.hcov19_observation_v1;
 
 drop view if exists shipping.observation_with_presence_absence_result_v2;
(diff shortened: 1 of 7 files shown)


schema/deploy/shipping/views.sql
```diff
@@ -2345,39 +2362,20 @@ grant select
...
+            when present is null then 'positive'
+          end as hcov19_result,
+          presence_absence_id,
+          hcov19_result_release_date
         from
-            warehouse.presence_absence as pa
-            join warehouse.target using (target_id)
-            join warehouse.organism using (organism_id)
-            join warehouse.sample using (sample_id)
+          shipping.hcov19_presence_absence_result_v1 as pa
```
kairstenfay, line 2532: Note to self: rename this alias to `hcov19_pa` to be consistent with other views and the commit message.

kairstenfay:
[Create new shipping view for IDM with sample + encounter + p/a result data](https://trello.com/c/PX3HmJKb/871-create-new-shipping-view-for-idm-with-sample-encounter-p-a-result-data)

kairstenfay:
Roy confirmed he wants all samples, not all samples + all encounters, in these views.

schema/deploy/shipping/views.sql
```diff
@@ -2161,16 +2204,18 @@ create or replace view shipping.return_results_v3 as
...
         end as status_code,
         result_ts,
         swab_type,
         staff_observed
     from
       samples
-      left join hcov19_presence_absence using (sample_id)
+      left join shipping.hcov19_presence_absence_result_v1 as hcov19_pa using (sample_id)
+    where
+      hcov19_pa.details is null or hcov19_pa.details @> '{"assay_type": "Clia"}'
```
tsibley, line 2375: I think this subtly changes the rows retrieved by the join.

In the existing version, rows with a null `details` would be omitted because the `assay_type` check (in the `hcov19_presence_absence` CTE) would evaluate to null.

In the new version, they're included because of the `is null` clause here.

I think the intent was to not require the `assay_type` check if the left join produced no rows. In that case, the correct condition is to check if the join key is null, e.g.:

```sql
hcov19_pa.sample_id is null
```

schema/deploy/shipping/views.sql
```diff
@@ -2274,16 +2289,18 @@ create or replace view shipping.scan_return_results_v1 as
...
         end as status_code,
         result_ts,
         swab_type,
         staff_observed
     from
       scan_samples
-      left join hcov19_presence_absence using (sample_id)
+      left join shipping.hcov19_presence_absence_result_v1 as hcov19_pa using (sample_id)
+    where
+      hcov19_pa.details is null or hcov19_pa.details @> '{"assay_type": "Clia"}'
```
tsibley, line 2460: As above re: what rows are included.

schema/deploy/shipping/views.sql
```diff
@@ -2485,8 +2452,9 @@ create or replace view shipping.scan_hcov19_result_counts_v1 as
           end as county
 
       from shipping.scan_encounters_v1
-      join hcov19_presence_absence using (sample)
-
+      join shipping.hcov19_presence_absence_result_v1 as hcov19_pa using (sample_id)
+      where
+        hcov19_pa.details is null or hcov19_pa.details @> '{"assay_type": "Clia"}'
```
tsibley, line 2614: As above re: what rows are included.

tsibley (changes requested):

schema/deploy/shipping/views.sql
kairstenfay: Thanks for catching that, Tom! I just re-pushed with `hcov19_pa.sample_id is null` replacing `hcov19_pa.details` is null. Please take a look when you get a chance to make sure I understood your feedback. 

A quick count of total rows returned in each view shows no regressions. I am going to continue doing a bit more testing, though, to be safe.

tsibley: Yep, that's what I meant. Thanks!

For testing things like this, which are intended to be non-changes, I like to dump the whole darn view to a file before and after, and then diff them to confirm no changes with existing data.

schema/deploy/shipping/views.sql
kschwabefry: Bump on the note to self :) Reading your commit message, I think hcov19_pa is a good rename! Thanks for documenting that.

schema/deploy/shipping/views.sql
kairstenfay: Ah, that's the most robust testing strategy of all! I thought to do that at one point but apparently forgot.

schema/deploy/shipping/views.sql
kairstenfay: Thanks!

schema/deploy/shipping/views.sql
kairstenfay: I just did a `diff` of the three shipping views that use the refactored hcov19 p/a view, and there were no changes :+1:

tsibley (approved):

PR #172, opened 2020-12-09
Compute encounter date from REDCap repeat instance

jccraft-uw:
For the UW reopening ETL, accommodate an issue when REDCap
fails to set the `attestation_date` on the `daily_attestation`
instrument. When this case occurs, compute the encounter date
from the repeat instance. This is safe to do because the Musher
computes the instance from the date.

This is where Musher computes instance ID from the date: 
https://github.com/seattleflu/husky-musher/blob/35c1e7fd14cf12065e4bdc86626116053c6bd056/lib/husky_musher/utils/redcap.py#L183

I tested with this record from HCT. Right now it's a good one to test with because it's in the situation where attestation_date didn't get set.
 id3c redcap-det generate --project-id=23854 "1818"

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -50,6 +50,7 @@ INTERNAL_SYSTEM = "https://seattleflu.org"
 ENROLLMENT_EVENT_NAME = "enrollment_arm_1"
 ENCOUNTER_EVENT_NAME = "encounter_arm_1"
 SWAB_AND_SEND_SITE = 'UWReopeningSwabNSend'
+STUDY_START_DATE = datetime(2020, 9, 24) # Study start date of 2020-09-24
 
 REQUIRED_ENROLLMENT_INSTRUMENTS = [
     'eligibility_screening',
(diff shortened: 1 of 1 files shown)


lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -372,6 +373,13 @@ def get_encounter_date(record: dict, event_type: EventType) -> Optional[str]:
         elif record.get('testing_date'): # from the 'Testing Determination - Internal' instrument
             encounter_date = record.get('testing_date')
 
+        # We have seen cases when the `attestation_date` is not getting set
+        # by REDCap in the `daily_attesation` instrument. Here, we get the date
+        # based on the instance ID of the `daily_attestation` instrument. It's safe
+        # to do this because the Musher computes the instance from the date.
+        if encounter_date == None and is_complete('daily_attestation', record) and record.repeat_instance:
```
tsibley, line 380: Minor nit, but it stood out to me: `is None` is widely preferred to `== None` within Python conventions/best practices. The two comparisons are subtly different, and `is None` is more precise for what is generally desired.

This is not merge-blocking at all, as in this case I believe it has no impact, but wanted to note it.

tsibley (approved):
LGTM by inspection. I didn't run the ETL routine myself, but it'd be good if at least one of us did before deploy. I assume you did to test it?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: Sorry about that. This is from me spending the day in SQL yesterday and switching back to python. I've made the change to `is`.

jccraft-uw: And I added the DET info to the description at the top of the PR. That's the record I used for testing.

jccraft-uw:
Deployed.

PR #173, opened 2020-12-22
Add `yakima` and `pierce` variables for SCAN

jccraft-uw:
To help filter data, ingest the `yakima` and `pierce` variables
from REDCap and include them in key views.

Bump the redcap_det_scan revision number in order to pull these
variables in for all records.


kschwabefry (approved):
Code changes look good to me. 
Wanted to talk through my redcap thinking here. Looking at the redcap and thinking about the conversation last week with that team, I figure these fields are calculated client-side, which makes them a little less reliable. But, the risk of inconsistencies in that field may still be lower than the alternate cost of maintaining a separate list of PUMAs here in id3c .

jccraft-uw:
> Code changes look good to me.
> Wanted to talk through my redcap thinking here. Looking at the redcap and thinking about the conversation last week with that team, I figure these fields are calculated client-side, which makes them a little less reliable. But, the risk of inconsistencies in that field may still be lower than the alternate cost of maintaining a separate list of PUMAs here in id3c .

Good point. For what it's worth, the REDCap reports that are used to report on Pierce County use those variables too.

joverlee521:
> the risk of inconsistencies in that field may still be lower than the alternate cost of maintaining a separate list of PUMAs here in id3c .

Instead of maintaining a list of PUMAs in ID3C, could we add "county" to the `warehouse.location.hierarchy` column?

jccraft-uw:
Deployed.

PR #174, opened 2020-12-29
Materialize UW Reopening encounter view

jccraft-uw:
To speed up the __uw_priority_queue_v1 view, materialize UW encounters.

Here are times running against the production DB:
__uw_priority_queue_v1 as it is in Master: 53 secs 501 msec
__uw_priority_queue_v1 as it is in this commit: 2 secs 363 msec
refresh the __uw_encounters materialized view: 19 secs 863 msec

For testing, I diff'ed sorted output from the master version of
__uw_priority_queue_v1 to that of the new version of the view. The
output was the same.

This needs a corresponding Backoffice change to create the cron job
to refresh the materialized view. That is seattleflu/backoffice#112

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -23,6 +23,7 @@ drop view if exists shipping.scan_demographics_v1;
 
 drop view if exists shipping.uw_priority_queue_v1;
 drop view if exists shipping.__uw_priority_queue_v1;
+drop materialized view if exists shipping.__uw_encounters;
 drop view if exists shipping.uw_reopening_ehs_reporting_v1;
 drop view if exists shipping.uw_reopening_encounters_v1;
 drop view if exists shipping.uw_reopening_enrollment_fhir_encounter_details_v1;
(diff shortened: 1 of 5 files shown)


jccraft-uw:
[Optimize shipping.__uw_priority_queue_v1 view](https://trello.com/c/zxwHi3jy/856-optimize-shippinguwpriorityqueuev1-view)

schema/deploy/shipping/views.sql
```diff
@@ -23,6 +23,7 @@ drop view if exists shipping.scan_demographics_v1;
 
 drop view if exists shipping.uw_priority_queue_v1;
 drop view if exists shipping.__uw_priority_queue_v1;
+drop materialized view if exists shipping.__uw_encounters;
```
joverlee521, line 26: Also need to include this drop statement in the revert script.

jccraft-uw: Sorry about that. I added the drop to schema/revert/shipping/views.sql.

joverlee521 (changes requested):
Thanks for working through this @jccraft-uw! Just a couple small things:
1. Please add the new materialized view to the verify script
2. This isn't very well documented within ID3C (not sure where we should document this), but our [refresh materialized view command](https://github.com/seattleflu/id3c/blob/master/schema/deploy/roles/materialized-view-refresher/grants.sql#L23) uses the [CONCURRENTLY](https://www.postgresql.org/docs/10/sql-refreshmaterializedview.html#Parameters) parameter. This can only be done if the materialized view has at least one unique index.

jccraft-uw:
> Thanks for working through this @jccraft-uw! Just a couple small things:
> 
> 1. Please add the new materialized view to the verify script
> 2. This isn't very well documented within ID3C (not sure where we should document this), but our [refresh materialized view command](https://github.com/seattleflu/id3c/blob/master/schema/deploy/roles/materialized-view-refresher/grants.sql#L23) uses the [CONCURRENTLY](https://www.postgresql.org/docs/10/sql-refreshmaterializedview.html#Parameters) parameter. This can only be done if the materialized view has at least one unique index.

1. I added the materialized view to the verify script. For this PR I failed to remember the cost of adding a new view .. revert, verify.
2. Oh this got interesting. For an "encounters" view, there should be one row per encounter_id, so easy, that should be the unique index. But it turns out that we have 9 encounters that join to more than one sample. For this, I ended up taking the min collection date.

This line: `sample.collected::date as sample_collection_date`
Became: `(select min(collected)::date from warehouse.sample where encounter_id = encounter.encounter_id group by encounter_id) as sample_collection_date`

We will get a duplicate key violation the next time we see that an encounter_id belongs to more than 1 individual. Because that is a data quality issue that we should clean up right away, I decided to keep that risk in here instead of indexing on encounter_id + individual.

joverlee521:
> 2\. Oh this got interesting. For an "encounters" view, there should be one row per encounter_id, so easy, that should be the unique index. But it turns out that we have 9 encounters that join to more than one sample. For this, I ended up taking the min collection date.
> 
> This line: `sample.collected::date as sample_collection_date`
> Became: `(select min(collected)::date from warehouse.sample where encounter_id = encounter.encounter_id group by encounter_id) as sample_collection_date`
> 
> We will get a duplicate key violation the next time we see that an encounter_id belongs to more than 1 individual. Because that is a data quality issue that we should clean up right away, I decided to keep that risk in here instead of indexing on encounter_id + individual.

Oh. Interesting. I wonder if these are user errors where the kiosk instrument got entered in the wrong instance....
I think we care about the latest collection date, so `max(collected)::date`?

jccraft-uw:
> > 2. Oh this got interesting. For an "encounters" view, there should be one row per encounter_id, so easy, that should be the unique index. But it turns out that we have 9 encounters that join to more than one sample. For this, I ended up taking the min collection date.
> > This line: `sample.collected::date as sample_collection_date`
> > Became: `(select min(collected)::date from warehouse.sample where encounter_id = encounter.encounter_id group by encounter_id) as sample_collection_date`
> > We will get a duplicate key violation the next time we see that an encounter_id belongs to more than 1 individual. Because that is a data quality issue that we should clean up right away, I decided to keep that risk in here instead of indexing on encounter_id + individual.
> 
> Oh. Interesting. I wonder if these are user errors where the kiosk instrument got entered in the wrong instance....
> I think we care about the latest collection date, so `max(collected)::date`?

Max makes more sense. The most recent time the PT was tested.

jccraft-uw:
@joverlee521 I just added a second commit for decreasing the time required between samples/invites. The group wants this change out as soon as we can, so I included it in this branch. 

If you're OK with this new commit and my changes from the first commit please approve. I will then sqitch tag and deploy.

joverlee521 (approved):

jccraft-uw:
Deployed.

PR #175, opened 2020-12-30
UW Reopening ETL changes for Winter

jccraft-uw:
This PR:
1. Brings in new fields added to the enrollment questionnaire.
2. Removes asserts/LOG.error (replaces the meaningful ones with LOG.warning) so that the job doesn't stop when we see an odd record.

I'm not bumping the revision # on this one. I'll do that after Jessica sets values for the new fields.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -470,7 +470,8 @@ def create_enrollment_questionnaire_response(record: REDCapRecord, patient_refer
         'pronouns',
         'on_campus_freq',
         'vaccine_method',
-        'vaccine_where'
+        'vaccine_where',
+        'uw_housing_group',
     ]
 
     date_questions = [
@@ -502,7 +503,9 @@ def create_enrollment_questionnaire_response(record: REDCapRecord, patient_refer
         'travel_states_phs_base',
         'swab_and_send_calc',
         'kiosk_calc',
-        'covid_test_week_base'
+        'covid_test_week_base',
+        'uw_housing_resident',
+        'on_campus_2x_week',
     ]
 
     decimal_questions = [
@@ -552,6 +555,25 @@ def create_enrollment_questionnaire_response(record: REDCapRecord, patient_refer
     record['tier'] = tier
 
     vaccine_item = create_vaccine_item(record["vaccine"], record['vaccine_year'], record['vaccine_month'], 'dont_know')
+
+    # Set the UW housing group
+    housing_group = None
+    if record.get('uw_housing_group_a') == '1':
+        housing_group = 'a'
+    elif record.get('uw_housing_group_b') == '1':
+        housing_group = 'b'
+    elif record.get('uw_housing_group_c') == '1':
+        housing_group = 'c'
+    elif record.get('uw_housing_group_d') == '1':
+        housing_group = 'd'
+    elif record.get('uw_housing_group_e') == '1':
+        housing_group = 'e'
+    elif record.get('uw_housing_group_f') == '1':
+        housing_group = 'f'
+    elif record.get('uw_housing_group_g') == '1':
+        housing_group = 'g'
+    record['uw_housing_group'] = housing_group
+
 
     return create_questionnaire_response(
         record = record,
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -85,12 +85,15 @@ def command_for_each_project(function):
 def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
     redcap_record_instances: List[REDCapRecord]) -> Optional[dict]:
 
-    assert redcap_record_instances is not None and len(redcap_record_instances) > 0, \
-        "The redcap_record_instances list was not populated."
+    if redcap_record_instances is None or len(redcap_record_instances) == 0:
+        LOG.warning(f"There are no record instances. Skipping record.")
+        return None
 
     enrollments = [record for record in redcap_record_instances if record.event_name == ENROLLMENT_EVENT_NAME]
-    assert len(enrollments) == 1, \
-        f"Record had {len(enrollments)} enrollments."
+
+    if not len(enrollments) == 1:
+        LOG.warning(f"There are {len(enrollments)} enrollment instances. Skipping record.")
+        return None
 
     enrollment = enrollments[0]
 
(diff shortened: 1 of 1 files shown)


lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -502,7 +502,16 @@ def create_enrollment_questionnaire_response(record: REDCapRecord, patient_refer
...
-        'covid_test_week_base'
+        'covid_test_week_base',
+        'uw_housing_resident',
+        'uw_housing_group_a',
+        'uw_housing_group_b',
+        'uw_housing_group_c',
+        'uw_housing_group_d',
+        'uw_housing_group_e',
+        'uw_housing_group_f',
+        'uw_housing_group_g',
```
joverlee521, line 512: Can we just consolidate these into one field `uw_housing_group` that then has a value of a-g?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -85,12 +85,11 @@ def command_for_each_project(function):
 def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
     redcap_record_instances: List[REDCapRecord]) -> Optional[dict]:
 
-    assert redcap_record_instances is not None and len(redcap_record_instances) > 0, \
-        "The redcap_record_instances list was not populated."
-
```
joverlee521, line 90: Do we need to log a warning and skip this record to replace this assert?

jccraft-uw: I thought about that, but since we're already ensuring that we have 1 enrollment instance that means we are already covering the len > 0 case.
Do we need to cover the None case?
I don't think I needed this assert from the start.

joverlee521: Hmmm, the list comprehension to create `enrollments` will error out if we get `redcap_record_instances` that is None.

jccraft-uw: I didn't know if would ever be possible for the instances to be None, but I added that check/warning just in case.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: That does make things nicer. I checked with Jessica to see if we could have one field in REDCap, but the calculation logic doesn't seem easily doable, so we're keeping multiple group fields in REDCap, and I've combined them in the ETL. Thanks for that suggestion.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -553,6 +559,25 @@ def create_enrollment_questionnaire_response(record: REDCapRecord, patient_refer
...
+        housing_group = 'c'
+    elif record.get('uw_housing_group_d') == '1':
+        housing_group = 'd'
+    elif record.get('uw_housing_group_e') == '1':
+        housing_group = 'e'
+    elif record.get('uw_housing_group_f') == '1':
+        housing_group = 'f'
+    elif record.get('uw_housing_group_g') == '1':
+        housing_group = 'g'
+    record['uw_housing_group'] = housing_group
```
joverlee521, line 578: Two questions about this: 
1. Do we care if more than one group equals "1" ? 
2. Do we care if `uw_housing_resident` is True, but none of these groups have calculated to "1"? 

These are QC things that would be nice to know about through warning messages?

jccraft-uw: Ah, an interesting topic: QC, who does it and where is it done?
My take is not to bake much QC into the ETL job. It should warn about just the pieces that matter to it (an unexpected event name, for example). It should warn about things that the Dev team should drop everything and investigate.
For the housing questions, REDCap reports could find those problems. It's painful to run the reports because the project is big, but I don't think that justifies adding more QC to the Dev team.
This is a really good question for me and Kristen to ask "the BBI". I could see doing research QC as a valuable part of our workload.
I'm calling this out to @seattleflu/id3c-devs to raise this as something we should talk about.

joverlee521: Yup, QC is always an interesting topic 😅 

I'm asking in this case because I don't have context on how these fields will be used downstream of the ETL. If these are important in the quota system or reporting to EHS, it would be good to know about these problems earlier.

jccraft-uw: These fields will be important for test offering, so they are important. I agree that it's good to know about issues with these, but that could be done in a REDCap report instead of a warning in id3c-alerts. But this would require someone running the report once in a while to check.

joverlee521: With this if/elif statement, we are giving "preferential" selection to group A. Could this accidentally inflate the count of people in group A that then messes up the priority queue? (Asking because I'm not sure how this variable will affect the priority queue)

jccraft-uw: The calculation rules for the uw_housing_group_a through _g fields put a participant who lives in UW housing into one of the 7 housing groups based on the PT's REDCap record number. The record ranges for the 7 fields' logic don't overlap, so a PT should not be in more than one group.

The priority queue will invite the PTs for testing when it is their group's day ... for example, invite group "c" PTs on Wednesdays.

kairstenfay: Assuming there's only one possible housing group with a value of '1', a more concise way to write this logic would be:

```
groups = 'abcdefg'
housing_group = next( g for g in groups if record.get(f"housing_group_{group}") == '1')
```

jccraft-uw: That's a really cool use of string's iterator! 
Not every participant will be in a uw_housing_group. It's only PTs who live on campus (apartments or dorms) that will get assigned to one of the uw_housing_groups.
When I try out this scenario--the PT having 0 for all the 7 housing groups--this throws an error when it gets to the end of the groups string (it throws on the iteration after g = 'g'). What we would want for this scenario is for housing_group to be None.
Is there a way to make that handle that case?

kairstenfay: Yep, you use `next()` in a try/except block and catch a `StopIteration` error to handle the special case.

jccraft-uw: Gotcha. Because the majority of participants won't have a '1' for any of the uw_housing_group_ fields, we would hit the except block the majority of the time. Given that, I'm not going to go with the next() approach for this logic.

jccraft-uw:
Deployed.

PR #176, opened 2021-01-06
Uw reopening system dates

jccraft-uw:
Favor REDCap survey timestamps over custom date fields when available when determining the encounter date or the sample collection date.
For sample collection date, do not use a value that the participants writes on the tube if that value is in the future.

I did not bump the ETL revision number in this PR. We're going to need to process all DETs soon once some new REDCap calculations get backfilled.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap.py
@@ -688,3 +688,17 @@ def create_questionnaire_response(record: REDCapRecord, question_categories: Dic
         return create_resource_entry(questionnaire_reseponse_resource, full_url)
 
     return None
+
+
+def extract_date_from_survey_timestamp(record: REDCapRecord, survey_name: str) -> Optional[str]:
+    """
+    Extracts as a string the date component of the *survey_name* REDCap survey timestamp, the system
+    timestamp that is captured automatically and is not dependent on the client. This timestamp
+    is in local (Pacific) time. The timestamp will be populated only if the instrument was filled out
+    as a survey. The timestamp field cannot be set via a REDCap data import.
+    """
+    if record and survey_name and record.get(f'{survey_name}_timestamp'):
+        return datetime.strptime(record.get(f'{survey_name}_timestamp'),
+            '%Y-%m-%d %H:%M:%S').strftime('%Y-%m-%d')
+
+    return None
--- a/lib/seattleflu/id3c/cli/command/etl/redcap.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap.py
@@ -14,7 +14,7 @@ from .fhir import *
 from .redcap_map import map_sex, map_symptom, UnknownVaccineResponseError
 from id3c.cli.command.geocode import get_geocoded_address
 from id3c.cli.command.location import location_lookup
-from id3c.cli.redcap import Record as REDCapRecord
+from id3c.cli.redcap import is_complete, Record as REDCapRecord
 from id3c.db.session import DatabaseSession
 import logging
 
@@ -697,7 +697,7 @@ def extract_date_from_survey_timestamp(record: REDCapRecord, survey_name: str) -
     is in local (Pacific) time. The timestamp will be populated only if the instrument was filled out
     as a survey. The timestamp field cannot be set via a REDCap data import.
     """
-    if record and survey_name and record.get(f'{survey_name}_timestamp'):
+    if record and survey_name and is_complete(survey_name, record) and record.get(f'{survey_name}_timestamp'):
         return datetime.strptime(record.get(f'{survey_name}_timestamp'),
             '%Y-%m-%d %H:%M:%S').strftime('%Y-%m-%d')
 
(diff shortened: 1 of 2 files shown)


jccraft-uw:
[uw-reopening ETL : Don't use dates if they are in the future](https://trello.com/c/cDpXbwIb/956-uw-reopening-etl-dont-use-dates-if-they-are-in-the-future)

lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -688,3 +688,16 @@ def build_questionnaire_items(question: str) -> Optional[dict]:
...
 
     return None
+
+
+def extract_date_from_survey_timestamp(record: REDCapRecord, survey_name: str) -> Optional[str]:
+    """
+    Extracts as a string the date component of the *survey_name* REDCap survey timestamp, the system
+    timestamp that is captured automatically and is not dependent on the client. This timestamp
+    is in local (Pacific) time.
+    """
```
kschwabefry, line 698: It might be worth to add a comment to highlight the null edge case with instrument timestamps that Redcap notes at the end of the api documentation entry,

(From https://redcap.iths.org/redcap_v10.5.2/API/help.php?content=exp_records )

exportSurveyFields:
true, false [default] - specifies whether or not to export the survey identifier field (e.g., 'redcap_survey_identifier') or survey timestamp fields (e.g., instrument+'_timestamp') when surveys are utilized in the project. If you do not pass in this flag, it will default to 'false'. If set to 'true', it will return the redcap_survey_identifier field and also the survey timestamp field for a particular survey when at least one field from that survey is being exported. **NOTE: If the survey identifier field or survey timestamp fields are imported via API data import, they will simply be ignored since they are not real fields in the project but rather are pseudo-fields.**

lib/seattleflu/id3c/cli/command/etl/redcap.py
```diff
@@ -688,3 +688,16 @@ def build_questionnaire_items(question: str) -> Optional[dict]:
...
+    if record and survey_name and record.get(f'{survey_name}_timestamp'):
+            return datetime.strptime(record.get(f'{survey_name}_timestamp'),
+                '%Y-%m-%d %H:%M:%S').strftime('%Y-%m-%d')
+    else:
+        return None
```
kairstenfay, line 703: A couple of minor style suggestions.

The `return` statement on line 700 is over-indented.

Because we `return` in the `if` statement, we don't need to write an `else` statement. You can write, `return None` after the `if` statement is finished for the same effect.

lib/seattleflu/id3c/cli/command/etl/redcap.py
jccraft-uw: I added in that note.

lib/seattleflu/id3c/cli/command/etl/redcap.py
jccraft-uw: I fixed that.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -412,16 +418,29 @@ def create_encounter_id(record: REDCapRecord, is_followup_encounter: bool) -> st
...
-        return record["nasal_swab_q"]
+        collection_date = extract_date_from_survey_timestamp(record, "kiosk_registration_4c7f") or record.get("nasal_swab_q")
+
     elif collection_method == CollectionMethod.SWAB_AND_SEND:
-        return record["date_on_tube"] or record["kit_reg_date"] or record["back_end_scan_date"]
-    else:
-        return None
+        # Don't use `date_on_tube` if it's in the future
+        if record.get("date_on_tube") and datetime.strptime(record.get("date_on_tube"), "%Y-%m-%d").date() <= datetime.today().date():
+            collection_date = record.get("date_on_tube")
```
kschwabefry, line 436: To detect bad/future dates in the date_on_tube, I'm not sure "today" is the right comparison point. That is dependent on when the ETL runs, and could give different data accordingly. It seems more accurate to compare the date_on_tube with dates from an instruments like kit reg.

Also, I know this issue was raised for HCT specifically, but I think its important to avoid drift in this type of logic across the various project ETLs (scan and childcare come to mind). I'd suggest putting this specific change to ignore future dates as a single commit across the etls, split off from the present commit here to pull dates from survey timestamps.

kschwabefry (changes requested):
Thanks for adding the comment about when the timestamp gets added. I added additional feedback below.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: OK, I've removed the handling of future date_on_tube values from this PR. I've updated the Trello card to apply to all relevant ETLs. We can make that change across the ETLs at a future point.

kschwabefry: Ok. Also, sorry if this wasn't clear - I was suggesting a different commit not a different PR, but if you want to handle that completely separately that's ok too. Just remember to update the commit message for 7f5f332569ad54c0a3068ab1b1b20dd8d823da01 to remove the language about future dates and this should be good to go on my end.

jccraft-uw: I don't have time right now to touch all the ETLs to add the future date check in.
I've updated the commit message.

kschwabefry (approved):

kairstenfay (approved):
Thanks for making those style changes. the code looks great!

jccraft-uw:
Deployed.

PR #178, opened 2021-01-08
Add function for uw-reopening to insert synthetic DETs

jccraft-uw:
This code is meant to be called from another simple script
in order to generate and insert syntethic DETs for UW Reopening
enrollment records. This is needed because there are so many DETs
that we do not want to bump the ETL revision number to process them all.
Instead, we will generate DETs for ranges of records via an external
script and "sprinkle" them into the database so that these DETs don't
delay real DETs from getting processed in a timely manner.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -5,7 +5,9 @@ import logging
 from dateutil.relativedelta import relativedelta
 from enum import Enum
 from id3c.cli.command.etl import redcap_det
-from id3c.cli.redcap import is_complete, Record as REDCapRecord
+from id3c.cli.redcap import det, is_complete, Project, Record as REDCapRecord
+from id3c.db.datatypes import Json
+from psycopg2.extras import execute_values
 from seattleflu.id3c.cli.command import age_ceiling
 from .redcap import *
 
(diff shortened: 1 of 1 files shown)


kairstenfay (commented):
I don't yet understand how this function is supposed to be used.

Do you plan to have the redcap-det ETL for uw-reopening call this function? If not, it might be better off living in the backoffice repo. 

*edit* If you could provide some insight into how you plan to call this function (e.g. Which simple script plans to call this, and where will it live?), I would find that very helpful.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -840,3 +842,41 @@ def get_date_from_repeat_instance(instance_id: int) -> str:
...
+    dets = []
+
+    while current_record_id <= end_record_id:
+        record['record_id'] = str(current_record_id)
+        dets.append(det(project, record, 'enrollment_questionnaire', generated_by))
+        current_record_id += 1
+
+    dets_json = [
+        (Json(d),)
+            for d in dets ]
```
kairstenfay, line 873: Since we've been bumping into memory issues, I might recommend creating only one `dets` list (instead of `dets` and `dets_json`), and converting to `Json` while appending to the `dets` list as is done in `offer_uw_testing`.

If you've broken up the DET + conversion to JSON steps separately for better readability, another approach could be to convert the det to JSON then append it in the following line. e.g.
```py
        current_det = det(project, record, 'enrollment_questionnaire', generated_by)
        dets.append(Json(current_det))
```

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -840,3 +842,41 @@ def get_date_from_repeat_instance(instance_id: int) -> str:
...
+    record = {
+        "project" : project,
+        "enrollment_questionnaire_complete" : "", # Don't set a value
+        "redcap_event_name" : ENROLLMENT_EVENT_NAME,
+        }
+
+    current_record_id = start_record_id
+    dets = []
+
+    while current_record_id <= end_record_id:
```
kairstenfay, line 866: I'm typically used to seeing `while` statements when we don't know when the loop will exit. In this case, we know in advance when we're finished (e.g. when `current_record_id` > `end_record_id`). In this case, I think it's better to use a `for` loop. e.g.

`for record_id in range(current_record_id, end_record_id + 1):`

tsibley:
Hmm, it seems to me that the `id3c redcap-det generate …` command already does what you want here, or at least nearly does and could be slightly amended.

jccraft-uw:
> Hmm, it seems to me that the `id3c redcap-det generate …` command already does what you want here (and at the least nearly does and could be slightly extended instead of duplicating this code).

@tsibley , the other day when we were talking about this (in standup, I think) you said "or synthetic DETs". That led me down this path. I'll end this path and close this PR.

tsibley:
Ah. Sorry for the misunderstanding. By "synthetic" I meant broadly "not pushed by REDCap".

PR #179, opened 2021-01-12
Update UW reopening views and offers for Winter

jccraft-uw:
In this commit, the views related to UW reopening test invitation
offers and the offer script itself are updated for the new study
protocol for Winter quarter 2021.

1. A new prioritization category has been added. Participants who
live in UW housing will be invited to test every week on the same
day of the week. These participants have been assigned to a uw_housing_group,
and that group gets invited on a certain day each week.

2. All three reasons for positive attestation test invitations
(symptomatic, exposure to known positive, large gathering) now get the
same priority, 1.

3. Surveillance testing now includes only participants who come to
campus 2 or more times a week. They should get invited every 14 days.

4. For baseline testing, invite up to two times a participant who has never
had a sample collected.

5. Participants are excluded from invitations if they have tested positive
for COVID-19 in the past 90 days, either with us (we have a positive test result)
or somewhere else (they give us the date on their daily attestation).

6. Participants who have asked not to receive project alerts (who have the
alerts_off field set to 'off') will not be invited for testing.

7. The answer options for the `daily_exposure_known_pos` variable now break
down 'yes' into one choice that says the participant was exposed less than
48 hours ago and one choice that says the participant was exposed 48 hours
or more ago.

8. `Tier` doesn't have much meaning now with the change in prioritization
categories. Do not sort final results by tier. Do bring it back, though,
so we can display it.

9. The `encountered` column is used in the final sort, just before the identifier
hash.


schema/deploy/shipping/views.sql
```diff
@@ -51,6 +51,8 @@ drop view if exists shipping.incidence_model_observation_v3;
 drop view if exists shipping.incidence_model_observation_v2;
 drop view if exists shipping.incidence_model_observation_v1;
 
+drop view if exists shipping.linelist_data_for_wa_doh_v1;
```
jccraft-uw, line 54: This is a correction to the DOH view that's already in master. The drop needs to be before shipping.hcov19_presence_absence_result_v1.

jccraft-uw:
[UW priority queue: Don't invite a PT who has tested positive in the past 90 days](https://trello.com/c/DUEWvfU3/921-uw-priority-queue-dont-invite-a-pt-who-has-tested-positive-in-the-past-90-days)

jccraft-uw:
[UW priority queue: Add attestation date to the order by list](https://trello.com/c/x0AaYy1W/889-uw-priority-queue-add-attestation-date-to-the-order-by-list)

jccraft-uw:
[UW priority queue: Adjust how selection works for surveillance testing](https://trello.com/c/TCJOgMjz/890-uw-priority-queue-adjust-how-selection-works-for-surveillance-testing)

jccraft-uw:
[UW priority queue: Sync to the clinical team's definition for who needs baseline testing](https://trello.com/c/kOkzjZPi/923-uw-priority-queue-sync-to-the-clinical-teams-definition-for-who-needs-baseline-testing)

schema/deploy/shipping/views.sql
```diff
@@ -3212,27 +3235,85 @@ create or replace view shipping.__uw_priority_queue_v1 as (
...
+            4 as priority,
+            'baseline' as priority_reason
+        from uw_enrollments
+        -- Filter to participants who enrolled after this date when
+        -- baseline test invitations were handled manually
+        where encountered > '2020-12-15'
+         -- Filter to enrollments have never had a sample collected
+        and latest_collection_date is null
+        -- Filter to enrollments that have been invited up to 2 times
+        and times_invited < 2
```
kschwabefry, line 3277: This could be renamed to invitation_count for clarity (in code, I often see time meaning time of day)

schema/deploy/shipping/views.sql
```diff
@@ -3195,12 +3214,16 @@ create or replace view shipping.__uw_priority_queue_v1 as (
...
         -- Filter for postive daily attestations only
         and screen_positive
-        -- Filter for participants who have not tested positive in the past week
-        and prev_pos is not true
+        -- Filter for participants who have not tested positive with us in the past 90 days
+        and (latest_positive_hcov19_collection_date is null or latest_positive_hcov19_collection_date < current_date - interval '90 days')
+        -- Filter for participants who have not tested positive somewhere else in the past 90 days
+        and (latest_prior_test_positive_date is null or latest_prior_test_positive_date < current_date - interval '90 days')
+        -- Filter for participants who will receive alerts from REDCap
+        and (alerts_off is null or alerts_off <> 'off')
```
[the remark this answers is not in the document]
jccraft-uw: We had a good discussion about alerts_off yesterday in the UW Reopening meeting. I asked that the PC team use more crisp/exact words when talking about this. "Opting out" sounds to me like it means opting out of the study, which this is not.
When alerts_off is set (the only value currently is 'off') none of the REDCap alerts will send an email or text to the participant. That means that if we continue to invite these participants they'll never know that they got invited to testing, so those would be "wasted" invites.

schema/deploy/shipping/views.sql
```diff
@@ -3195,12 +3214,16 @@ create or replace view shipping.__uw_priority_queue_v1 as (
         and testing_trigger is null
         -- Filter for postive daily attestations only
         and screen_positive
-        -- Filter for participants who have not tested positive in the past week
-        and prev_pos is not true
+        -- Filter for participants who have not tested positive with us in the past 90 days
+        and (latest_positive_hcov19_collection_date is null or latest_positive_hcov19_collection_date < current_date - interval '90 days')
+        -- Filter for participants who have not tested positive somewhere else in the past 90 days
+        and (latest_prior_test_positive_date is null or latest_prior_test_positive_date < current_date - interval '90 days')
```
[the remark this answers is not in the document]
jccraft-uw: To do that, we would have to include those columns in the queries that make it to the union at the end of the view. I don't want this view to return COVID-19 positivity info because it would then need to be restricted. What I could do is modify those 5 queries to include the columns that we want to filter for and then in the union select specific columns instead of *.
This would be a touch of work, but I do prefer selecting specific columns in the final query that unions, so I'll do it.

schema/deploy/shipping/views.sql
```diff
@@ -3212,27 +3235,85 @@ create or replace view shipping.__uw_priority_queue_v1 as (
...
+            'uw_housing_resident' as priority_reason
         from uw_enrollments
-        -- Filter to enrollments that have never been invited to test or last invite was over 3 days before today
-        where (latest_invite_date is null or latest_invite_date < current_date - interval '3 days')
-        -- Filter to enrollments have never had a sample collected or last sample collection was over 3 days before today
-        and (latest_collection_date is null or latest_collection_date < current_date - interval '3 days')
+        -- Filter to the housing group for today
+        -- CHR 97 is 'a'
+        -- DOW is 0 through 6, starting on Sunday; therefore, Sunday is group 'a'
+        where uw_housing_group = CHR(EXTRACT(DOW from current_date)::int + 97)
```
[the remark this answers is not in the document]
jccraft-uw: That's right. In REDCap, the fields are uw_housing_group_a through uw_housing_group_g. After ingest we end up with values a through g. Seven values, one for each day of the week.

schema/deploy/shipping/views.sql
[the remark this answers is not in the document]
kschwabefry: Oh! That's a really good point about wasted tests, if even that notification gets stopped on the redcap side. That would be really useful to add to the comment here.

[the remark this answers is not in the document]
jccraft-uw: I added that info in. This is now at the very bottom of the view.

schema/deploy/shipping/views.sql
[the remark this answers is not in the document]
jccraft-uw: I made this change.

schema/deploy/shipping/views.sql
jccraft-uw: I renamed that.

kschwabefry (approved):

schema/deploy/shipping/views.sql
```diff
@@ -3043,6 +3049,10 @@ create or replace view shipping.uw_reopening_ehs_reporting_v1 as
   , encounters.live_other_uw as lives_with_uw_students_or_employees
   , encounters.uw_apt_yesno as lives_in_uw_apartment
   , encounters.tier as study_tier
+  , encounters.uw_housing_resident
+  , encounters.on_campus_2x_week
+  , encounters.uw_housing_group
+  , encounters.alerts_off
```
joverlee521, line 3055: Slightly off top, will these fields get added to the EHS REDCap project?

jccraft-uw: I thought it would be good to have these new fields available in the EH&S view in case EH&S do end up wanting to see them in their REDCap project. There's no plan yet to add them, though.

schema/deploy/shipping/views.sql
```diff
@@ -3134,9 +3138,14 @@ create or replace view shipping.__uw_priority_queue_v1 as (
...
             tier,
-            current_prev_pos,
             latest_invite_date,
-            latest_collection_date
+            latest_collection_date,
+            coalesce(invitation_count, 0) as invitation_count,
+            uw_housing_group,
+            on_campus_2x_week,
+            latest_positive_hcov19_collection_date,
+            latest_prior_test_positive_date,
```
joverlee521, line 3147: Do we need to include `prior_test_positive_base` field here in case a participant has never done a daily attestation?

schema/deploy/shipping/views.sql
```diff
@@ -3160,29 +3169,43 @@ create or replace view shipping.__uw_priority_queue_v1 as (
...
-                when daily_exposure_known_pos and age(__uw_encounters.encountered) >= '2 days' then 1
-                when daily_exposure and age(__uw_encounters.encountered) >= '2 days' then 4
-                else null
+                when daily_exposure_known_pos in ('yes', 'yes_now') and age(__uw_encounters.encountered) >= '2 days' then 1
+                when daily_exposure and age(__uw_encounters.encountered) >= '2 days' then 1
+
+                /* The `yes_late` option takes care of the 48 hours. The option label is
+                 * `Yes, I was exposed to COVID-19 more than 48 hours ago`
+                 */
+                when daily_exposure_known_pos = 'yes_late' then 1
```
joverlee521, line 3195: Just want to verify the `yes_now` and `yes_late` options are set in stone. I don't see them in the current HCT REDCap project.

jccraft-uw: From Jessica just now: "I am submitting changes today to update the exposure questions."
She confirmed those two new 'yes' values.

schema/deploy/shipping/views.sql
joverlee521: Ah okay. `uw_housing_resident` and `on_campus_2x_week` are calculated from existing fields, so I wonder if they are redundant. Not sure if `uw_housing_group` and `alerts_off` make any sense outside of the HCT project.

schema/deploy/shipping/views.sql
jccraft-uw: I've brought in the `prior_test_positive_date_base` date from the enrollment survey and used that at the bottom for uber filtering. We might as well use it since we've got it!

schema/deploy/shipping/views.sql
jccraft-uw: I could imagine EH&S wanting to know which day of the week a UW Housing participant gets invited for testing.
I could also see them wanting them to know if a PT doesn't get alerts from REDCap, potentially.
We don't have to put them in the transfer project. I put them in the view for us to have if we need them.

schema/deploy/shipping/views.sql
joverlee521: Would it make sense to coalesce the two fields instead of adding an additional field/filter?

jccraft-uw: The individual could have a different value for `prior_test_positive_date_base` and `latest_prior_test_positive_date`. For example, they give '2020-10-01' on their enrollment. In their daily attestation once they give '2020-12-15'.
We wouldn't want a coalesce here. We'd want the max value and compare that against 90 days.
We could max() all three values, assuming that max works like this ... 
In Postgres this does not work:
select max('2020-01-01'::date, '2020-02-01'::date, '1980-02-02'::date)
Ah, it's `greatest`  https://www.postgresql.org/docs/9.5/functions-conditional.html#FUNCTIONS-GREATEST-LEAST :+1: 

"The GREATEST and LEAST functions select the largest or smallest value from a list of any number of expressions. The expressions must all be convertible to a common data type, which will be the type of the result (see Section 10.5 for details). NULL values in the list are ignored. The result will be NULL only if all the expressions evaluate to NULL."

This does work in Postgres:
select greatest(null, '2020-02-01'::date, '1980-02-02'::date)

This ends up looking complicated:
where
 ( greatest(latest_positive_hcov19_collection_date, latest_prior_test_positive_date, prior_test_positive_date_base) 
    is null or greatest(latest_positive_hcov19_collection_date, latest_prior_test_positive_date, prior_test_positive_date_base) < current_date - interval '90 days')
I could compute the `greatest` earlier and then use that, but to keep things simpler I'm going to leave this as it is with the 3 different date evaluation clauses.

joverlee521: Ah, so you can use the `greatest` function within the `uw_enrollments` CTE to consolidate all three dates into one date field.

jccraft-uw: Right. I could do that and not carry down the individual dates. That ends up with cleaner looking code at the bottom, but I lose the flexibility of having the separate columns. But I could keep all 3 columns and do the `greatest` and carry them all down. 
The end result is the same, so I don't see a huge reason to do that.

joverlee521 (approved):

jccraft-uw:
Deployed

PR #182, opened 2021-01-14
Clia views

joverlee521:
Update views that would be affected once we ingest Research OpenArray SARS-CoV-2 results. 


joverlee521:
[Update shipping views that include hcov19 results](https://trello.com/c/q9tvK51o/959-update-shipping-views-that-include-hcov19-results)

kschwabefry (approved):

schema/deploy/shipping/views.sql
```diff
@@ -1207,7 +1211,7 @@ create or replace view shipping.hcov19_observation_v1 as
         ) as deduplicated_hcov19_bbi
         left join jsonb_to_recordset(replicates) as r("crt" text) on true
 
-        group by sample_id, hcov19_result_received_bbi, hcov19_present_bbi
+        group by sample_id, hcov19_result_received_bbi, hcov19_present_bbi, hcov19_device_bbi, hcov19_assay_type_bbi
```
jccraft-uw, line 1214: It would be good to look at the distinct'ing and group'ing in here in the future to see if this is the right way to come up with one row per sample. Since those operations are not being added via this change, that doesn't hold up this commit.

schema/deploy/shipping/views.sql
```diff
@@ -2235,6 +2235,7 @@ create or replace view shipping.reportable_condition_v1 as
                                    'collections-household-general',
                                    'collections-childcare')
     and coalesce(encountered::date, date_or_null(sample.details ->> 'date')) >= '2020-01-01'
+    and presence_absence.details @> '{"assay_type": "Clia"}'
```
jccraft-uw, line 2246: That makes perfect sense ... reportable conditions ... needs to be Clia.

jccraft-uw (approved):
The added filtering makes sense.
In the perfect universe we would name the views so that it's obvious from the name what kind of results are included. From now on those of us (me) who can't remember the details will have to look at the SQL to know if we're filtering for only Clia, for example.

PR #183, opened 2021-01-15
Add expiration date for UW reopening surges

jccraft-uw:
Before this change, we would invite a participant for a test because of
a surge no matter how long ago the encounter was in which the surge was
marked. We don't want to invite for old surges, therefore this commit
adds a 7 day 'expiration' for surges like we have with positive
attestations.


joverlee521 (approved):

jccraft-uw:
Deployed.

PR #184, opened 2021-01-20
Reduce time between surveillance invites for UW queue

jccraft-uw:
For the UW priority queue view, reduce the time between invitations
from 14 days to 3 days. With 14 days, we end up with days
when we run out of participants to invite for testing. 3 days is a balance.
With 1 day between invitations we would end up nagging participants too
soon.


jccraft-uw:
Deployed.

PR #187, opened 2021-01-23
Add ETL for Adult Family Homes and Workplace project

jccraft-uw:
DRAFT: Awaiting the OK from the clinical team that it's OK to ingest what's being ingested.

Some notes about this project and the ETL:

- This is a classic, not longitudinal project, so we have one instance per record.
- We're not using real AFH or Workplace locations because of PII concerns and also because the clinical team will add locations quite dynamically; the set of sites will not be locked down when the project begins. They'll sign up sites as demand comes in.
- I'm logging the REDCap record number in the logs to make it easier to handle operation support, so we don't have to grep through the log.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/__init__.py
+++ b/lib/seattleflu/id3c/cli/command/etl/__init__.py
@@ -212,4 +212,5 @@ from . import (
     redcap_det_scan,
     redcap_det_uw_reopening,
     redcap_det_childcare,
+    redcap_det_adult_family_home_workplace_outbreak,
 )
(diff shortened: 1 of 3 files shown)


jccraft-uw:
[Create ETL for the Adult Family Homes / Workplace Outbreak project](https://trello.com/c/aDOfHS98/967-create-etl-for-the-adult-family-homes-workplace-outbreak-project)

jccraft-uw:
The project team decided we do not want to ingest this data into ID3C, so I'm closing this PR.

PR #189, opened 2021-01-25
Use datetime instead of date when setting `testing_date`

jccraft-uw:
Use the datetime so we have more info to track which job run
updated which records.


tsibley (approved):

jccraft-uw:
Deployed

PR #195, opened 2021-02-18
Don't warn for unexpected event names for UW-reopening

jccraft-uw:
The UW reopening (HCT) project now includes an event for each month. This
event contains only a monthly survey. We don't ingest that
survey, so it makes sense to skip the record instance that this event
is in. Rather than log a warning, now log info when we see an event
that is not `enrollment` or `encounter`.


kschwabefry (approved):

jccraft-uw:
Deployed.

PR #196, opened 2021-02-20
Childcare ETL: replace 'illness_q_date'

jccraft-uw:
The 2021 Childcare project doesn't have the `illness_q_date` variable like
the pilot version ("Childcare") does. Both projects have this info
collected on a "Symptom Check" survey, so both have
`symptom_check_timestamp' available, so we can use that variable.
The revision is bumped so that we will have this collected consistently
across the two studies.

With this change made locally, the ETL successfully processed all the waiting DETs in the queue.


kschwabefry (approved):

jccraft-uw:
Deployed.

PR #198, opened 2021-02-26
Assembly metadata

joverlee521:
Create an "assembly-exporter" role and a new shipping view `shipping.genome_submission_metadata_v1`. 

For context, this is how I intend to use the view for submissions:
```sql
    create temp table batch(nwgc_id text, sfs_sample_barcode text);
    \copy batch ("$columns") from '"$1"' csv header

    create temp view batch_export as (
        select
            nwgc_id,
            sfs_sample_identifier,
            sfs_sample_barcode,
            collection_date,
            swab_type,
            source,
            puma,
            state
        from batch
        left join shipping.genome_submission_metadata_v1 using (sfs_sample_barcode)
        order by nwgc_id
    );

    \copy (select * from batch_export) to pstdout with csv header
```

### DIFF
--- a/schema/deploy/roles/assembly-exporter/create.sql
+++ b/schema/deploy/roles/assembly-exporter/create.sql
@@ -1,0 +1,10 @@
+-- Deploy seattleflu/id3c-customizations:roles/assembly-exporter/create to pg
+
+begin;
+
+create role "assembly-exporter";
+
+comment on role "assembly-exporter" is
+    'For exporting metadata for consensus genome assembly and submissions';
+
+commit;
--- a/schema/deploy/roles/assembly-exporter/grants.sql
+++ b/schema/deploy/roles/assembly-exporter/grants.sql
@@ -1,0 +1,15 @@
+-- Deploy seattleflu/id3c-customizations:roles/assembly-exporter/grants to pg
+
+begin;
+
+revoke all on database :"DBNAME" from "assembly-exporter";
+revoke all on schema receiving, warehouse, shipping from "assembly-exporter";
+revoke all on all tables in schema receiving, warehouse, shipping from "assembly-exporter";
+
+grant connect on database :"DBNAME" to "assembly-exporter";
+
+grant usage
+    on schema shipping
+    to "assembly-exporter";
+
+commit;
--- a/schema/deploy/shipping/views@2021-03-18.sql
+++ b/schema/deploy/shipping/views@2021-03-18.sql
@@ -37,7 +37,6 @@ drop view if exists shipping.reportable_condition_v1;
 drop view if exists shipping.metadata_for_augur_build_v3;
 drop view if exists shipping.metadata_for_augur_build_v2;
 drop view if exists shipping.genomic_sequences_for_augur_build_v1;
-drop view if exists shipping.genome_submission_metadata_v1;
 drop view if exists shipping.flu_assembly_jobs_v1;
 
 drop view if exists shipping.scan_follow_up_encounters_v1;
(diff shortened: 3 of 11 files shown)


joverlee521:
[New shipping view for exporting metadata for consensus genome submissions](https://trello.com/c/UxxrNeUF/1037-new-shipping-view-for-exporting-metadata-for-consensus-genome-submissions)

schema/deploy/shipping/views.sql
```diff
@@ -2083,6 +2084,46 @@ comment on view shipping.flu_assembly_jobs_v1 is
...
+        sample.collected::date as collection_date,
+        case sample.details ->> 'swab_type'
+          when 'ans' then 'Anterior nasal swab'
+          when 'mtb' then 'Mid-turbinate nasal swab'
+          when 'np' then 'Nasopharyngeal swab'
+          else null
+        end as swab_type,
+        case
+          when identifier_set.name in ('collections-scan', 'collections-scan-kiosks') then 'SCAN'
+          else 'SFS'
```
jccraft-uw, line 2101: If not SCAN then SFS ... this won't hold forever. Apple, for example. 
There's also discussion about us not including Apple data with other data. That's still being discussed, so we don't have to filter anything just yet. I can imagine having to exclude the two Apple collection sets from this view.

jccraft-uw (approved):

schema/deploy/shipping/views.sql
tsibley: SFS/SCAN is the org, not the study. I think all current and future studies are happening under the aegis of SFS.

schema/deploy/shipping/views.sql
```diff
@@ -2083,6 +2084,46 @@ comment on view shipping.flu_assembly_jobs_v1 is
 alter view shipping.flu_assembly_jobs_v1 owner to "view-owner";
 
 
+create or replace view shipping.genome_submission_metadata_v1 as
+
+    select
+        sample.identifier as sfs_sample_identifier,
+        sampleid.barcode as sfs_sample_barcode,
```
tsibley, line 2091: Little weird to me to have these called `sfs_…` when they're not all SFS.

tsibley (approved):
> For context, this is how I intend to use the view for submissions:
>
>     create temp table batch(nwgc_id text, sfs_sample_barcode text);
>     \copy batch ("$columns") from '"$1"' csv header
>     …
>         left join shipping.genome_submission_metadata_v1 using (sfs_sample_barcode)


I'd suggest using `citext` for `sfs_sample_barcode` in the temp view so that the join on it with the same column from `genome_submission_metadata_v1` is robust to case.

schema/deploy/shipping/views.sql
joverlee521: Everything is still under the SFS org, the separation of SCAN is to add extra authors from PHSKC. 
I've asked Trevor on [Slack](https://seattle-flu-study.slack.com/archives/C01MQTSJH44/p1614717832026400) to verify that all other SFS studies can use the general SFS authors list.

joverlee521:
> I'd suggest using citext for sfs_sample_barcode in the temp view so that the join on it with the same column from genome_submission_metadata_v1 is robust to case.

Ah, thank you for the suggestion! I'll change the type to citext in the assembly script.

tsibley (approved):

joverlee521:
Specifically named the schools collections to "Snohomish Schools" since the new Yakima schools study is going to be a completely different/separate study. 

Also added `coalesce` for the collection date to fall back on the encountered date if the collected date is null for some reason. I realized this is needed for retro samples that do not have a date on the sample tubes, but have an encountered date from the clinical metadata.

kschwabefry (approved):

PR #199, opened 2021-03-01
Prefer survey timestamp for UW reopening follow up

jccraft-uw:
For the UW Reopening ETL, when computing the encounter date for the
follow up encounter, use the survey timestamp if it exists then fall
back to the timestamp variable. We had a case where the ETL failed
because the timestamp variable had not been populated.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -322,7 +322,9 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
                 # Don't set locations because the f/u survey doesn't ask for home address.
                 follow_up_encounter_entry, follow_up_encounter_reference = create_encounter(
                     encounter_id = create_encounter_id(redcap_record_instance, True),
-                    encounter_date = redcap_record_instance['fu_timestamp'].split()[0],
+                    encounter_date = extract_date_from_survey_timestamp(redcap_record_instance, 'week_followup') \
+                        or datetime.strptime(redcap_record_instance.get('fu_timestamp'),
+                        '%Y-%m-%d %H:%M:%S').strftime('%Y-%m-%d'),
                     patient_reference = patient_reference,
                     site_reference = site_reference,
                     collection_code = CollectionCode.HOME_HEALTH,


jccraft-uw:
[UW reopening ETL: for the followup encounter, use the survey timestamp](https://trello.com/c/XhT3KpKr/1039-uw-reopening-etl-for-the-followup-encounter-use-the-survey-timestamp)

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -322,7 +322,8 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
                 # Don't set locations because the f/u survey doesn't ask for home address.
                 follow_up_encounter_entry, follow_up_encounter_reference = create_encounter(
                     encounter_id = create_encounter_id(redcap_record_instance, True),
-                    encounter_date = redcap_record_instance['fu_timestamp'].split()[0],
+                    encounter_date = extract_date_from_survey_timestamp(redcap_record_instance, 'week_followup') \
```
kschwabefry, line 325: Adding in the system survey timestamp is a good improvement. Looking at this specific line of code, do you know why for `fu_timestamp this uses` `split()` instead of `datetime.strptime()`?

jccraft-uw: I stole that `split()[0]` from other ETLs, for example, in lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py.

kschwabefry: Gotcha, thanks. It might be cleaner and more explicit to use strptime like down in [line 382](https://github.com/seattleflu/id3c-customizations/blob/6fc37041dff5477be273542f7f41972b62447431/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py#L382). More of a suggestion though, doesn't need to hold up the PR if you disagree.

kschwabefry (approved):

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
jccraft-uw: That is a great thing to do. There's no reason to split that datetime as a string. I've made that change.

jccraft-uw:
Deployed.

PR #200, opened 2021-03-03
Childcare ETL: Use survey timestamp correctly

jccraft-uw:
When using a REDCap server generated survey timestamp, we should do so
by using the `extract_date_from_survey_timestamp` function instead
of accessing the timestmap variable directly. If the survey has not
been completed, a value of `[not completed]` will be returned for
the timestamp. The function handles this correctly.

After this is merged, generate a DET for the 2021 Childcare project record 178.
For example:
REDCAP_API_URL="https://redcap.iths.org/api/" envdir /home/jccraft/gitroot/backoffice/id3c-production/env.d/redcap id3c redcap-det generate --project-id=29351 "178" > /home/jccraft/jccraft_temp/det_work/det.ndjson


kschwabefry (approved):

jccraft-uw:
Deployed. Also pulled and ingested the DETs for that record.

PR #201, opened 2021-03-05
Use the symptom_check_timestamp the right way

jccraft-uw:
In the Childcare ETL I fixed how we use the `symptom_check_timestamp` for the
collection_date, but I missed the second place where we use it. This
commit calls `extract_date_from_survey_timestamp` to use the
`symptom_check_timestamp` for the questionnaire item.

After this is merged, generate a DET for the 2021 Childcare project record 178.
For example:
REDCAP_API_URL="https://redcap.iths.org/api/" envdir /home/jccraft/gitroot/backoffice/id3c-production/env.d/redcap id3c redcap-det generate --project-id=29351 "178" > /home/jccraft/jccraft_temp/det_work/det.ndjson


jccraft-uw:
Deployed and pulled the DETs.

PR #202, opened 2021-03-09
Add surveillance questionnaire item for UW

jccraft-uw:
The revision is not being updated. Instead, I will pull all records.

PR #203, opened 2021-03-10
Add added_surveillance_groups into uw_priority_queue

jccraft-uw:
For UW Reopening (HCT), a new variable `added_surveillance_groups` has
been added to the enrollment questionnaire. The values for this
variable are the days of the week (lower case). We should invite
a participant if for `added_surveillance_groups` they have the value
that matches today's day of the week, unless they're also a UW Housing
resident.
Details are in https://trello.com/c/OyvaRQpD.

This does relate to (but is not blocked by) https://github.com/seattleflu/id3c-customizations/pull/202
And pulling all records to populate that new item.


jccraft-uw:
[Test offers: include new weekly invite group](https://trello.com/c/OyvaRQpD/1029-test-offers-include-new-weekly-invite-group)

schema/deploy/shipping/views.sql
```diff
@@ -3283,10 +3286,43 @@ create or replace view shipping.__uw_priority_queue_v1 as (
...
+            redcap_record_id,
+            redcap_event_name,
+            redcap_repeat_instance,
+            encountered,
+            individual,
+            tier,
+            latest_invite_date,
+            latest_collection_date,
+            2 as priority,
+            'surveillance' as priority_reason,
```
joverlee521, line 3309: I just want to confirm my understanding that this new surveillance group has the same priority as uw_housing_resident. 

Additionally would it be confusing for people if some surveillance have priority of 2 while other surveillance have priority of 5.

schema/deploy/shipping/views.sql
```diff
@@ -3283,10 +3286,43 @@ create or replace view shipping.__uw_priority_queue_v1 as (
         -- CHR 97 is 'a'
         -- DOW is 0 through 6, starting on Sunday; therefore, Sunday is group 'a'
         where uw_housing_group = CHR(EXTRACT(DOW from current_date)::int + 97)
+        -- Filter to participants who are not in an added surveillance group becuase they already get invited weekly
+        and added_surveillance_groups is null
```
joverlee521, line 3290: Is it possible for a participant to be (accidentally) marked for both `uw_housing_group` and `added_surveillance_groups` in REDCap? If that's the case, then that PT would be excluded from both groups in this priority queue...

I'm not sure if this is something we can guard against in REDCap. Is this something we can guard/alert on in the [ETL update](https://github.com/seattleflu/id3c-customizations/pull/202)?

schema/deploy/shipping/views.sql
jccraft-uw: That's right. This new group has a priority of 2 to increase the odds that everybody who should get invited on a day gets invited. We have enough quota every day to get through every group except surveillance, priority 5.
While this group has priority 2, it's labeled as "surveillance". From the REDCap point of view on the testing determination form, there won't be a distinction between PTs invited for regular surveillance and for this new added surveillance. See Slack thread https://seattle-flu-study.slack.com/archives/C018V0WGGJ0/p1615327884004000

schema/deploy/shipping/views.sql
jccraft-uw: I've resisted putting this kind of check into the ETL because I didn't want to deal with more warnings in  id3c-alerts. But we say that the value of our ETLs is the data checks that we do inside them, so that is the right place to put this check.
I put that into Trello card https://trello.com/c/OFqlR28R
It's quick work, so I stuck it into Next Up. It won't block this SQL change but it would be good to have done soon.

jccraft-uw:
Deployed.

PR #205, opened 2021-03-16
Accommodate unset UW housing group variables

jccraft-uw:
The UW Reopening ETL's data quality check needs to accommodate records
that didn't have their UW Housing Group variables set when the REDCap
calculations were implemented. If the value is '' consider it 0 for the
housing group count.


jccraft-uw:
Deployed.

PR #207, opened 2021-03-26
Pull `testing_date` from the testing_determination_internal form

jccraft-uw:
For the UW-reopening ETL, by pulling the `testing_date` from the
testing_determination_internal form we will be able to give a more accurate
analysis of how much time passes between invitations for a given
participant.

The revision number is not incremented in this commit because all
records will be pulled manually.

I removed a sentence from the code comment that was not accurate.


jccraft-uw:
[UW Reopening: Improve the "days between invitations" data on the dashboard](https://trello.com/c/FSwihsek/1060-uw-reopening-improve-the-days-between-invitations-data-on-the-dashboard)

jccraft-uw:
Deployed.

PR #214, opened 2021-05-26
Extract Metabase details for each connection from the query comment 

tsibley:
These details will let our connection watchdog processes (e.g. for slow
queries) be more selective in what they terminate without revealing the
full query of every connection to non-superusers.

The format of this comment (or "remark" in Metabase's parlance) is
reasonably guessable, but I also checked the code.¹  If the format
changes in the future and the parse fails, the new "metabase" column
added here will just be null.  This seems like a reasonable failure
mode.

¹ https://github.com/metabase/metabase/blob/f76d8333/src/metabase/query_processor/util.clj#L26-L37

--- 

See also https://github.com/seattleflu/backoffice/pull/160#issuecomment-848177312.


kschwabefry (approved):
LGTM, and I agree this is a good approach given the choices.
My only suggestion is to update https://github.com/seattleflu/backoffice/blob/master/metabase/README.md to mention this PR, as its something we would want to inspect during metabase version upgrades.

PR #216, opened 2021-06-11
Fix Mypy

tsibley:
Fixes failing Mypy tests on master, which started failing due to newer Mypy. Complement to https://github.com/seattleflu/id3c/pull/227.


kschwabefry (approved):
Thanks for this Tom!

PR #221, opened 2021-07-12
Add shipping.scan_hcov19_result_counts_v2 view for SCAN dashboard

seidelma:
Add a new view that groups results by date and priority code, so Formative can exclude specific codes from the result set displayed on the SCAN dashboard.


kschwabefry (changes requested):
Thanks for adding in priority codes. We also need to provide information on whether it was a group enrollment. You can use `scan_study_arm` for that.

For your commit, please also add a description to give context and explain why this v2 of the view was added.

kschwabefry:
Also, just making sure this is on your radar - since this ended up being a new version of the view, we'll want to update the API to add a new endpoint that uses this view so the clients (Formative) can switch over.

kschwabefry (approved):

lib/seattleflu/id3c/api/routes.py
```diff
@@ -131,6 +131,19 @@ def get_scan_positives(session):
     return Response((row[0] + '\n' for row in positives), mimetype="application/x-ndjson")
 
 
+@api_v1.route("/shipping/scan-hcov19-positives-prio", methods = ['GET'])
```
kschwabefry, line 134: I wonder if we want to make this 
`@api_v2.route("/shipping/scan-hcov19-positives", methods = ['GET'])`
since its very similar data, with additional data/grouping. Thoughts?

seidelma: Done and done!

kschwabefry (approved):
Hi Todd, the changes look good. 
My only last feedback is similar to last time about commit structure, that the last three could be squashed into one. Because they are essentially changing the api function name a few times - it would be a clearer commit history for future us and other team members if https://github.com/seattleflu/id3c-customizations/pull/221/commits/112bf0eedf89e34b6a3d2a9e28b451af73841ced , https://github.com/seattleflu/id3c-customizations/pull/221/commits/66cdd0d44d29f53ad595d1bbfd79e7a838eccacf and https://github.com/seattleflu/id3c-customizations/pull/221/commits/a78033c9714bbca49c068db6e0ec3c6ae9a28455 were squashed into one commit.

PR #222, opened 2021-07-12
Add RetrospectivePHSKC site for residuals from PHSKC's COVID-19 testing sites

tsibley:
These come from UW Virology, but unlike other "retrospectives", they're
not originating from the hospital.  The site category and type reflect
this.  The site type is unique so that these samples can be easily
identified in downstream data.


schema/deploy/warehouse/site/data.sql
```diff
@@ -49,6 +49,8 @@ insert into warehouse.site(identifier, details)
             "sample_origin": "hmc_retro"}'),
         ('RetrospectiveNorthwest',                  '{"category": "hospital",   "type": "retrospective",
             "sample_origin": "nwh_retro"}'),
+        ('RetrospectivePHSKC',                      '{"category": "community",  "type": "phskc",
```
kschwabefry, line 52: You mention in the commit message that you used `"type": "phskc"` so it could be distinguished from the hospital retrospectives. Could we rely on `category` to distinguish, if this was added as `'{"category": "community",  "type": "retrospective", "sample_origin": "phskc_retro"}'),`? I think the other retrospectives do have `"category": "hospital"`.

tsibley: We could now, yes. Or even just rely on site identifier now. That commit message is slightly out of date now, I realize, since https://github.com/seattleflu/backoffice/pull/182 changed to map sites → channels in SQL instead of in the viz, so the full site record is now available instead of only site type.

I think I'm ambivalent about changes here, so happy to do what you'd prefer.

kschwabefry: I guess it boils down to a taxonomy decision. My understanding is that we came up with the vocabulary here, and this isn't defined as part of FHIR - is that correct? 
If its up to us to define, I think I personally would lean towards `"category": "community", "type": "retrospective"`, because it seems the differentiation can be equally acheived using `category`, and that lets us maintain some cohesion in `type`.

tsibley: Right, it's our vocabulary, so any clarity/muddiness is ours to create. ;-)

I think category: community, type: retro is good. Updated.

tsibley:
Deployed.

PR #226, opened 2021-07-22
Tear down the hCoV-19 wall implemented by row-level security policies

tsibley:
In early March 2020, we built a wall in our production ID3C database
instance between "hCoV-19" (SARS-CoV-2, COVID-19) test results and
everything else.  This allowed us to grant only a small number of
internal study team members access to the SARS-CoV-2 test results.
These results were considered extremely sensitive at the time because of
their high public interest.

The wall persisted until now but is largely meaningless (and has been
for a while).  Much of the study team has visibility into SARS-CoV-2
test results and the results are no longer considered nearly as
sensitive.  The lab's testing and return of results to participants is
also on much firmer ground than it was back when the wall went up.

Tearing down the wall simplifies the internals of our production ID3C
database instance and will make data queries easier in Metabase.  It is a bit
of maintenance that's overdue.

These Sqitch reworks remove the functional partition of and restrictions
in the database.  An organizational partition still exists in Metabase,
which can be addressed separately at any future point.


schema/deploy/roles/hcov19-visibility/create.sql
```diff
@@ -2,9 +2,6 @@
 
 begin;
 
-create role "hcov19-visibility";
-
-comment on role "hcov19-visibility" is
-    'Allows access to HCoV-19 presence/absence results';
+drop role "hcov19-visibility";
```
kschwabefry, line 5: Would this need to be `drop role if exists` so it could be rerun again even after the role has been removed, without throwing an error? (And similar question for the `view-owner` role)

tsibley: Yes, if it had to be rerun. But I'm not sure when that scenario would arise? The change would have to be `sqitch rework`ed, at which point I'd expect the `drop role` line to change anyway?

tsibley:
Repushed with a minor tidy up to remove an unnecessary column-level select grant on `receiving.presence_absence` from reporter, which gained back a table-level select grant.

I've tested deploying this against `testing.db.seattleflu.org/testing` and it appears to work as-expected without any surprises.

schema/deploy/roles/hcov19-visibility/create.sql
kschwabefry: Mhmm. Yes, that makes sense. Thanks!

kschwabefry (approved):

davereinhart:
Deployed.

kschwabefry:
[Remove hcov-19 visibility restrictions in ID3C and metabase](https://trello.com/c/73IkTXeH/1045-remove-hcov-19-visibility-restrictions-in-id3c-and-metabase)

PR #231, opened 2021-07-28
Update view shipping.genome_submission_metadata_v1

joverlee521:
Adds new study collection identifier and the new tiny swabs that will
be used for HCT study.

Adds Washington state counties based on the FIPS code in the linked
census tract.

Adds a boolean `baseline_surveillance` column to indicate baseline
samples. Baseline samples include clinical residuals, normal SCAN
samples, and a subset of SCAN priority codes that are used to
shift the sampling frame of the study. This was determined based on the
guidelines set by CDC
(see https://www.aphl.org/programs/preparedness/Crisis-Management/Documents/Technical-Assistance-for-Categorizing-Baseline-Surveillance.pdf)

See [Slack convo](https://seattle-flu-study.slack.com/archives/C01MQTSJH44/p1626284273033800) regarding baseline samples.


joverlee521:
[Update genome metadata views to include SCAN priority codes/ group enrollment](https://trello.com/c/SgdYUqDy/1150-update-genome-metadata-views-to-include-scan-priority-codes-group-enrollment)

schema/deploy/shipping/views.sql
```diff
@@ -2091,19 +2093,74 @@ create or replace view shipping.genome_submission_metadata_v1 as
...
+          when '53065' then	'Stevens'
+          when '53067' then	'Thurston'
+          when '53069' then	'Wahkiakum'
+          when '53071' then	'Walla Walla'
+          when '53073' then	'Whatcom'
+          when '53075' then	'Whitman'
+          when '53077' then	'Yakima'
+          else null
+        end as county,
         -- If location is null, assume the sample is from within Washington.
```
seidelma, line 2145: Is it still appropriate to assume the sample is from WA when the above CASE block results in NULL?

seidelma (commented):
I just had one question about the location logic

schema/deploy/shipping/views.sql
joverlee521: As far as I know, SFS/SCAN only tests samples within Washington state. 
Even if we don't have detailed location data, I think it's safe to assume it's from WA.

seidelma: Sounds good to me, I'll mark it resolved and approve away.

joverlee521:
Deployed

PR #232, opened 2021-08-05
Update shipping views for uw-ehs-report

davereinhart:
Optimizations for 2 related shipping views: return_results_v3 and
uw_reopening_ehs_reporting_v1. The uw-ehs-report cron job has had
performance issues related to the response time of the latter view which
is used to export to CSV.

These changes improve the performance of the view by updating the where
clause to use a boolean value instead of matching using `any` with an
array of text values.


davereinhart:
Run `SELECT * FROM shipping.uw_reopening_ehs_reporting_v1` on local db before and after `sqitch deploy dev` to see performance improvement (`EXPLAIN ANALYZE` can be added to show more details).

seidelma (approved):
Looks good to me. In the interest of time let's merge this now, and I'll add a writeup of the root cause when it's ready.

davereinhart:
Deployed.

PR #234, opened 2021-08-10
Reorder participants in HCT priority testing queue

kschwabefry:
Prioritize participants who have been least recently invited
to test over participants who have been least recently tested
in our priority queue used for automated testing invitations.

This change was decided on with the study team to try to mitigate
the issue of surveillance invitations getting mostly assigned
to the same group of not very responsive pariticpants. The goal
here is to start trying to spread invitations more evenly across
participants who are eligible for surveillance testing.


kschwabefry:
[Update prioritization order for HCT priority queue](https://trello.com/c/0wnzKOzw/1148-update-prioritization-order-for-hct-priority-queue)

davereinhart:
This looks good!

Just reviewing the numbers and the Slack thread, I'm seeing that there are currently ~4600 records in the priority queue, and with them having us set the quota to 1400 for weekdays, x 3 days before participants are re-invited is 4200 people who should receive be receiving them, so there would liklely be ~400 people who are not getting invites because they're at the bottom of list.  It seems like bumping up the quota bit to make sure a 3 day stretch covers the number of total # of records in the queue might help make sure everyone gets an invite, or bumping from 3 to 4 days before someone is re-invited.

davereinhart (approved):

kschwabefry:
> This looks good!
> 
> Just reviewing the numbers and the Slack thread, I'm seeing that there are currently ~4600 records in the priority queue, and with them having us set the quota to 1400 for weekdays, x 3 days before participants are re-invited is 4200 people who should be receiving them, so there would likely be ~400 people who are not getting invites because they're at the bottom of list. It seems like bumping up the quota a bit to make sure a 3 day stretch covers the total number of records in the queue might help make sure everyone gets an invite, or bumping from 3 to 4 days before someone is re-invited.

Thanks for digging into it! I think there are two pieces that may not obvious there that factor into the math, so it should turn out there are about 3k-4k participants who should again start getting invitiations for testing.

I'm guessing you looked at the length of the priority queue being 4600 after exclusions. If you include participants who have been invited in the past three days, you get a slightly higher total in the queue, roughly 6700 in the surveillance testing pool. Another thing is that not all invites are going to surveillance - we run through some invites for postitive attestations, but also a good number of standarized weekly testing because of living in UW residences or other affiliations. So we send out ~9k invites/week, and lets say ~7k invites are going to surveillance.

It roughly breaks down about 50/50: about half of the surveillance-eligible participants haven't been invited or tested since roughly June or longer ago, and the other half are getting regular invites twice a week (~3500 participants * 2 invitations per week ~= 7000 invites), but generally not using those invites, so that their last collection date remains pretty long ago and they remain perpetually at the top of the queue.

davereinhart:
Ok, that makes sense.  I was looking at the total count in `shipping.__uw.priority_queue_v1`, but there's a lot going on there so it's tough to follow, but I see there is a where clause filtering on `latest_invite_date < CURRENT_DATE - 3`.

kschwabefry:
Sqitch deployment done in prod.

PR #235, opened 2021-08-11
Switch to new REDCap instance for HCT

seidelma:
This PR completely cuts the affected views over to the new REDCap server. It shouldn't be deployed to prod until the new server is actually getting fed with prod data.


schema/deploy/shipping/views.sql
```diff
@@ -3271,7 +3271,7 @@ create or replace view shipping.__uw_priority_queue_v1 as (
         join warehouse.individual using (individual_id)
         join shipping.uw_reopening_enrollment_fhir_encounter_details_v1 using (encounter_id)
         left join uw_individual_summaries on uw_individual_summaries.individual = individual.identifier
-        where encounter.details @> '{"_provenance": {"redcap": {"url": "https://redcap.iths.org/", "project_id":23854, "event_name": "enrollment_arm_1"}}}'
```
kschwabefry, line 3274: 👍 I think this is right for the uw_priority_queue view. We'd only want to be sending out invitations in the new project once we roll over.

schema/deploy/shipping/views.sql
```diff
@@ -2989,7 +2989,7 @@ create or replace view shipping.uw_reopening_encounters_v1 as
   , site_id
   , details
   from warehouse.encounter
-  where encounter.details @> '{"_provenance": {"redcap": {"url": "https://redcap.iths.org/", "project_id":23854, "event_name":"encounter_arm_1"}}}'
```
kschwabefry, line 2992: For encounters, since those aren't getting copied over, I think we'd want this to include both old and new server projects.

schema/deploy/shipping/views.sql
```diff
@@ -3220,7 +3220,7 @@ create materialized view shipping.__uw_encounters as (
 	left join shipping.fhir_questionnaire_responses_v1 q_testing_trigger on q_testing_trigger.encounter_id = encounter.encounter_id and q_testing_trigger.link_id = 'testing_trigger'
 	left join shipping.fhir_questionnaire_responses_v1 q_surge_selected_flag on q_surge_selected_flag.encounter_id = encounter.encounter_id and q_surge_selected_flag.link_id = 'surge_selected_flag'
 	left join shipping.fhir_questionnaire_responses_v1 q_prior_test_positive_date on q_prior_test_positive_date.encounter_id = encounter.encounter_id and q_prior_test_positive_date.link_id = 'prior_test_positive_date'
-	where encounter.details @> '{"_provenance": {"redcap": {"url": "https://redcap.iths.org/", "project_id":23854, "event_name": "encounter_arm_1"}}}'
+	where encounter.details @> '{"_provenance": {"redcap": {"url": "https://hct.redcap.rit.uw.edu/", "project_id":45, "event_name": "encounter_arm_1"}}}'
```
kschwabefry, line 3223: For encounters, since those aren't getting copied over, I think we'd want this to include both old and new server projects. BUT, if it does include both, it would be great to test how this will interact with the uw_priority_queue_v1 view below, used for selecting HCT participants for testing invitations. I think this would be great to run with some test data in the new project to see how it would handle different situations.

schema/deploy/shipping/views.sql
```diff
@@ -2954,7 +2954,7 @@ create or replace view shipping.uw_reopening_enrollment_fhir_encounter_details_v
   ,(select date_response[1] from shipping.fhir_questionnaire_responses_v1 where encounter_id = encounter.encounter_id and link_id = 'vaccine') as received_this_season_flu_vaccine_date
 
   from warehouse.encounter encounter
-  where encounter.details @> '{"_provenance": {"redcap": {"url": "https://redcap.iths.org/", "project_id":23854, "event_name":"enrollment_arm_1"}}}'
+  where encounter.details @> '{"_provenance": {"redcap": {"url": "https://hct.redcap.rit.uw.edu/", "project_id":45, "event_name":"enrollment_arm_1"}}}'
```
kschwabefry, line 2957: I think all the enrollments are getting copied into the new project, so this should be ok to just pull from new. (But, if we aren't guaranteed to copy enrollment for all records, maybe we'd want to do something here to include both and then deduplicate on the invididual)

seidelma:
I rebased with the changes Kristen suggested--I hope that simply checking for both Redcap projects is a sane way to do it.

seidelma:
I've done some testing with `EXPLAIN` on these view-definition queries to make sure that using `WHERE .. OR ..` to select multiple Redcap servers isn't going to blow up like we saw with `ARRAY`. 

There's a significant performance penalty due to the BitmapOr/Bitmap Heap Scan operation, such that queries with `WHERE .. OR ..` are many times slower:

shipping.uw_reopening_encounters_v1:
  old: Execution time: 3.744 ms
  new: Execution time: 10432.748 ms

shipping.__uw_encounters:
  old: Execution time: 4.864 ms
  new: Execution time: 1477.948 ms

I guess the question is, is this an acceptable performance penalty?

kschwabefry:
@seidelma good thinking ahead on looking at performance. Can you share the explain output?

seidelma:
I've done some testing with a few different strategies to deal with this performance hit.

First, using the JSON 'get' operator (`->`) to reach into the `encounter.details` field and pull out the `url` tag. This is actually much slower than the 'contains' (`@>`) operator, because Postgres can't use the index for that, and it's got to do a full table scan:

    explain analyze verbose select encounter_id as enrollment_encounter_id
      , individual_id as enrollment_individual_id
      from warehouse.encounter
      where
          encounter.details @> '{"_provenance": {"redcap": {"url": "https://hct.redcap.rit.uw.edu/", "project_id":45, "event_name": "encounter_arm_1"}}}';
    Execution time: 1.315 ms

    explain analyze verbose select encounter_id as enrollment_encounter_id
      , individual_id as enrollment_individual_id
      from warehouse.encounter
      where
           encounter.details->'_provenance'->'redcap'->>'url' = 'https://hct.redcap.rit.uw.edu/';
     Execution time: 4560.658 ms

Next, to avoid the use of `WHERE .. OR ..` through `UNION ALL`, to basically concatenate the two queries (one for each Redcap server) together. Performance for this is similar to `WHERE .. OR ..`, because the query planner reorganizes it into the same parallel bitmap heap scan.

Failing that, I tried @davereinhart's solution to troubleshooting slow queries the other day: `encounter.details @> ANY( VALUES(..))`. That doesn't seem to work here--the query ends up being much, much slower than any of the other stuff I've done because it fails to use the index, and instead falls back to a seq scan on the encounter table.

Then I thought about reorganizing things into subqueries so we can avoid the JSON and look at integers using `CASE` to assign a number to each server. Unfortunately that also results in a seq scan.

Haven't given up yet, but that's where I'm currently at.

seidelma:
'Rebased' to current master so we avoid a merge conflict.

PR #236, opened 2021-08-20
etl redcap: ignore 'not complete' survey timestamps

kschwabefry:
We've recently started seeing redcap records with survey instruments that
have a status of complete, but the automated timestamp value is the string
['not complete'] instead of an actual timestamp.

When looking for survey timestamps, we can consider that 'not complete'
value to mean the survey is indeed not complete, and return no extracted
timestamp, instead of throwing an error.


davereinhart (approved):

davereinhart:
Deployed.

PR #238, opened 2021-08-24
backfill-sample-ids: New command to backfill sampleid for sample records

kschwabefry:
Add a new id3c command to mint new sample identifiers and assign them to
warehouse.sample records that are missing sample identifiers. This is useful
for tiny swabs samples, which do not have room on the tube for a sample
identifier, only a collection identifier. This process fills in the sample
record in our warehouse table but notably does not notify the lab of any
identifier assignments. We can use this backfilled sample ids as an
identifier when sharing data with third parties.

Once this is ready, we can add a cronjob in backoffice to run this periodically.


kschwabefry:
[Tinyswabs: Create and assign sample ids to samples](https://trello.com/c/sNa09ZHp/1134-tinyswabs-create-and-assign-sample-ids-to-samples)

davereinhart:
This approach makes sense but this could potentially be handled without a new cron job by adding this process to the `upsert_sample` function so the sample identifiers for tiny swabs are created at the same time as the sample records. One advantage of that approach is that when LIMS starts using the new API endpoint, the response could include the new sample identifier in case they want to store that in LIMS.

kschwabefry:
> This approach makes sense but this could potentially be handled without a new cron job by adding this process to the `upsert_sample` function so the sample identifiers for tiny swabs are created at the same time as the sample records. One advantage of that approach is that when LIMS starts using the new API endpoint, the response could include the new sample identifier in case they want to store that in LIMS.

Yep! There was a few possible designs for when to generate and assign sample ids. When we discussed a few weeks ago, we had decided not to do it in real time with that /samples API because it would add too much overhead time to that call to mint the sample identifier in real time. Another option was to mint them and pair them up when we made the collection identifiers, but that is also a semi-real time process for someone, and would effectively double the run time there (and potentially create sample records that aren't guaranteed to be used). So, that being said, without a clear need for the LIMS/other processes to know about it up front - I think we landed on a regular running backfill as the path forward. Did it seem like there was a new need for LIMS to know about the sample id we are assigning?

davereinhart:
> > This approach makes sense but this could potentially be handled without a new cron job by adding this process to the `upsert_sample` function so the sample identifiers for tiny swabs are created at the same time as the sample records. One advantage of that approach is that when LIMS starts using the new API endpoint, the response could include the new sample identifier in case they want to store that in LIMS.
> 
> Yep! There was a few possible designs for when to generate and assign sample ids. When we discussed a few weeks ago, we had decided not to do it in real time with that /samples API because it would add too much overhead time to that call to mint the sample identifier in real time. Another option was to mint them and pair them up when we made the collection identifiers, but that is also a semi-real time process for someone, and would effectively double the run time there (and potentially create sample records that aren't guaranteed to be used). So, that being said, without a clear need for the LIMS/other processes to know about it up front - I think we landed on a regular running backfill as the path forward. Did it seem like there was a new need for LIMS to know about the sample id we are assigning?

Ok, I forgot we had already discussed that option. I don't know of a specific need for LIMS to know the sample ID but it seems like it could be useful for QC. I'd be curious to see how much overhead it would add to the samples API, but if there's no need for it to happen in real-time then what you have here makes more sense.

kschwabefry:
> Ok, I forgot we had already discussed that option. I don't know of a specific need for LIMS to know the sample ID but it seems like it could be useful for QC. I'd be curious to see how much overhead it would add to the samples API, but if there's no need for it to happen in real-time then what you have here makes more sense.

Sounds good. Fwiw, minting a new identifier currently takes roughly a few seconds per identifier, depending a bit on database load and luck (but is expected to keep slowly but steadily rising over time as we mint more identifiers and distance checks fail more frequently.)

davereinhart:
@kschwabefry Trying to test this locally I got this error (`jsonschema` is a recently added dependency on `id3c`):
`pkg_resources.DistributionNotFound: The 'jsonschema' distribution was not found and is required by id3c`

The only way I found to resolve it was to run `pipenv lock` and `pipenv sync`, and that updates `Pipfile.lock` here, so that may need to be added to this branch.

lib/seattleflu/id3c/cli/command/backfill_sample_ids.py
```diff
@@ -0,0 +1,51 @@
...
+        LOG.info(f"Assigning sample id {sample_identifier.uuid} to collection id {collection_identifier.collection_identifier} in warehouse.samples")
+
+        upsert_sample(db,
+            update_identifiers= True,
+            identifier = sample_identifier.uuid,
+            collection_identifier = collection_identifier.collection_identifier,
+            collection_date = None,
+            encounter_id = None,
+            additional_details = {}
+        )
```
davereinhart, line 50: Works great! My only thought is that `upsert_sample` may be overkill here since it's going to run additional queries to check whether the record exists before it runs the update query. In this case you already have confirmed the records exist, so could get away with a simple `cursor.execute("update ...")`

davereinhart (approved):
Looks good to me. See comment above regarding missing `jsonschema` dependency.

PR #239, opened 2021-08-27
etl uw-reopening: Include kit registration barcodes

kschwabefry:
When creating encounters in the uw reopening ETL, consider kit
registration barcode when trying to create the specimen.

There has been a change in the testing workflow, such that some
samples will be unobserved testing (not kiosk, but also not
swab and send), and they won't have a PCDEQC filled out.
For these records, the way to link the redcap encounter to
the sample is to use the kit registration barcode field,
called barcode_swabsend.


davereinhart:
@kschwabefry Does this one need to be deployed now that we're processing DETs from the new HCT project?

kschwabefry:
Deployed.

PR #240, opened 2021-08-31
SCAN ETL COVID-19 vaccination fields

davereinhart:
Adding newly requested REDCap fields for ingestion into ID3C from SCAN illness questionnaire. Newly ingested fields include overall risk (Jan 2021) plus COVID-19 vaccination status, dates, and details. Also adding these new fields plus related calculated fields to relevant shipping views.


davereinhart:
[Add more questions from SCAN REDCap illness questionnaire to ETL and shipping views.](https://trello.com/c/eGB0U9QA/1168-add-more-questions-from-scan-redcap-illness-questionnaire-to-etl-and-shipping-views)

davereinhart:
This will have some merge conflict to resolve with PR #235. Since merging/rebasing gets messy with `sqitch`, one of these PRs will should probably be redone with fresh commits after the other one is merged.

davereinhart:
I just realized these same questions are also asked in the enrollment form. So this branch may need some changes if we also decide to ingest those.

If we do ingest them from the enrollment form, it brings up the issue of how to deal with conflicting information entered in the enrollment and illness questionnaires. Illness questionnaire will contain more recent information and should probably override what's in the enrollment form in most cases. It's unlikely (but possible) that enrollment form could indicate vaccine was received and illness questionnaire indicates differently, but what's probably more likely is that the dates of vaccination could be different if participant doesn't recall exact date (same could be true with vaccine manufacturer).

We could handle this by always relying on the more recent information, or basing `fully_vaccinated` status on the most recent specified dose date, and/or setting `vaccination_status` to "unknown" if the information doesn't match up.

davereinhart:
@kschwabefry, @seidelma Rebased and ready for review.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
```diff
@@ -913,13 +913,22 @@ def combine_multiple_fields(field_prefix: str) -> Optional[List]:
         'social_precautions',
         'no_mask',
         'high_risk_feb2021',
+        'overall_risk_jan2021',
+        'covid_vax',
+        'covid_doses',
+        'vac_name_1',
+        'vac_date',
+        'vac_name_2',
+        'vac_date_2',
```
kschwabefry, line 922: I think `vac_date` and `vac_date_2` should only need to be in `date_questions`, not `string_questions`

schema/deploy/shipping/views.sql
```diff
@@ -1486,7 +1549,37 @@ create materialized view shipping.scan_encounters_v1 as
...
+        vac_date,
+        vac_name_2,
+        vac_date_2,
+        case
+          when covid_vax = 'no' then 'not_vaccinated'
+          when covid_vax in (values('dont_know'), ('dont_say')) then 'unknown'
+          when covid_vax = 'yes' then
+            case
+              when vac_name_1 = 'johnson' and date_part('day', encountered::timestamp - vac_date::timestamp) >= 14 then 'fully_vaccinated'
+              when vac_name_1 in (values('pfizer'), ('moderna')) and vac_name_2 in (values('pfizer'), ('moderna')) and date_part('day', encountered::timestamp - vac_date_2::timestamp) >= 14 then 'fully_vaccinated'
```
kschwabefry, line 1565: There's a funny edge case here, if a participant happened to mix manufacturers so one does of Pfizer and one dose of Moderna -- do we want to apply the same rules for calculating `vaccination_status` and `fully_vaccinated_date` in that case or do we need to differentiate that?

I'm not sure how these calculations will hold us as booster doses come into play, so I have mixed feelings about incorporating all of this logic into the shipping view. I think we'd need to get pretty specific guidance on how to keep these calculations up to date when SCAN starts collecting booster dose info from participants.

kschwabefry (changes requested):

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
davereinhart: You're right, thanks for catching that!

schema/deploy/shipping/views.sql
davereinhart: Good point. I was basing this off of the original request that stated "fully_vaccinated is everyone who is 14 days after both mRNA doses or 14 days after J&J dose", which I interpreted as lumping the Pfizer and Moderna vaccines as mRNA doses.  So someone who got one of each would still be "fully_vaccinated" after their second dose, and the `vaccine_manufacturer` value would be "multiple" so would be possible to separate them out from those who got both doses from the same manufacturer.

I had the same thought about whether this logic belongs in the shipping view. It would be more flexible to have in Metabase, but I wasn't sure how comfortable they might be doing it there. But I'm be happy to drop it here if Metabase would be a better place for it.

davereinhart: Some added context here: the request from Trevor was to add these to `shipping.scan_encounters_v2` which makes me think he's accessing the view directly. Since it was just adding columns to the view, I kept it as `scan_encounters_v1`.

lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
davereinhart: - [ ] Removed `vac_date` and `vac_date_2` from `string_questions`

schema/deploy/shipping/views.sql
davereinhart: Confirmed with Trevor that he would prefer to keep these calculations here since there are multiple use cases (dashboards, manuscripts) and having consistency for `vaccination_status`, `vaccine_manufacturer` and `fully_vaccinated_date` would be useful. He and two others reviewed the updated query and agreed with how those values are being calculated.

kschwabefry: Great! Thanks for checking. Re: the booster dose situation, Trevor answered that too. Posting here for posterity/ when we update for that too.

> I'd think to keep fully_vaccinated_date and later add a vaccine_boosted_date for 2 weeks post booster

kschwabefry (approved):

davereinhart:
Deployed.

PR #242, opened 2021-09-09
Point EH&S views solely at UW REDCap instance

seidelma:
The HCT project isn't working with both ITHS & UW REDCap instances, so point it only at the UW one.


kschwabefry (approved):

davereinhart:
It occurred to me that we could filter on the encounter identifier value instead of details _provenance. So using:
`WHERE encounter.identifier LIKE 'https://hct.redcap.rit.uw.edu/45/%/encounter_arm_1/%`
and 
`WHERE encounter.identifier LIKE 'https://hct.redcap.rit.uw.edu/45/%/enrollment_arm_1/`

Results should be the same, but might have better performance (especially is identifier field is indexed) and it would allow us to leave the provenance values as is for now.

seidelma:
The views now use a where-clause against `encounter.identifier` instead of `encounter.details`.

davereinhart (approved):

PR #243, opened 2021-09-10
Handle missing provenance workbook value in Slack post function

davereinhart:
With use of accessioning API, provenance data no longer contains a
`workbook` value. Updating function to avoid exception when workbook
value is null in shipping view.


kschwabefry (approved):

davereinhart:
Deployed.

PR #244, opened 2021-09-16
Skip HCT migrated records with no non-enrollment encounters

davereinhart:
We are postponing processing of enrollment records for the HCT migrated
records until the data issues with those record are resolved in REDCap. This is
causing warnings when there are DETs being processed for migrated records'
enrollment forms that do not have any non-enrollment forms filled out.

Each FHIR bundle with a patient resource must contain at least one encounter resource,
which typically would be covered because the enrollment itself counts as an encounter.
Because those are not being processed for migrated records there needs to be at least
one non-enrollment encounter to create a valid FHIR bundle. This change checks to make
sure that is the case before generating the FHIR bundle.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -158,6 +158,10 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
         system_identifier = INTERNAL_SYSTEM)
 
     persisted_resource_entries = [patient_entry, *location_resource_entries]
+    
+    REDCAP_RECORD_ID_CUTOFF = 32024
+    MIGRATED_RECORD = True if int(enrollment['record_id']) <= REDCAP_RECORD_ID_CUTOFF else False
+    migrated_record_encounter_found = False
 
     for redcap_record_instance in redcap_record_instances:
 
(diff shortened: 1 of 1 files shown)


lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -158,6 +158,10 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
         system_identifier = INTERNAL_SYSTEM)
 
     persisted_resource_entries = [patient_entry, *location_resource_entries]
+    
+    REDCAP_RECORD_ID_CUTOFF = 32024
+    MIGRATED_RECORD = True if int(enrollment['record_id']) <= REDCAP_RECORD_ID_CUTOFF else False
```
kschwabefry, line 163: Might be safer to use `int(enrollment.get('record_id'))` here

kschwabefry (approved):

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
davereinhart: Either will throw an Exception if 'record_id' isn't present. `int(enrollment['record_id'])` will throw a KeyError:
```
Keyerror: 'record_id'
```
whereas `int(enrollment.get('record_id'))` will throw TypeError:
```
int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```
I think the first one is more appropriate error in this case since there should always be a `record_id` on these.

davereinhart:
Deployed.

PR #245, opened 2021-09-23
Move check for HCT migrated record encounter present

davereinhart:
The boolean to indicate whether or not a non-enrollment encounter is present on
an HCT migrated record was happeneing too early, which was still resulting in
warnings that the FHIR bundle was incomplete because it did not contain any
encounter events (for example when daily attestation form was incomplete).

Setting the boolean flag to True only after checking that daily attestatation,
kiosk registration, or test order survey forms are complete, or that the
`testing_date` from internal testing determination form is not empty, since
those can result in the encounter record being skipped.


davereinhart:
Included this change in PR #246

PR #246, opened 2021-09-27
Add UW Dropbox as collection method and site

davereinhart:
ETL updates for unobserved tiny swabs received via dropboxes.


davereinhart:
- [ ] After deployment, generate and upload DETs for `husky_test_kit_registration` instrument starting from when tiny swab dropbox process was first implemented

schema/deploy/warehouse/site/data.sql
```diff
@@ -85,6 +85,7 @@ insert into warehouse.site(identifier, details)
         ('UWSuzzalloLibrary',                       '{"category": "community",  "type": "collegeCampus",
             "swab_site": "%suzzal+o"}'),
         ('UWReopeningSwabNSend',                    '{"category": "community",  "type": "uw-reopening"}'),
+        ('UWReopeningDropbox',                      '{"category": "community",  "type": "uw-reopening"}'),
```
kschwabefry, line 88: Does the lab record information about which specific dropbox the sample was collected from? There could be analytical benefits to distinguishing the sites, rather than lumping them all together, the same way we do for swab and send

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -194,6 +196,8 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
                 collection_method = CollectionMethod.KIOSK
             elif is_complete('test_order_survey', redcap_record_instance):
                 collection_method = CollectionMethod.SWAB_AND_SEND
```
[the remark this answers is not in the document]
davereinhart: Ok, I was wondering about that overlap as well, since presence of a huksy_test_kit_registration on its own may not be a reliable enough way to determine if it's coming from a dropbox. Maybe it would be worth checking the barcode against ID3C to see if it's in the tiny swabs unobserved set.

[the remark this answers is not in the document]
kschwabefry: (Discussed on the call, just adding a note here for visibility) Unfortuantely, the barcode set would be the same for both of them too. I'm not sure we have a better way to distinguish other than the absense of the test order form, though I wish we did!

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -289,7 +294,9 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
         specimen_received = (collection_method == CollectionMethod.SWAB_AND_SEND and \
             is_complete('post_collection_data_entry_qc', redcap_record_instance)) or \
             (collection_method == CollectionMethod.KIOSK and \
-            is_complete('kiosk_registration_4c7f', redcap_record_instance))
+            is_complete('kiosk_registration_4c7f', redcap_record_instance)) or \
```
seidelma, line 297: Let's take this opportunity to make the conditionals a little clearer with indentation. If we're not worried about 80-column widths, we can put each separate group-of-conditions (ie, collection_method == kiosk && is_complete(kiosk_registration)) on its own line, maybe?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -289,7 +294,9 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+            (collection_method == CollectionMethod.UW_DROPBOX and \
+            is_complete('huksy_test_kit_registration', redcap_record_instance))
```
seidelma, line 299: Looks like this should be `husky_test_kit_registration`.

seidelma (changes requested):

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
davereinhart: Thanks! Typo fixed.

davereinhart:
This update could result in the creation of `warehouse.sample` records that are never actually received. Because the kits received via dropbox only have `husky_test_kit_registration` form in REDCap, and there's no PCDEQC form filled out to confirm the sample was received, a Specimen entry will be included in the FHIR bundle if the `husky_test_kit_registration` form is marked complete in REDCap. That encounter-related specimen will then result in a sample record being created during the FHIR ETL:

https://github.com/seattleflu/id3c/blob/8e3f81ce07927ceb82425d1a8f79eaf13c62ca72/lib/id3c/cli/command/etl/fhir.py#L675

This sample record will have limited amount of data in ID3C since there will be no associated data from LIMS or Samplify, but still need to make sure that having it in the `warehouse.sample` table does not cause any issues and/or require further changes to shipping views (i.e. it should not appear in RoR or EHS reporting)

davereinhart:
Deployed.

davereinhart:
[Encounters are not being processed right now for HCT](https://trello.com/c/FROp2Z37/1206-encounters-are-not-being-processed-right-now-for-hct)

PR #248, opened 2021-09-29
Fix logic for checking collection method to set default site

davereinhart:


davereinhart:
Deployed.

davereinhart:
[Encounters are not being processed right now for HCT](https://trello.com/c/FROp2Z37/1206-encounters-are-not-being-processed-right-now-for-hct)

PR #249, opened 2021-09-30
UW Reopening ETL - only process updated event instances

davereinhart:


davereinhart:
Tabling this for now, due to increased amount of QC happening on the HCT REDCap project. There may be cases where modifications made to one repeat instance might require reprocessing of a separate one, such as when duplicate barcodes or multiple collection method (i.e. Kiosk + Dropbox) are being corrected.

PR #252, opened 2021-10-20
Reverse order of prioritized timestamps for kiosk specimen collection

davereinhart:
Reversing order of date/timestamps being used to determine specimen
collection date for inclusion in FHIR bundle. The new order will check
for the `nasal_swab_q` field from the instrument first, and the REDCap
built-in instrument completed timestamp second.

This is being done due to cases where the instrument is completed on a
different date, such as during QC/data corrections in REDCap. In those
cases the `nasal_swab_q` date will accurate and the instrument completed
timestamp will not be.


davereinhart:
Closed in favor of seattleflu/id3c#264

PR #256, opened 2021-10-26
Notify REDCap of bad addresses

cooperqmarshall:
When the get_geocoded_address() function from the geocode class returns
None, update the 'bad_address' field in redcap to yes. This can be used
to trigger a notification to the participant and or PC team for rapid
resolution.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py
@@ -2,6 +2,7 @@
 Process DETs for the greater Seattle Coronavirus Assessment Network (SCAN) REDCap projects.
 """
 import re
+import os
 import click
 import json
 import logging
@@ -258,7 +259,7 @@ def site_map(record_location: str) -> str:
     return location_site_map[record_location]
 
 
-def locations(db: DatabaseSession, cache: TTLCache, record: dict) -> list:
+def locations(db: DatabaseSession, cache: TTLCache, record: REDCapRecord) -> list:
     """ Creates a list of Location resource entries from a REDCap record. """
     lodging_options = [
         'shelter',
@@ -286,6 +287,8 @@ def locations(db: DatabaseSession, cache: TTLCache, record: dict) -> list:
 
     lat, lng, canonicalized_address = get_geocoded_address(address, cache)
     if not canonicalized_address:
+        if os.environ.get('GEOCODING_ENV') == 'production':
+            record.update_record({"valid_address": "no", "redcap_event_name": record["redcap_event_name"]})
         return []  # TODO
 
     tract_location = residence_census_tract(db, (lat, lng), housing_type)


seidelma:
It looks like the automated tests that happen on each `git push` have failed because:
> lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py:289: error: "Dict[Any, Any]" has no attribute "update_record"
Could you correct that and push it again?

cooperqmarshall:
The checks failed because `update_record()` is not in master yet. I have another PR open for that addition: https://github.com/seattleflu/id3c/pull/266

davereinhart:
@cooperqmarshall The failing check is due to the issue I mentioned yesterday of the `record` parameter being a `dict` instead of `REDCapRecord` object. That should be safe to change, but definitely needs testing.

truong97:
@davereinhart not sure if Cooper mentioned already but him and I tested the changes on the latest commit. We tested for both the `dict` and `REDCapRecord` (both are okay and did not throw an errors/bug). The testing gave us the expected outputs and changed the record's `valid address` from `yes` to `no` on REDCap. We decided to keep it as `dict` and create another PR to change to `REDCapRecord`, unless you see a reason not to. I can send you the testing script if needed. Note that locally, for both `id3c` and `id3c-customizations`, we did remove the DatabaseSession db and cache from the required parameters, enable to run `Location()` without them. Let me know if there is any other questions or comment!

cooperqmarshall:
@truong97 Apologies! I got pulled into a logistics item and forgot to send the update yesterday. @davereinhart If needed, I can provide more detail about what areas of the code we needed to remove so  `locations()` could run without the TTLCache and DatabaseSession args.

davereinhart:
Thanks @cooperqmarshall and @truong97!  Melissa mentioned you were thinking of chaning the record from dict to REDCapRecord in a separate PR.  I don't think that's neccessary, you can just include it here and that should allow the automated tests to succeed.  The only place the `locations` function is called is here: https://github.com/seattleflu/id3c-customizations/blob/d34ed2c852544191e583ea3ac9b9f8b0af735386/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py#L152 so you don't need to worry about it impacting anything else.

I'll take a look at the test script Melissa sent me (`update-one-record.py`) but ultimately we'll want to test this locally by running the SCAN ETL to make sure it's working as expected in that context.

cooperqmarshall:
@davereinhart Ok I will change the type from `dict` to `REDCapRecord` at `locations()` then recommit

cooperqmarshall:
The checks failed again because `update_record()` is not in id3c's master branch yet

davereinhart:
> The checks failed again because `update_record()` is not in id3c's master branch yet

Ok, this PR looks good to me: 
https://github.com/seattleflu/id3c/pull/266

Go ahead and merge that then we can re-run these tests.

cooperqmarshall:
Is there a way to rerun the tests? I can force push another commit to trigger the tests.

cooperqmarshall:
Some checks are failing:

cooperqmarshall:
All checks failed so I will change the type of the record arg back to `REDCapRecord`

kschwabefry:
@cooperqmarshall one thing that might help with development is that you can also run the test suite locally. See https://github.com/seattleflu/id3c-customizations#tests for details.

cooperqmarshall:
@kschwabefry Thanks for the tip!

davereinhart (changes requested):
This change is pretty unique in that it will be the only place in any ingestion ETL where we're writing back to REDCap. We want to be sure that this is only happening in production, and not during local development.

Fortunately, a new environment variable was recently added for another purpose that will work well here:
https://github.com/seattleflu/backoffice/commit/5ed2d166f91b38ac1a6077b5d72714570b0c961c

You can see how it's being used here, you'll want to do something similar:
https://github.com/seattleflu/id3c/blob/9375ada48bf05b051a202f11fea1bdf3bd8bd012/lib/id3c/cli/command/geocode.py#L286

cooperqmarshall:
@davereinhart So I should add the check `if environ.get('GEOCODING_ENV') == 'production'` before calling `update_record()`?

davereinhart:
> @davereinhart So I should add the check `if environ.get('GEOCODING_ENV') == 'production'` before calling `update_record()`?

Exactly!

cooperqmarshall:
@davereinhart Added!

davereinhart (approved):

truong97:
Deployed.

PR #258, opened 2021-11-03
ETL: REDCap DET - HCT process migrated enrollments and drop unused fields

davereinhart:
Remove filter on migrated enrollments to allow those records and instruments to be ingested from the new HCT REDCap project. This filter was originally implemented when issues were identified with the data migration of records from the old REDCap project to the new one. Now that data cleanup and QC has occurred, this filter can be safely removed.

Also removing fields no longer being used in REDCap from the ETL process. These fields accounted for some of the differences between enrollment records in the old and new projects. Those REDCap fields that are being removed will subsequently be dropped from existing HCT records in ID3C as well as any shipping views that reference them.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -168,7 +168,7 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
 
         if redcap_record_instance.event_name == ENROLLMENT_EVENT_NAME:
             event_type = EventType.ENROLLMENT
-            check_enrollment_data_quality(redcap_record_instance)
+            #check_enrollment_data_quality(redcap_record_instance)
 
         elif redcap_record_instance.event_name == ENCOUNTER_EVENT_NAME:
             event_type = EventType.ENCOUNTER
@@ -472,8 +472,8 @@ def create_enrollment_questionnaire_response(record: REDCapRecord, patient_refer
     # Do not include `core_age_years` because we calculate the age ourselves in the computed questionnaire.
     integer_questions = [
         'weight',
-        'height_total',
-        'tier'
+        #'height_total',
+        #'tier'
     ]
 
     string_questions = [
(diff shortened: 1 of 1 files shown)


davereinhart:
[Filter out ID3C alerts for HCT Unrecognized Barcodes - participant driven error](https://trello.com/c/IsyHIiss/1218-filter-out-id3c-alerts-for-hct-unrecognized-barcodes-participant-driven-error)

davereinhart:
[HCT: Process migrated enrollments from new redcap project in ETL](https://trello.com/c/k9qn4fmv/1227-hct-process-migrated-enrollments-from-new-redcap-project-in-etl)

kschwabefry:
@davereinhart changes look good to me. My only lingering thought is I was originally thinking we would change the shipping views to exclude those variables, prior to this change to drop them from the ETL. But, if this is deployed before shipping view changes, and we drop all the data for certain variables (on_campus_2x_week, tier, housing_group, etc.), do views like `shipping.__uw_priority_queue_v1`, `shipping.uw_reopening_enrollment_fhir_encounter_details_v1`, `shipping.uw_reopening_encounters_v1` still work correctly, (just with empty values for those variables)? If that all still works, then we should be good to go.

kschwabefry:
Also, I went ahead and checked some of the dropped variables against active [metabase queries](https://backoffice.seattleflu.org/metabase/question/320) to see if we needed to notify any users, but I don't see them being used, so at least from that point of view, there may not be so much impact to downstream analytics consumers if these are removed. (That doesn't include direct id3c queries, but at least covers users who use metabase).

davereinhart:
> uw_reopening_enrollment_fhir_encounter_details_v1

Yes, the way the views are designed, if the value isn't in the FHIR bundle, the corresponding row won't be in the materialized view `shipping.fhir_questionnaire_responses_v1`, and if it's not found there, then the column will be null in `shipping.uw_reopening_enrollment_fhir_encounter_details_v1` view.  That view is referenced from `uw_reopening_encounters` and `__uw_priority_queue_v1` so null values will appear there as well.

kschwabefry (approved):

seidelma (approved):
Ran it on my dev box and it looks good to me. To reduce tech debt, I would like to see the stuff that's commented out (e.g. the call to `check_enrollment_data_quality()`, and removed fields) just deleted completely or a comment added as to why they're disabled.

davereinhart:
Deployed.

davereinhart:
> Ran it on my dev box and it looks good to me. To reduce tech debt, I would like to see the stuff that's commented out (e.g. the call to `check_enrollment_data_quality()`, and removed fields) just deleted completely or a comment added as to why they're disabled.

Thanks, Todd, good suggestion! I added some comments and I agree we should just drop these lines once these fields and values are fully purged from ID3C.

PR #260, opened 2021-11-05
Check barcode before including specimen entry in FHIR bundle

davereinhart:
Adding a check to determine if a barcode is valid before creating a
specimen entry in the FHIR bundle. This will prevent downstream warnings
in the FHIR ETL when an invalid barcode is encountered.

This is being implemented on UW reopening only for now due to the higher number
of invalid barcodes received and QC process transferring to LIMS.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
+++ b/lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
@@ -6,6 +6,7 @@ from dateutil.relativedelta import relativedelta
 from enum import Enum
 from id3c.cli.command.etl import redcap_det
 from id3c.cli.redcap import is_complete, Record as REDCapRecord
+from id3c.db import find_identifier
 from seattleflu.id3c.cli.command import age_ceiling
 from .redcap import *
 
@@ -272,6 +273,7 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
  
         specimen_entry = None
         specimen_observation_entry = None
+        specimen_identifier = None
         specimen_received = (collection_method == CollectionMethod.SWAB_AND_SEND and \
             is_complete('post_collection_data_entry_qc', redcap_record_instance)) or \
             (collection_method == CollectionMethod.KIOSK and \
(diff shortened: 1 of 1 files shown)


lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -307,22 +309,40 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
-                encounter_reference = initial_encounter_reference)
+            # Check if specimen barcode is valid
+            specimen_barcode = None
+            for barcode in prioritized_barcodes:
+                specimen_barcode = barcode.strip()
+                if specimen_barcode:
+                    # Disable logging when calling find_identifier here to suppess alerts
+                    # for HCT unrecognized barcodes which are being managed in LIMS
+                    logging.disable(logging.WARNING)
+                    specimen_identifier = find_identifier(db, specimen_barcode)
```
kschwabefry, line 320: Ah, clever approach - I think this should work ok! It opens up a small probability of missing other warnings from that method, if any, but since its log output and not a specific exception we could catch, I can see why you took this approach, and I think it will succeed in quieting the warnings.

Some alternative options would be to write the database lookup here directly, without using find_identifier - and then you could leave out warnings, or you could modify find_identifier() to take an optional param to disable the warning, but maybe with a default to warn.

davereinhart: I see what you mean, but I'm tempted to leave as is. The `find_identifier` function only has two possible results, the warning we're trying to suppress is the only one that's ever logged if the identifier isn't found:
https://github.com/seattleflu/id3c/blob/1d236e4e4d2b335514721ab2b4ef4861c584b2e0/lib/id3c/db/__init__.py#L94

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_reopening.py
```diff
@@ -307,22 +309,40 @@ def redcap_det_uw_reopening(*, db: DatabaseSession, cache: TTLCache, det: dict,
...
+                    able_to_test = redcap_record_instance['able_to_test'],
+                    system_identifier = INTERNAL_SYSTEM)
+
+                specimen_observation_entry = create_specimen_observation_entry(
+                    specimen_reference = specimen_reference,
+                    patient_reference = patient_reference,
+                    encounter_reference = initial_encounter_reference)
+            else:
+                LOG.debug(f"No identifier found for barcode «{specimen_barcode}»")
+                LOG.info("Creating encounter for record instance without sample")
```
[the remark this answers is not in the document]
davereinhart: Good point!  Updated to indicate no identifier was found.

kschwabefry (approved):

davereinhart:
Deployed.

PR #263, opened 2021-11-10
Remove comma from reorder string check

davereinhart:
Since non-alphanumeric characters are being removed in the previous line, removing
comma from this string comparison.


davereinhart:
Related to #262 . Deployed.

PR #265, opened 2021-11-18
Handle unknown test results more gracefully

seidelma:
Add UnknownTestResult exception a la UnknownSampleOrigin/UnknownHospitalDischargeDisposition to gracefully handle bad test result values without dying from the uncaught Exception.

Prior logic said that if a test result isn't one of a very narrow set of values (e.g. 'detected', 'positive', 'negative', 'not detected'), just skip it. This will do the same thing, for all values that are not one we can confidently map to True or False.


davereinhart (approved):
This looks change good to me. Still not sure how to best handle these when we are alerted to them, but agree it would be best not to fail the entire job since it's coming up more often now.

PR #266, opened 2021-11-18
Update regex on clinical parse collapse_columns function

davereinhart:
Regex pattern on the collapse_columns function was incorrectly
defined allowing the last character of the `stub` parameter to be
optional. This had the effect of unintentionally collapsing `census_tract`
column when trying to collapse `race` columns because both contain `rac`.

The updated pattern requires all characters of the `stub` parameter to
be present in the column name.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/clinical.py
@@ -417,6 +417,12 @@ def collapse_columns(df: pd.DataFrame, stub: str, pid='enrollid') -> pd.DataFram
     encoding of multiple race options into a single array in a resulting
     column called "Race". Removes the original "Race*" option columns. Returns
     the new DataFrame.
+
+    >>> df = pd.DataFrame(columns = ['enrollid', 'racewhite', 'raceblack', 'raceasian', 'census_tract'])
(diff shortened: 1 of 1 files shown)


davereinhart:
[Ingest new KP clinical data](https://trello.com/c/csIuvthR/355-ingest-new-kp-clinical-data)

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -418,7 +418,7 @@ def collapse_columns(df: pd.DataFrame, stub: str, pid='enrollid') -> pd.DataFram
     column called "Race". Removes the original "Race*" option columns. Returns
     the new DataFrame.
     """
-    stub_data = df.filter(regex=f'{pid}|{stub}*', axis='columns')
+    stub_data = df.filter(regex=f'{pid}|{stub}.*', axis='columns')
```
kschwabefry, line 421: Good bug find/fix! I think it would be a great improvement to add a doctest here as well if possible.

davereinhart: Good suggestion, I need to get in the habit of that! Added and ready for review.

kschwabefry (approved):

PR #267, opened 2021-11-24
UW retrospectives - ingest new fields

davereinhart:
We are reactivating the REDCap SFS Clinical Data Pulls on the UW/HMC retrospective samples.
The project in REDCap pulls EMR data to give us Metadata on these samples. As part of this process we
are also adding new fields for ingestion into ID3C, requiring these updates to the ETL.

Additional fields to add:
- race
- ethnicity
- ICD10_primary
- ICD10_secondary
- covid_status_1
- covid_status_2
- covid_vaccine
- covid_date_1
- covid_date_2
- flu_status
- flu_date

Race and ethnicity questions are being added to the FHIR [QuestionnaireResponse](http://www.hl7.org/fhir/questionnaireresponse.html) entry, ICD-10 codes are being included as [Condition codes] (http://www.hl7.org/fhir/condition.html), and vaccine data is being added as [immunization entries](https://www.hl7.org/fhir/immunization.html)


davereinhart:
[Reprocess DETs for SFS Clinical Data Pulls - UW Retros & Add in additional fields for ingestion](https://trello.com/c/IF9NYjpM/1226-reprocess-dets-for-sfs-clinical-data-pulls-uw-retros-add-in-additional-fields-for-ingestion)

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -202,6 +202,8 @@ def create_encounter(db: DatabaseSession,
     encounter_class = create_encounter_class(record)
     encounter_status = create_encounter_status(record)
 
+    encounter_reason_codes = parse_icd10_codes(record, ('icd10_primary', 'icd10_secondary'))
```
kschwabefry, line 210: ICD10 codes is an interesting one to bring in. In the FHIR docs, I'm seeing them used for various things related to billing/claims. Did you find prior art for bringing them in as encounter reason code? On the encounter, it might also be a good fit for diagnosis, but I'm not sure what is the most standard way to approach incorporating these into the FHIR bundle. http://www.hl7.org/FHIR/encounter-definitions.html#Encounter.reasonCode

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -469,6 +474,100 @@ def create_encounter_status(redcap_record: dict) -> str:
     return mapper[standardized_status]
 
 
+def create_immunization(record: dict, patient_reference: dict) -> Optional[list]:
+    """ Returns a FHIR Immunization resource entry """
+    immunization_entries = []
```
kschwabefry, line 479: Good find. immunization records seems like a great approach here for storing the covid vaccine information for the participant.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -453,6 +453,9 @@ def determine_questionnaire_items(record: dict) -> List[dict]:
     if record["age"]:
         items["age"] = [{ 'valueInteger': age_ceiling(int(record["age"]))}]
 
+    if record["race"]:
+        items["race"] = [{ 'valueString': race(record["race"])}]
```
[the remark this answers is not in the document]
davereinhart: Unfortunately, it's still just open text in the REDCap codebook. The excel file that was provided has race in a single column, and does has values where there are multiple, comma-separated values, but I looked up a couple of the same records in REDCap and only a single race value was imported.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -362,6 +362,31 @@ def discharge_disposition(redcap_record: dict) -> Optional[str]:
 
     return mapper[standardized_disposition]
 
+def ethnicity(redcap_record: dict) -> Optional[str]:
```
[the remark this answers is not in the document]
davereinhart: Good idea! I'll move it.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
davereinhart: No, I didn't have any previous reference for this, and I wasn't quite sure where else to put them. Most of the ones I looked up were symptoms, so I though encounterCode would be a logic place. Now that I'm looking at FHIR docs, it looks like `ChargeItem.reason` may be a better place to put them: https://build.fhir.org/valueset-icd-10.html (build version of FHIR has a more complete list of codes, but current version does not: http://hl7.org/fhir/valueset-icd-10.html)

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
davereinhart: Ok, we'll need to make some related changes to the FHIR ETL to get these included in `individual.details`, and to shipping views to parse them out.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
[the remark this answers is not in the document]
davereinhart: Codebook is being updated so this won't be open text. I'll wait to see how it's defined there before finalizing this piece.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
davereinhart: After further review, it looks like `Condition.code` is another (and maybe the best) option. I found a couple of references outside of FHIR documentation where it's being used that way, and that location would make more sense than `ChargeItem.reason`:

https://fhir.cerner.com/millennium/dstu2/general-clinical/condition/
https://mednet.swiss/fhir/StructureDefinition-mni-familyMemberHistory.html

kschwabefry: `Condition.code` looks great - good find.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -460,6 +463,115 @@ def create_encounter_status(redcap_record: dict) -> str:
...
+        # Validate vaccine name and determine CVX code
+        vaccine_code = None
+        if vaccine_name in vaccine_mapper:
+            vaccine_code = vaccine_mapper[vaccine_name]
+        else:
+            raise UnknownVaccine (f"Unknown vaccine «{vaccine_name}».")
+
+        # Standardize date format
+        immunization_date = record[column_map["date"]]
+        #immunization_date = datetime.strptime(immunization_date_str, "%m/%d/%Y") if immunization_date_str else None
```
kschwabefry, line 559: Can this be dropped?

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -460,6 +463,115 @@ def create_encounter_status(redcap_record: dict) -> str:
...
+        
+        if immunization_status == "y" and vaccine_code:
+            immunization_identifier_hash = generate_hash(f"{record['mrn']}{vaccine_code['code']}{immunization_date}".lower())
+            immunization_identifier = create_identifier(f"{SFS}/immunization", immunization_identifier_hash)
+        
+            immunization_resource = create_immunization_resource(
+                patient_reference = patient_reference,
+                immunization_identifier = [immunization_identifier],
+                immunization_date = immunization_date, #datetime.strftime(immunization_date, "%Y-%m-%d") if immunization_date else None,
```
kschwabefry, line 568: Can this be dropped too?

kschwabefry (approved):

kschwabefry:
Nice work @davereinhart . Left a few minor comments, but this looks good to me.

lib/seattleflu/id3c/cli/command/etl/__init__.py
```diff
@@ -137,6 +137,33 @@ def standardize_race(race):
     return list(map(standardize_race, races))
 
 
+def ethnicity(redcap_record: dict) -> Optional[str]:
```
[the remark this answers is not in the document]
davereinhart: @kschwabefry Good thought, I changed it to string parameter. In looking into ethnicity further, I realized that it should be using a coding system (`valueCoding`) instead of `valueString`, so I modified the ethnicity function to return a dict with the appropriate FHIR ethnicity coding. 

Looking at our other redcap_det ETLs, `ethnicity` typically gets ingested as a boolean value (most likely because that's how it's defined in REDCap) but I think those should all be updated use coded values like I have here. That will also require updates in shipping views, so if it's better to keep it a boolean for now (for consistency) I can update my PR to remove the coding.

davereinhart:
Related to https://github.com/seattleflu/id3c/pull/276

lib/seattleflu/id3c/cli/command/etl/__init__.py
```diff
@@ -133,6 +137,58 @@ def standardize_race(race):
...
+    # to tackle this at a later date.
+    # -drr 2021-12-30
+
+    #hispanic_or_latino = create_coding("http://hl7.org/fhir/v3/Ethnicity", "2135-2", "hispanic or latino")
+    #not_hispanic_or_latino = create_coding("http://hl7.org/fhir/v3/Ethnicity", "2186-5", "not hispanic or latino")
+
+    mapper = {
+        "hispanic or latino":               True,           # hispanic_or_latino,
+        "not hispanic or latino":           False,          # not_hispanic_or_latino,
+        "white/caucasian":                  False,          # not_hispanic_or_latino,
```
kschwabefry, line 180: I'm not sure if this came from pre-existing code, but if we explicitly map this, I would code this as `None`, as I think its a race category, not ethnicity.

davereinhart: > I'm not sure if this came from pre-existing code, but if we explicitly map this, I would code this as `None`, as I think its a race category, not ethnicity.

That came from the "Codebook" they provided us (which is an Excel document, not actually defined in REDCap):
https://seattle-flu-study.slack.com/archives/C0173B4S1LY/p1639175618340200?thread_ts=1638557927.266100&cid=C0173B4S1LY

It might be safe to just drop it value since it doesn't look like it has been used in REDCap based on Cooper's summary.

kschwabefry: Nod. We could confirm with the folks analyzing data downstream, like Amanda or Cecile, but based on my understanding, if we want this to be standardized across projects, we should just map that as `None`

davereinhart: Removed that value from the map, since it's not present in REDCap.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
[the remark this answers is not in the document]
davereinhart: Defining codebook in REDCap is on hold indefinitely.

davereinhart:
Deployed.

PR #271, opened 2021-12-16
shipping.__uw_priority_queue_v1: reintroduce HCT surveillance testing

kschwabefry:
HCT study design is changing to reintroduce surveillance testing.
Add this back into the eligibility criteria for inclusion in
the shipping.uw_priority_queue_v1 view.

Also, reduce the delay for post-positive test invation from 90 days
to 14 days, due to potential for sooner reinfection, per study team.

- [x] tag schema


schema/deploy/shipping/views.sql
```diff
@@ -3474,6 +3475,34 @@ create or replace view shipping.__uw_priority_queue_v1 as (
...
+            5 as priority,
+            'surveillance' as priority_reason,
+
+            latest_positive_hcov19_collection_date,
+            latest_prior_test_positive_date,
+            prior_test_positive_date_base,
+            alerts_off
+        from uw_enrollments
+        -- Filter to participants who come to campus
+        where (on_campus_freq != 'not_on_campus')
```
davereinhart, line 3499: Should this include `AND on_campus_freq IS NOT NULL`?

kschwabefry: It should already be excluding rows where `on_campus_freq` is NULL, because the `'somestring' != NULL` (and for that matter `'somestring' = NULL` will both evaluate to null. I confirmed with Jess yesterday thats what she wants. 

If you think it helps readability, I can add that clause to make it more explicit.

davereinhart: Ok, that makes sense, no need to add more code if they're already excluded.

davereinhart (approved):

kschwabefry:
[Add back surveillance testing for HCT](https://trello.com/c/saGiRaBA/1276-add-back-surveillance-testing-for-hct)

PR #274, opened 2021-12-29
Redesign of shipping.sample_with_best_available_encounter_data_v1

davereinhart:
This view suffers from poor performance due to the way it is using regex
to match samples to sites. To reduce the number of rows that require regex
matching, this updated version joins the data in three steps, first with
exact matches on site id, then on LIKE matches, then finally on regex matches,
with each iteration only joining to the records that haven't already found a
match.


davereinhart:
[Investigate Metabase Latency](https://trello.com/c/WrRgpuWH/1280-investigate-metabase-latency)

schema/deploy/shipping/views.sql
```diff
@@ -68,74 +68,85 @@ drop view if exists shipping.sample_with_best_available_encounter_data_v1;
...
+    samples_with_site_match as (
+      select samples_with_manifest_data.sample_id,
+        samples_with_manifest_data.sample,
+        samples_with_manifest_data.has_encounter_data,
+        samples_with_manifest_data.best_available_encounter_date,
+        samples_with_manifest_data.site_manifest_details,
+        case
+            when samples_with_manifest_data.best_available_encounter_date < '2019-10-01'::date then 'Y1'::text
+            when samples_with_manifest_data.best_available_encounter_date < '2020-11-01'::date then 'Y2'::text
+            when samples_with_manifest_data.best_available_encounter_date < '2021-11-01'::date then 'Y3'::text
```
truong97, line 103: @davereinhart Would it be worthwhile to add Y4 to the season?

davereinhart: Yes, great suggestion! I'll add that now.

truong97 (approved):

davereinhart:
Deployed.

PR #275, opened 2021-12-30
Update to redesign of shipping.sample_with_best_available_encounter_data_v1

davereinhart:
A recent rework of `shipping.sample_with_best_available_encounter_data_v1` (#274) incorrectly
applied `similar_escape` when using a LIKE comparison. This did not impact the results
of the query because it was followed up with a regex comparison that caught anything not
already matched. This change yields the same result but performs better since there
are fewer records that need to matched using regex.


truong97:
@davereinhart Looks good and makes sense! 

Is there a reason why ~~ was used over ~~* (it is unlikely that there is weird capitalization within swab sites but possible edge case?) or would the regex match cover that case without compromising performance?

davereinhart:
> @davereinhart Looks good and makes sense!
> 
> Is there a reason why ~~ was used over ~~* (it is unlikely that there is weird capitalization within swab sites but possible edge case?) or would the regex match cover that case without compromising performance?

Good observation! The way it was originally written assumes the `sample.details ->> 'sample_origin'` is already lowercase, which should be the case since those values are coming from LIMS, but it wouldn't hurt to ignore case just to be sure. Since the `site_details.manifest_regex` values are already lowercase, lowercasing the `samples_with_site_match.site_manifest_details` value before comparing might perform better than ILIKE (`~~*`). I'll try that out and see if there's any difference in results or performance.

truong97 (approved):

davereinhart:
Deployed.

PR #277, opened 2022-01-07
Update shipping views to drop unused fields

bencap:
This change removes references to the fields
```
height_total
tier
uw_greek_member
uw_housing_resident
uw_housing_group_(a-g)
on_campus_2x_week
bmi
```
from the shipping views `uw_reopening_enrollment_fhir_encounter_details_v1` and 
`uw_reopening_encounters_v1`. These fields were unused in the ETL since #258.


seidelma (approved):
LGTM, I took a peek at #258, and these changes do seem to be all that's required. 

Two minor things I noticed: the date comment of the new schema tag in `schema/sqitch.plan` seems to be wrong ("01 January 2022" instead of "07 January 2022"). Since there isn't a merge conflict I think it'll be okay, just potentially confusing if someone's trying to reconstruct the history. Assuming sqitch actually supports that.

Also, your sqitch email setting is set to your laptop's local address, which is probably not what you want. If you end up doing any development using Vagrant or other environments, using a different user/email will pollute the sqitch history. Again, probably not a big deal since the `sqitch.plan` file is just text.

bencap:
@seidelma Thanks for looking and noting those things. I just went in and changed my username and email configurations so it shouldn't be an issue in the future. Will go ahead with merging/deploying these.

bencap:
deployed

PR #280, opened 2022-01-25
Handle missing REDCap instrument when checking completeness

davereinhart:
The `is_complete` function will return `None` if the given instrument
does not exist. Updating in two ETLs where it was comparing the result to
`False` so that it will also handle `None` result appropriately.


kschwabefry (commented):
Makes sense, I don't see a reason not to treat `none` and `false` the same in these cases. Would this impact any existing records and would we want to do any retroactive data cleanup?

davereinhart:
> Makes sense, I don't see a reason not to treat `none` and `false` the same in these cases. Would this impact any existing records and would we want to do any retroactive data cleanup?

Yes, this seems to be responsible for at least some of the individual records that are not linked to an encounter. There's nothing in the individual table to indicate what the data source is, so they have to be traced back to `receiving.fhir` based on individual identifier, and then to `receiving.redcap_det` to see where they originated.  My thought was to get this deployed so no new ones are being added, then go back and clean up any got created due to the bad checks.

kschwabefry (approved):

davereinhart:
Deployed.

PR #282, opened 2022-01-29
Modify shipping view uw_reopening_encounters_v1 to use inner join

davereinhart:
The uw-ehs-report job has in some cases pushed empty values to REDCap EHS
project when values do exist in ID3C. This occurred most recently during
a sqitch deployment that reworked shipping views, during which the uw-ehs-report
and refresh materialized views cron jobs were running.

While not able to recreate this same behavior locally, the most likely issue is
with the materialized view that these fields are orginating from. To prevent
these from being empty, using an inner join to `uw_reopening_enrollment_fhir_encounter_details_v1`
still returns the same number of records (at time of writing), and will ensure that
the `uw_reopening_encounters_v1` view is not returning mostly null values
when it shouldn't be.


davereinhart:
[Investigate and correct EH&S job](https://trello.com/c/Lvbqouxy/1320-investigate-and-correct-ehs-job)

kschwabefry (approved):

davereinhart:
Deployed.

PR #283, opened 2022-02-02
SCH Retro Additional Field Ingestion

truong97:
The fields, `admit_during_this_encounter` and `admit_to_icu` from Seattle Children's are added for ingestion. These fields gets parsed and ingested, then stored under `details` -> `responses` in the warehouse.encounter table as either `yes` or `no` responses.
Additionally,financial aid and self-pay values were added to the Insurance mapping.

Testing Data Used:
1. SFS_Year1_Main_Data_20210922
2. SFS_Year2_Main_Data_20210923
3. SFS_Year3_Main_Data_20211207
4. SFS_Year4_Main_Data_20220124
5. SFS_Year4_Main_Data_20220208
6. SFS_Year4_Main_Data_20220209

Testing Done Include for All 6 Test Data Under Local:
1. id3c clinical parse-sch
2. id3c clinical upload
3. id3c etl clinical

Testing went well where the fields showed up as expected and the new insurance values were parsed. Y1, Y2, Y3, and Y4 all have the 2 additional fields, thus no pre-processing needed.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/etl/clinical.py
@@ -30,6 +30,9 @@ from id3c.cli.command.etl import (
     UnknownCovidShotResponseError,
     UnknownCovidShotManufacturerError,
     UnknownSiteError,
+    UnknownAdmitEncounterResponseError,
+    UnknownAdmitICUResponseError,
+
 )
 from . import race
 
@@ -215,7 +218,9 @@ def encounter_details(document: dict) -> dict:
                 "FluShot": flu_shot(document.get("FluShot")),
                 "AssignedSex": [sex(document.get("AssignedSex"))],
                 "HispanicLatino": hispanic_latino(document.get("HispanicLatino")),
-                "MedicalInsurance": insurance(document.get("MedicalInsurance"))
+                "MedicalInsurance": insurance(document.get("MedicalInsurance")),
+                "AdmitDuringThisEncounter": admit_encounter(document.get("AdmitDuringThisEncounter")),
+                "AdmitToICU": admit_icu(document.get("AdmitToICU")),
             },
         }
 
@@ -228,7 +233,7 @@ def encounter_details(document: dict) -> dict:
     for k in ["CovidShot1", "CovidShot2"]:
         if k in document:
             details["responses"][k] = covid_shot(document[k])
-    
+
     if "CovidShotManufacturer" in document:
         details["responses"]["CovidShotManufacturer"] = covid_shot_maunufacturer(document.get("CovidShotManufacturer"))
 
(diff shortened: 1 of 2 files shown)


truong97:
[SCH Retrospectives - Additional fields to add for ingestion and investigate Data QC on different IDs](https://trello.com/c/UYYX5XKO/1298-sch-retrospectives-additional-fields-to-add-for-ingestion-and-investigate-data-qc-on-different-ids)

bencap (approved):
These changes look good to me and all seems to be in order when testing locally, with the fields being ingested into my local database. Still may want a senior dev to sign off on them as well though :)

truong97:
For adding new mapping to insurance name, `financial aid` and `self`:

Testing Data Used:

- SFS_Year4_Main_Data_20220124
- SFS_Year4_Main_Data_20220208
- SFS_Year4_Main_Data_20220209

Testing Done Include All 3 Test Data Under Local Production:

- id3c clinical parse-sch
- id3c clinical upload
- id3c etl clinical

Testing went well where the fields, including `admit_during_this_encounter` and `admit_to_icu` showed up as expected. No errors were thrown when new values adding to insurance name mapping.

lib/seattleflu/id3c/cli/command/etl/clinical.py
```diff
@@ -215,7 +215,9 @@ def encounter_details(document: dict) -> dict:
                 "FluShot": flu_shot(document.get("FluShot")),
                 "AssignedSex": [sex(document.get("AssignedSex"))],
                 "HispanicLatino": hispanic_latino(document.get("HispanicLatino")),
-                "MedicalInsurance": insurance(document.get("MedicalInsurance"))
+                "MedicalInsurance": insurance(document.get("MedicalInsurance")),
+                "AdmitDuringThisEncounter": document.get("admit_during_this_encounter"),
+                "AdmitToICU": document.get("admit_to_icu"),
```
davereinhart, line 220: For consistency, these values should probably be converted to "yes/no" using a function (similar to `FluShot`).

davereinhart (changes requested):

truong97:
Checks are currently failing b/c https://github.com/seattleflu/id3c/pull/284 needs to be reviewed and approved first.

davereinhart (approved):
Looks good. 

It might be possible to combine the `admit_encounter` and `admit_icu` mapping functions into a single function, which would work well if these two remain consistent going forward, but keeping them separate will be better in case they diverge.

truong97:
The checks were failing, even after deploying https://github.com/seattleflu/id3c/pull/284. I realized that they were because the tests were calling on the wrongly-named functions upon invoking `>>>` inside the comment block (I was calling the parameter instead of the function itself - big oof). There was no actual changes to the code - just changes to the comment block under `admit_encounter()` and `admit_icu()`. This will be corrected before deployment. 

cc @davereinhart, so that you are aware.

lib/seattleflu/id3c/cli/command/etl/clinical.py
```diff
@@ -304,6 +309,78 @@ def flu_shot(flu_shot_response: Optional[Any]) -> list:
...
+    >>> admit_encounter(0.0)
+    ['no']
+
+    >>> admit_encounter('TRUE')
+    ['yes']
+
+    >>> admit_encounter('maybe')
+    Traceback (most recent call last):
+        ...
+    id3c.cli.command.etl.UnknownAdmitEncounterResponseError: Unknown admit during this encounter response «maybe»
```
davereinhart, line 325: This string needs to match exactly with that the function is returning below (
"Unknown admit during this encounter response " doesn't match "Unknown admit during encounter response ")

lib/seattleflu/id3c/cli/command/etl/clinical.py
```diff
@@ -304,6 +309,78 @@ def flu_shot(flu_shot_response: Optional[Any]) -> list:
...
+    >>> admit_icu(0.0)
+    ['no']
+
+    >>> admit_icu('TRUE')
+    ['yes']
+
+    >>> admit_icu('maybe')
+    Traceback (most recent call last):
+        ...
+    id3c.cli.command.etl.UnknownAdmitICUResponseError: Unknown admit to icu response «maybe»
```
davereinhart, line 362: This string also needs to match what is being returned below (case-sensitive).

davereinhart (changes requested):

truong97:
Deployed.

PR #287, opened 2022-02-09
Updating reportable_conditions_v1 to use values

truong97:
The Reportable Conditions CRON job used to take 4 seconds, but is now taking 39-40 minutes. Upon investigation, the job can run faster if we add `values` to reportable_conditions_v1 view.

Testing done:

- ran runtime test prior to changes to reportable_conditions_v1 on local (deployed on local)
- ran runtime test after to changes to reportable_conditions_v1 on local (revert on local)
- compared the two and the new changes are faster


truong97:
[Investigate Reportable Conditions CRON job](https://trello.com/c/Pc0bvUwp/1327-investigate-reportable-conditions-cron-job)

davereinhart (approved):

truong97:
Deployed as of Feb 10th 2022

PR #288, opened 2022-02-23
Add Cascadia Identifiers to Backfill Sample IDs

bencap:
When the Cascadia identifier set was created, the identifier set
was not added to the `backfill-sample-ids` job. This change adds
that identifier set to the job so that samples with `null` identifiers get
backfilled for this collection. I'll also go ahead and add some 
documentation around this task on the barcodes page so that it is 
clearer you need to do this when new collection sets for tinys are created.

This was tested by running `backfill-sample-ids` locally and making
sure that samples that were a part of the `collections-cascadia-tiny-swabs-home`
got pulled in to be backfilled. Since there are currently no Cascadia
samples, I just created one with a `null` identifier and tested that
it got pulled in to be backfilled, which it did.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/backfill_sample_ids.py
+++ b/lib/seattleflu/id3c/cli/command/backfill_sample_ids.py
@@ -25,8 +25,9 @@ def backfill_sample_ids(*, db: DatabaseSession, action: DatabaseSessionAction):
             join warehouse.identifier on sample.collection_identifier = identifier.uuid::text
             join warehouse.identifier_set using (identifier_set_id)
             where identifier is null and
-            identifier_set.name in ('collections-uw-tiny-swabs-home', 'collections-uw-tiny-swabs-observed', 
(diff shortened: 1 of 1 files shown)


bencap:
[Add cascadia barcodes to backfill sample ids](https://trello.com/c/4FR9Ae6K/1331-add-cascadia-barcodes-to-backfill-sample-ids)

truong97 (approved):
Looks good! Trello Card mentioned maybe converting the query to a view, were you planning on doing that as well?

bencap:
@truong97 I was still thinking about it. I guess I am not super clear on the benefits of a view over just doing it like this - is it just for the simplicity of having the backfill query in a spot with our other queries vs. in `backfill_sample_ids.py`? For now I just added some documentation on the barcodes page since it seemed like the main issue, separate from view vs. query, is just remembering to do this for collection sets that use tiny swabs in the first place.

truong97:
Gotcha, yeah I was thinking the same thing! I don't think runtime will be quicker if it did end being a view. But yes, I did see you add the documentation, so that should suffice.

kschwabefry (approved):

bencap:
Deployed

PR #289, opened 2022-02-24
Rework gitignore to function as allowlist

davereinhart:
Reworking of .gitignore files to exclude everything up front, with
exceptions for files and folder that should be included. This is to
reduce the chance of pushing any files unintentionally.

As part of this rework, a `schema/revert/templates` folder and files were found
that should not be there. They already exist in `schema/templates`
where they belong, so are being dropped from `schema/revert`.


davereinhart:
[RNI followup from linelist on github](https://trello.com/c/DpFsHAGb/1326-rni-followup-from-linelist-on-github)

bencap (commented):
One question: Is the `/**/` applying the gitignore instructions to all subdirectories of the directory in which the gitignore is in?

davereinhart:
> One question: Is the `/**/` applying the gitignore instructions to all subdirectories of the directory in which the gitignore is in?

Yes, that's correct. For example with this configuration, `.sql` files are ignored everywhere except for `/schema/deploy/`, `/schema/revert/`, and `/schema/verify/`.

kschwabefry (approved):

PR #290, opened 2022-02-25
Disable parallel queries and increase memory for offer-uw-testing

davereinhart:
This job has been encountering issues related to the select from
__uw_priority_queue_v1 hanging indefinitely. In those cases, the parallel
workers are in different states, some stuck with a wait event of `IPC:
MessageQueueSend` and others being spawed and killed repeatedly.

This change will disable parallel queries when selecting from this
particular view, and increase memory to avoid writing to disk for sorting
and hash tables. The performance is comparable with these changes tested locally.
This may not resolve the issue entirely, but should at least eliminate parallelism
or memory as the root cause.


truong97:
Hey Dave, just for my own note-keeping:

How would you go about testing these changes locally? Would you run offer-uw-testing under your local with the changes?

bencap (approved):
Looks good to me, no parallel workers when running locally but similar performance.

davereinhart:
Deployed.

PR #293, opened 2022-03-04
Use Left Join instead of Inner Join in Reportable Conditions View

bencap:
Currently, the reportable conditions shipping view uses an inner join to join sample
identifiers with presence absence results and collection identifiers. When a sample
uses a tiny swab, this has the side effect of excluding reportable tiny swabs from this
view until they can have their sample identifiers backfilled. By using a left join in
place of this inner join, not having this identifier no longer excludes these samples
from the view, where they will just have a `null` sample identifier until it is backfilled.

This was tested locally by adding a dummy sample and presence absence result to my local
database, and then inserting only a collection identifier - leaving the sample identifier
null. Whereas this dummy sample got pulled into the shipping view using left join, it was not
pulled into the view currently in use prior to this change. Testing this altered view on the
production data, it had the same amount of records as the current view - which makes sense
given there were no records awaiting backfill that were positive when I was testing this change.


bencap:
[Rework reportable conditions shipping view to not require sample ids](https://trello.com/c/ygv3gFla/1330-rework-reportable-conditions-shipping-view-to-not-require-sample-ids)

truong97 (approved):

kschwabefry (approved):
Looks good Ben! Thanks for the write up of how you tested these changes.

bencap:
Deployed.

PR #295, opened 2022-03-09
Re-Add `added_surveillance_groups` to the Priority Queue

bencap:
Last fall, the `added_surveillance_groups` CTE was removed from the priority
queue. This invite logic is once again desired, so this change adds it back
in. The logic is the same as it was before it was removed, but with the change 
that it filters participants who have been invited within 3 days out (previously 
allowed invites for participants as long as they weren't invited on the current 
day). Tested locally by adding a value for this field to some encounter details 
and making sure the individuals got pulled into the queue at the proper priority.


schema/deploy/shipping/views.sql
```diff
@@ -3624,6 +3625,34 @@ create or replace view shipping.__uw_priority_queue_v1 as (
...
+            prior_test_positive_date_base,
+            alerts_off
+        from uw_enrollments
+        -- Filter to added_surveillance_groups for today.
+        -- To_Char(current_date, 'day') returns a string padded to 9 characters, so trim it.
+        -- Example: where 'tuesday' = 'tuesday'
+        where added_surveillance_groups = trim(To_Char(current_date, 'day'))
+        -- Filter for participants who haven't ever been invited or haven't been invited in 3 days
+        and (latest_invite_date is null or latest_invite_date < current_date - interval '3 days')
+	  ),
```
kschwabefry, line 3654: Let's additionally filter out participants who had a sample collected within previous 3 days (see CTEs for positive attestation or normal surveillance for an example).

kschwabefry (changes requested):

kschwabefry (approved):

bencap:
Deployed.

PR #296, opened 2022-03-10
Update mapping for clincial ETL to include new sex value

davereinhart:
A new value was encountered in a recent batch of clinical data for non-binary
sex. Adding to map so that this value can be standardized.


davereinhart:
Deployed.

PR #297, opened 2022-03-15
Modify shipping.scan_encounters_v1's vaccination_status calculated field

truong97:
It was noted that the shipping view did not have as many people showing up as `boosted` as it should be. For example, there are people with 3 doses >14 days ago that are only listed as `fully vaccinated`. Upon investigation, the CASE expression was a bit out of order.

Changes made include reorder the CASE expression where the logic for `boasted` goes before `fully vaccinated`, `partially vaccinated`, `not vaccinated`, etc.

The changes were tested - deployed and reverted on local.


davereinhart (approved):

davereinhart:
Deployed.

PR #299, opened 2022-04-04
SCAN encounters shipping view vaccine_manufacturer calc fix

davereinhart:
The `shipping.scan_encounters_v1` materialized view was returning `null` for the
vaccine manufacturer value in cases where there were only 2 doses received. The logic
here has been updated to ensure that the calculation is accurate regardless of the number
of doses being compared.


davereinhart:
[Modify vaccine_manufacturer calculated field on shipping.scan_encounters_v1 view](https://trello.com/c/AtsC76ox/1355-modify-vaccinemanufacturer-calculated-field-on-shippingscanencountersv1-view)

davereinhart:
Deployed.

PR #300, opened 2022-04-11
Modify clinical ETL to process PHSKC data into FHIR bundles

davereinhart:
Updates to clinical ETL for processing of PHSKC data. This data has being parsed and pre-processing
from Excel worksheets, and inserted into the clinical receiving table. This update to the clinical
ETL will transform the data into FHIR bundle format and inserted into the fhir receiving table to be
processed by the FHIR ETL.

This was modeled after the UW Retrospectives process, which is currently the only clinical data ETL that
uses the FHIR format. To expediate the development and avoid duplicating code, it utilizes existing
functions that were originally designed for REDCap ETLs wherever possible.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/etl/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/etl/clinical.py
@@ -324,9 +324,9 @@ def determine_questionnaire_items(record: dict) -> List[dict]:
     if record["survey_have_symptoms_now"]:
         items["survey_have_symptoms_now"] = [{ 'valueBoolean': survey_have_symptoms_now(record["survey_have_symptoms_now"])}]
 
-    # TODO
-    # add the remaining questionnaire responses:
-    # - survey_testing_because_exposed
+    if record["survey_testing_because_exposed"]:
+        items["survey_testing_because_exposed"] = [{ 'valueString': survey_testing_because_exposed(record["survey_testing_because_exposed"])}]
+
 
     questionnaire_items: List[dict] = []
     for key,value in items.items():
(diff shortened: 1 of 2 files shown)


lib/seattleflu/id3c/cli/command/etl/clinical.py
```diff
@@ -104,45 +115,266 @@ def etl_clinical(*, db: DatabaseSession):
...
+
+    return create_entry_and_reference(encounter_resource, "Encounter")
+
+
+def create_resident_locations(record: dict) -> Optional[tuple]:
+    """
+    Returns FHIR Location resource entry and reference for resident address
+    and Location resource entry for Census tract.
+    """
+    if not record["address"]:
```
kschwabefry, line 250: Do we expect the address itself to be included at this level, or just the census tract?

lib/seattleflu/id3c/cli/command/etl/clinical.py
```diff
@@ -104,45 +115,266 @@ def etl_clinical(*, db: DatabaseSession):
...
+        encounter_date = encounter_date,
+        encounter_status = encounter_status,
+        patient_reference = patient_reference,
+        location_references = encounter_location_references,
+    )
+
+    return create_entry_and_reference(encounter_resource, "Encounter")
+
+
+def create_resident_locations(record: dict) -> Optional[tuple]:
```
kschwabefry, line 245: Doesn't need to happen now, but this looks very similar to create_resident_locations from uw retro etl (minus the geocoding). Seems like a good opportunity to refactor.

lib/seattleflu/id3c/cli/command/etl/clinical.py
davereinhart: The key name is a little misleading here, but the `address` in the ndjson file is already hashed so that's what's being ingested here. That could be renamed to `address_hash` during parsing.

lib/seattleflu/id3c/cli/command/etl/redcap_map.py
```diff
@@ -10,13 +10,16 @@ def map_sex(sex_response: str) -> Optional[str]:
...
         'female': 'female',
+        'f': 'female',
         'indeterminate/other': 'other',
         'other (please specify)': 'other',
         'other': 'other',
         'prefer not to say': 'unknown',
         'dont_say': 'unknown',
+        'u': 'unknown',
         'unknown': 'unknown',
         '': 'unknown'
```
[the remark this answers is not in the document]
davereinhart: Yes, we already discussed this. Both of our branches needed this change, so we'll just resolve any conflicts when we merge.

lib/seattleflu/id3c/cli/command/etl/__init__.py
```diff
@@ -146,7 +146,7 @@ def ethnicity(ethnicity: str) -> Optional[bool]:
 
     >>> ethnicity("NOT HISPANIC OR LATINO")
     False
-
```
kschwabefry, line 149: It looks like the only changes to this file are whitespace changes. Not a big deal, but since the file doesn't need to be modified here, its generally cleaner to exclude it and reset this file.

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
```diff
@@ -471,7 +472,9 @@ def create_encounter_status(redcap_record: dict) -> str:
 
     standardized_status = standardize_whitespace(status.lower())
 
-    if standardized_status not in mapper:
+    if standardized_status in mapper.values():
+        return standardized_status
+    elif standardized_status not in mapper:
         raise Exception(f"Unknown encounter status «{standardized_status}».")
 
     return mapper[standardized_status]
```
kschwabefry, line 480: I'm not sure I see the motivation for this change - is there a reason to not let the existing `return mapper[standardized_status]` handle it? With the change, I'm not sure it will ever reach line 480.

davereinhart: It checks for a match in the map values before checking the map keys (it will get to line 480 if its in keys). I had Ben add a couple of additional fields to the ndjson file to avoid hardcoding them in the ETL, one of them being `"encounter_status":"finished"`, which really doesn't need to be mapped but I included it just for the validation piece. There are other mapping functions that use a similar approach, checking map values before keys:
https://github.com/seattleflu/id3c-customizations/blob/2248106401316b688d846421f2de42d6ea36c5dc/lib/seattleflu/id3c/cli/command/etl/clinical.py#L543

bencap (approved):
Looks good to me and is functioning with my branch.

davereinhart:

lib/seattleflu/id3c/cli/command/etl/redcap_det_uw_retrospectives.py
kschwabefry: Thanks Dave! I see what you're doing here with the values now. Thanks!

kschwabefry (approved):

davereinhart:
Deployed.

PR #301, opened 2022-04-12
Parse PHSKC Clinical Data

bencap:
Adds functionality to parse PHSKC clinical data from excel files and join that clinical
data with lab manifest data. Outputs parsed and joined data to stdout in ndjson format.

Parsing of the data involves generating an individual and encounter identifier, calculating
a patients age, and geocoding their address. Once parsed and columns of interest are
selected, this data is joined with sample data from aliquoting sheets provided by the lab.
Samples are joined using `main_cid`, `all_cids`, and `barcode` (with priority given in that
order). Any records that do not have an associated sample are dropped before being output.
Because CIDs are meant to link samples to phskc records, any duplicate CIDs in the lab
manifest are dropped before joining the data.

Note this parsing depends on a parsed aliquoting sheet (can be generated with the command
`pipenv run id3c manifest parse-using-config`) and a geocoding cache file to store
smarty streets geocoding information.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/clinical.py
@@ -489,6 +489,9 @@ def parse_phskc(phskc_filename: str, phskc_specimen_manifest_filename: str, geoc
         ), axis=1
     )
 
+    # localize encounter timestamps to pacific time
+    clinical_records['encountered'] = clinical_records['collect_ts'].dt.tz_localize('America/Los_Angeles')
+
     # calculate age based on sample collection date and birth day
     clinical_records['birth_date'] = pd.to_datetime(clinical_records['birth_date'])
     clinical_records['age'] = clinical_records.apply(
--- a/lib/seattleflu/id3c/cli/command/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/clinical.py
@@ -160,22 +160,6 @@ def remove_pii(df: pd.DataFrame) -> pd.DataFrame:
     df["identifier"] = df["identifier"].apply(generate_hash)
 
 
-def generate_hash(identifier: str):
-    """
-    Generate hash for *identifier* that is linked to identifiable records.
-    Must provide a "PARTICIPANT_DEIDENTIFIER_SECRET" as an OS environment
-    variable.
-    """
-    secret = os.environ["PARTICIPANT_DEIDENTIFIER_SECRET"]
-
-    assert len(secret) > 0, "Empty *secret* provided!"
-    assert len(identifier) > 0, "Empty *identifier* provided!"
-
-    new_hash = hashlib.sha256()
-    new_hash.update(identifier.encode("utf-8"))
-    new_hash.update(secret.encode("utf-8"))
-    return new_hash.hexdigest()
-
 def drop_missing_rows(df: pd.DataFrame, column: str) -> pd.DataFrame:
     """
     Filters the given *df* by removing rows with ``null`` values
--- a/lib/seattleflu/id3c/cli/command/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/clinical.py
@@ -443,8 +443,10 @@ def add_kp_manifest_data(df: pd.DataFrame, manifest_filenames: tuple, manifest_f
 @clinical.command("parse-phskc")
 @click.argument("phskc_filename", metavar = "<PHSKC Clinical Data filename>")
 @click.argument("phskc_specimen_manifest_filename", metavar = "<PHSKC Specimen Manifest filename(s)>")
-@click.argument("geocoding_cache_file", metavar = "<Geocoding cache filename>")
-def parse_phskc(phskc_filename: str, phskc_specimen_manifest_filename: str, geocoding_cache_file: str) -> None:
+@click.argument("geocoding_cache_file", envvar = "GEOCODING_CACHE", metavar = "<Geocoding cache filename>",
+            type = click.Path(dir_okay=False, writable=True))
+
+def parse_phskc(phskc_filename: str, phskc_specimen_manifest_filename: str, geocoding_cache_file: str = None) -> None:
     """
     Process clinical data from PHSKC.
 
--- a/lib/seattleflu/id3c/cli/command/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/clinical.py
@@ -304,7 +304,7 @@ def add_icd10(df: pd.DataFrame) -> None:
 
 def add_insurance(df: pd.DataFrame) -> None:
     """
-    Adds a new column for insurance type to a given *df*. 
+    Adds a new column for insurance type to a given *df*.
     """
     def insurance(series: pd.Series) -> pd.Series:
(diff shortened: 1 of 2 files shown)


davereinhart (changes requested):

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -433,6 +424,212 @@ def add_kp_manifest_data(df: pd.DataFrame, manifest_filenames: tuple, manifest_f
...
+    clinical_records['encounter_status'] = 'finished'
+
+    # generate encounter and individual identifiers for each record
+    clinical_records["individual"] = clinical_records.apply(
+        lambda row: generate_patient_hash(
+            row['pat_name'].split(',')[::-1],
+            map_sex(row['sex']),
+            str(row['birth_date']),
+            str(row["pat_address_zip"])
+        ), axis=1
```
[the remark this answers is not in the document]
bencap: Is the best way to do this just generate some hashes and try to see if they already exist in the database?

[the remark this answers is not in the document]
davereinhart: We could use the other un-merged branch (PR https://github.com/seattleflu/id3c-customizations/pull/300) to process some records and see how many individuals are updated vs. inserted.

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -433,6 +424,201 @@ def add_kp_manifest_data(df: pd.DataFrame, manifest_filenames: tuple, manifest_f
...
+    )
+
+    clinical_records['identifier'] = clinical_records.apply(
+        lambda row: generate_hash(
+            f"{row['mrn']}{row['acc_num']}{row['collect_ts']}".lower()
+        ), axis=1
+    )
+
+    # calculate age based on sample collection date and birth day
+    clinical_records['birth_date'] = pd.to_datetime(clinical_records['birth_date'])
```
kschwabefry, line 470: We should be sure to apply id3c's `age_ceiling` here

lib/seattleflu/id3c/cli/command/clinical.py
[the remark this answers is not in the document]
bencap: Wrapping this up, there were 15 individual identifier matches. Looks like there were some phskc retros ingested previously. Those individuals are associated with 14 mail samples, 2 kiosk samples, and 16 phskc retro samples.

lib/seattleflu/id3c/cli/command/clinical.py
bencap: What is the necessity of having an `age_ceiling` from id3c's perspective? (just curious)

kschwabefry: Thats a great question, here's the reasoning: for participants in the upper age groups, there is an increased risk of re-identifiability because there are simply far fewer of them. The ceiling lowers this risk by grouping participants with an age over what we've defined as the ceiling into the same age value.

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -433,6 +424,203 @@ def add_kp_manifest_data(df: pd.DataFrame, manifest_filenames: tuple, manifest_f
...
+    clinical_records = clinical_records[columns_to_keep]
+    clinical_records = clinical_records.rename(columns=column_map)
+
+    # phskc data is sent with some rows duplicated, so before we add manifest data
+    # we should drop rows with duplicated values in their join criteria
+    clinical_records.drop_duplicates(subset=['phskc_barcode', 'main_cid', 'all_cids'], inplace=True)
+    clinical_records = add_phskc_manifest_data(clinical_records, phskc_specimen_manifest_filename)
+    clinical_records.drop('merge_col', axis=1, inplace=True)
+
+    dump_ndjson(clinical_records)
```
davereinhart, line 540: We should also drop `main_cid`, `all_cids`, and `phskc_barcode` from the ndjson file, since they aren't needed after we have the sample barcode and we don't want to ingest them.

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -433,6 +424,206 @@ def add_kp_manifest_data(df: pd.DataFrame, manifest_filenames: tuple, manifest_f
...
+            )
+        )
+
+    db = DatabaseSession()
+    clinical_records = clinical_records.apply(
+       lambda row: encode_addresses(db, row), axis=1
+    )
+
+    column_map = {
+        "collect_ts": "encountered",
```
davereinhart, line 504: The timestamps in the metadata files _claim_ to be in UTC (they end in "Z" for Zulu), but looking at the values they appear to be in local time.  This is causing them to be shifted by 8 hours when they are ingested.

davereinhart: Correction, @bencap pointed out that the values in the metadata files have no timezone info, so are defaulting to UTC. A timezone-aware column is being added to handle it.

kschwabefry: Great catch on this!

davereinhart (approved):

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -433,6 +424,208 @@ def add_kp_manifest_data(df: pd.DataFrame, manifest_filenames: tuple, manifest_f
     return df.merge(manifest_data[['barcode', 'enrollid']], how='left')
 
 
+@clinical.command("parse-phskc")
+@click.argument("phskc_filename", metavar = "<PHSKC Clinical Data filename>")
+@click.argument("phskc_specimen_manifest_filename", metavar = "<PHSKC Specimen Manifest filename(s)>")
+@click.argument("geocoding_cache_file", metavar = "<Geocoding cache filename>")
```
kschwabefry, line 430: It would be convenient to pass this in as an environment variable if thats ok on your end. Would you mind adding 
` envvar = "GEOCODING_CACHE"` here?

kschwabefry (approved):

davereinhart:

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -433,6 +424,210 @@ def add_kp_manifest_data(df: pd.DataFrame, manifest_filenames: tuple, manifest_f
...
+    """
+    Process clinical data from PHSKC.
+
+    Given a <PHSKC Clinical Data filename> of an Excel document, selects specific
+    columns of interest and reformats the queried data into a stream of JSON
+    documents suitable for the "upload" sibling command.
+
+    All clinical records parsed are output to stdout as newline-delimited JSON
+    records.  You will likely want to redirect stdout to a file.
+    """
```
davereinhart, line 443: Non-blocker, but would be worthwhile to update the comments here to reflect that parse-phskc command is actually parsing and transforming, removing PII, requires <phskc_specimen_manifest_filename>, and only matching records are returned.  It's quite a bit different than the other parse commands.

davereinhart:
Deployed.

PR #302, opened 2022-04-14
Adding new value to test result mapping function

davereinhart:
Multiple new test result values were encountered during ingestion of UW Retros metadata
that begin with "invalid", which had not been previously seen or mapped. Adding to the
`test_result_prefix_map` to re-process those records that failed and handle these values going forward.


bencap (approved):

davereinhart:
Deployed.

PR #303, opened 2022-04-21
Convert NaN to Empty String Prior to Geocoding

bencap:
Geocoding expects either NoneType or String objects for address fields,
but NaN values were being sent to the geocoding function. This change
converts any NaNs in the parsed clinical records to empty strings before
geocoding so that the expected data type matches what is passed in.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/clinical.py
@@ -489,9 +489,20 @@ def parse_phskc(phskc_filename: str, phskc_specimen_manifest_filename: str, geoc
             ), axis=1
     )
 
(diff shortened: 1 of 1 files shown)


kschwabefry (approved):
Thanks for the fix Ben. And good to fixup the lang vs lng too!!

davereinhart (commented):
Can we check for values in the required fields before geocoding? It could get noisy if there are a lot of records missing addresses and it gets to this warning-level log:
https://github.com/seattleflu/id3c/blob/00bd7d051ff5c84befa83ec47ccc78ff40dd146a/lib/id3c/cli/command/geocode.py#L322

bencap:

bencap:
Deployed.

PR #304, opened 2022-04-26
offer_uw_testing: dial down redcap batch size

kschwabefry:
We are seeing 1-2x/daily failures of this job due to timeouts
on the call to redcap to update records. As the project has grown
larger and there have been performance issues in the redcap service,
let's try dialing this threshold down to update only 150 records in
a batch to see if it improves the reliability.


bencap (approved):

kschwabefry:
Deployed.

PR #305, opened 2022-04-29
Deliverables log

davereinhart:
Add a new table to shipping schema to track deliverables sent, and a corresponding ID3C command to insert log entries into this table.


davereinhart:
After attempting to implement in return of results, I think I'm going to refactor this code and move it to the id3c repo as part of the db module.

davereinhart:
Can be tested using this backoffice branch: https://github.com/seattleflu/backoffice/pull/300

davereinhart:
Deployed.

PR #306, opened 2022-05-03
uw retro etl: map new encounter_status and encounter_class values

kschwabefry:
To reflect data model changes upstream, include new values:

encounter_status:
  no show

encounter_class:
  series pt-ot-st
  deceased - organ donor
  hospital outpatient surgery
  observation

There was some ambiguity in how to map "deceased-organ donor"
as an encounter class (we do already handle it as a discharge
disposition), but we've decided to default to mapping it to an
ambulatory/outpatient encounter because that's generally our
default and there wasn't an unknown option.


davereinhart (approved):

kschwabefry:
Deployed.

PR #307, opened 2022-05-03
AIRS Redcap->FHIR ETL

seidelma:


lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+            }
+
+    if incomplete_enrollment_instruments:
+        LOG.debug(f"The following required enrollment instruments «{incomplete_enrollment_instruments}» are not yet marked complete.")
+        return None
+
+    # If the participant's age < 18 do not continue
+    if (enrollment['scr_age'] == "" or int(enrollment['scr_age']) < 18):
+        LOG.debug("The participant age is blank or marked as under 18. Skipping record.")
+        return None
```
kschwabefry, line 152: I think we can take this part out because we'd want to leave this responsibility to ensure consent to the study itself. If there's redcap data, we should be able to assume there is proper consent.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+    """
+    # For all surveys, try the survey _timestamp field (which is in Pacific time)
+    # before custom fields because the custom fields aren't always populated and when
+    # they are populated they use the browser's time zone.
+    collection_date = None
+
+    if collection_method == CollectionMethod.KIOSK:
+        collection_date = extract_date_from_survey_timestamp(record, "kiosk_registration_4c7f") or record.get("nasal_swab_q")
+
+    elif collection_method in [CollectionMethod.SWAB_AND_SEND, CollectionMethod.UW_DROPBOX]:
```
kschwabefry, line 431: For swab and send in this study, I think the redcap instruments you'll want to check for the collection date are airs_kit_activation and airs_kit_activation_2 . Kiosk and dropbox won't be relevant for airs.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+    ]
+
+LANGUAGE_CODE = {
+project.id: project.lang
+    for project in PROJECTS }
+
+class CollectionMethod(Enum):
+    SWAB_AND_SEND = 'swab_and_send'
+    KIOSK = 'kiosk'
+    UW_DROPBOX = 'uw_dropbox'
```
kschwabefry, line 44: I just confirmed with Cooper that this study is swab and send only, so you should be safe to drop out all the logic for drop box and kiosk and the TODO a few lines down.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+    "week_22_arm_1",
+    "week_23_arm_1",
+    "week_24_arm_1",
+    "week_25_arm_1",
+    "week_26_arm_1",
+]
+
+SWAB_AND_SEND_SITE = 'AIRSSwabNSend'
+
+STUDY_START_DATE = datetime(2021, 9, 9) # TODO: I dunno lol
```
kschwabefry, line 88: This shouldn't be relevant for AIRS and can be dropped. It was used in HCT to calculate repeat instance numbers.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+        #    they will be prompted to self-collect a nasal swab,
+        #    which will be delivered and returned via courier, 
+        #    and complete a brief symptom survey
+        # which doesn't even rhyme.
+        elif redcap_record_instance.event_name in ENCOUNTER_EVENT_NAMES:
+            if is_complete('weekly', redcap_record_instance):
+            #TODO: Do we skip events when no kit was sent, or are they still relevant?
+            #and \
+            #    (is_complete('airs_kit_activation_complete', redcap_record_instance) or \
+            #     is_complete('airs_kit_activation_2_complete', redcap_record_instance)):
```
kschwabefry, line 188: Are we still waiting for an answer on this one?

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+
+        # TODO: I believe there's only one collection_method: swab_and_send
+        # The study documents say: 
+        #    If they report symptoms on a survey, 
+        #    they will be prompted to self-collect a nasal swab,
+        #    which will be delivered and returned via courier, 
+        #    and complete a brief symptom survey
+        # which doesn't even rhyme.
+        elif redcap_record_instance.event_name in ENCOUNTER_EVENT_NAMES:
+            if is_complete('weekly', redcap_record_instance):
```
kschwabefry, line 184: It wasn't obvious to me from the redcap setup, but I know you've looked at this more in depth than me - do we know that the "weekly" instrument is required to be filled out for the event? (If not, it might make sense to relax this requirement so we aren't for example, excluding events where a sample was collected but the weekly survey wasn't filled out.)

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+            'uw_club':  'UWClub'
+            }
+
+        # TODO: since there's only one collection method
+        #  (courier-delivered swabs) I believe this will always be the same.
+        site_reference = create_site_reference(
+            location = None,
+            site_map = location_site_map,
+            default_site = SWAB_AND_SEND_SITE,
+            system_identifier = INTERNAL_SYSTEM)
```
kschwabefry, line 234: These look HCT specific. I think you're correct, that since these are always swab and send, this can be simplified.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+    "week_20_arm_1",
+    "week_21_arm_1",
+    "week_22_arm_1",
+    "week_23_arm_1",
+    "week_24_arm_1",
+    "week_25_arm_1",
+    "week_26_arm_1",
+]
+
+SWAB_AND_SEND_SITE = 'AIRSSwabNSend'
```
kschwabefry, line 86: We'll want to map this new site in the warehouse.sites table. That doesn't necessarily need to be handled as part of this PR, but maybe once this is finalized, can you please add a trello task to add the site to schema/deploy/warehouse/site/data.sql  ? Thanks!

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+            symptom_onset_map = {
+                # sic--this misspelling is in the redcap form.
+                'symptoms': redcap_record_instance['ss_ealiest_date'],
+            }
+        # Irrelevant, no symptom questions are present in enrollment
+        elif event_type == EventType.ENROLLMENT:
+            symptom_onset_map = {'symptoms_base': redcap_record_instance['wk_symp_start_date']}
+        # Should never get here due to the event_type check at the top of the loop
+        else:
+            LOG.error(f"Invalid event_type {event_type} fell through")
```
kschwabefry, line 254: Are ss_ealiest_date and wk_symp_start_date both from the repeating weekly encounters? I 
I the referenced think daily_symptoms_covid_like in the comment is a holdover from HCT.

Is the enrollment instrument current_symptoms applicable here?  I'm not sure if it collects a start date though. If there isn't one, we could potentially use the date of that instrument as a best effort, but that might cause other data consistency issues 🤔

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+    }
+
+    for field in checkbox_fields:
+        record[field] = combine_checkbox_answers(record, field)
+
+    # Combine all fields answering the same question
+    record['countries_visited_base'] = combine_multiple_fields(record, 'country', '_base')
+    record['states_visited_base'] = combine_multiple_fields(record, 'state', '_base')
+
+    # TODO: Do we need this? There are multiple date questions encompassing COVID & flu vaccinations
```
kschwabefry, line 521: @davereinhart do you have some suggestions on how to store the vaccine data?

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+        'valueDate': date_questions,
+        #'valueDecimal': decimal_questions (none so far)
+    }
+
+    for field in checkbox_fields:
+        record[field] = combine_checkbox_answers(record, field)
+
+    # Combine all fields answering the same question
+    record['countries_visited_base'] = combine_multiple_fields(record, 'country', '_base')
+    record['states_visited_base'] = combine_multiple_fields(record, 'state', '_base')
```
kschwabefry, line 519: Are these questions asked in AIRS? They caught my eye as perhaps questions specific to HCT.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+
+
+def get_date_from_repeat_instance(instance_id: int) -> str:
+    """
+    Returns the date associated with a REDCap repeat instance.
+    This is safe to do only when the Daily Attestation exists
+    because the Musher sets the repeat instance based on the date.
+    Returns the date as a string because that is how it is used.
+    """
+    return (STUDY_START_DATE + relativedelta(days=(instance_id -1))).strftime('%Y-%m-%d')
```
kschwabefry, line 726: This was specific for HCT, and shouldn't be needed here.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+    `build_residential_location_resources()` expects, so we've got to
+    map them to one of the allowed values.
+    """
+    # Is this group housing? If so, map it to a value that results in
+    #  build_residential_location_resources() assigning it a 'lodging'
+    #  housing_type. 
+    # TODO: verify with study this is appropriate
+    if enrollment.get('enr_living_area') == '1':
+        # Maps to 'lodging'
+        housing = 'shelter'
```
kschwabefry, line 775: `enr_living_group` might be useful in making this data more specific.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+                specimen_reference = specimen_reference,
+                patient_reference = patient_reference,
+                encounter_reference = initial_encounter_reference)
+        else:
+            LOG.debug(f"No identifier found for barcode «{specimen_barcode}»")
+            LOG.info("No identifier found for barcode. Creating encounter for record instance without sample")
+    else:
+        LOG.debug("Creating encounter for record instance without sample")
+
+    # Do not log warning for unrecognized barcodes, which are being tracked outside of ID3C.
```
kschwabefry, line 850: This might be worth some discussion of how QC for unrecognized barcodes should be handled for this study.

kschwabefry (changes requested):
I know there are still some TODOs sprinkled in the code, but here's some notes from an initial review. Thanks.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
davereinhart: I think we would just want to follow the design of SCAN ETL, where the vaccination-related questions are included along with other questionnaire responses here:
https://github.com/seattleflu/id3c-customizations/blob/743d8f0a3ebce48ae006b869dad198785bc182b3/lib/seattleflu/id3c/cli/command/etl/redcap_det_scan.py#L844

The `create_vaccine_item` function in redcap.py is currently being used only for the flu vaccine, and only in childcare and UW retro ETLs.

lib/seattleflu/id3c/cli/command/etl/redcap_det_airs.py
```diff
@@ -0,0 +1,908 @@
...
+    lang: str
+    command_name: str
+
+    def __init__(self, project_id: int, lang: str, command_name: str) -> None:
+        self.id = project_id
+        self.lang = lang
+        self.command_name = command_name
+
+
+REDCAP_URL  = 'https://redcap.fredhutch.org/'     # REDCap server hostname
```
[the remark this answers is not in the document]
seidelma: Yep, sounds like a good idea.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,910 @@
...
+        if redcap_record_instance.event_name == ENROLLMENT_EVENT_NAME:
+            event_type = EventType.ENROLLMENT
+
+        # TODO: I believe there's only one collection_method: swab_and_send
+        # The study documents say: 
+        #    If they report symptoms on a survey, 
+        #    they will be prompted to self-collect a nasal swab,
+        #    which will be delivered and returned via courier, 
+        #    and complete a brief symptom survey
+        # which doesn't even rhyme.
```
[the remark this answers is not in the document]
seidelma: That's part of a todo which I've removed, sorry!

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: Noted and removed.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: Yes, this was left over from the UW reopening code. Removed.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: Noted and removed.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: I've kept the `SWAB_AND_SEND_SITE` because that is required for `create_site_reference()` below.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: No, via Kayla's conversation with them we know that the study wants data for participants who haven't received kits.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: I believe that the weekly instrument is the initial point-of-contact between participants and the study--that is, the symptom formula on the weekly page is what dictates whether a (first) kit is sent to a participant. I don't see any instances of people getting kits without a complete weekly form. That's the basis for the encounter here.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: I've removed the `location_map` and set `site_map` to None because it's always swab-and-send. That will trigger `create_site_reference()` to use the specified `default_site`.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: That's done: https://trello.com/c/aEd5FD0U/1381-add-airsswabnsend-site-to-warehousesite-table

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: Yes, I removed the comment referring to `daily_symptoms_covid_like`.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: I'll look into this a bit further.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: Noted and removed.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: Noted and removed.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: Unfortunately, the `enr_living_group` values don't map very well between AIRS and redcap.py. What I'm doing here is to short-circuit the check in redcap.py's [build_residential_location_resources()](https://github.com/seattleflu/id3c-customizations/blob/8b1763133a9ffa966ce3103cd58d2a4bd748939a/lib/seattleflu/id3c/cli/command/etl/redcap.py#L136) because that's only checking to see if the residence type is in a list of values, and if so, flagging it as a group home in the location. It's effectively boolean, so I picked a value that would cause the flag to be set if a participant marks that they live in a group home setting.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
seidelma: Yeah, this is just another holdover from the UW reopening ETL.

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,861 @@
...
+    # before custom fields because the custom fields aren't always populated and when
+    # they are populated they use the browser's time zone.
+    collection_date = None
+
+    if collection_method == CollectionMethod.SWAB_AND_SEND:
+        collection_date = record.get("date_on_tube") \
+            or extract_date_from_survey_timestamp(record, "husky_test_kit_registration") \
+            or record.get("kit_reg_date") \
+            or extract_date_from_survey_timestamp(record, "test_fulfillment_form") \
+            or record.get("back_end_scan_date")
```
kschwabefry, line 398: These look like the instruments for HCT still. Can you please double check what they should be for AIRS?

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,861 @@
...
+    lang: str
+    command_name: str
+
+    def __init__(self, project_id: int, lang: str, command_name: str) -> None:
+        self.id = project_id
+        self.lang = lang
+        self.command_name = command_name
+
+
+REDCAP_URL  = os.environ.get('FH_AIRS_REDCAP_URL') # REDCap server hostname
```
kschwabefry, line 30: Thanks for adding this. Small change but by convention, can the suffix be `_REDCAP_API_URL` to match the others we use? Thanks!

lib/seattleflu/id3c/cli/command/etl/redcap_det_fh_airs.py
```diff
@@ -0,0 +1,857 @@
...
+    # they are populated they use the browser's time zone.
+    collection_date = None
+
+    if collection_method == CollectionMethod.SWAB_AND_SEND:
+        collection_date = record.get("kit_reg_date_v2") \
+            or extract_date_from_survey_timestamp(record, "post_collection_data_entry_qc_2") \
+            or record.get("back_end_scan_v2") \
+            or record.get("kit_reg_date") \
+            or extract_date_from_survey_timestamp(record, "post_collection_data_entry_qc") \
+            or record.get("back_end_scan")
```
kschwabefry, line 394: A couple thoughts on this part.

* Should this be split out into two different functions or pathways depending if the swab is the 1 or 2 for the week? If two samples were collected in the same week, I think we'd need to be able to keep the collection dates straight and check the appropriate instrument.

* Can you comment on the instruments chosen for the sample collection timestamp and the ordering? I was thinking that for `swab_and_send` samples, we would be using the `airs_kit_activation` 1 and 2 respectively . It looks like this is also checking back end mail scans and post collection data entry qc forms -- are those intended as backups in case the participant didnt activate the kit?
I'm not sure if I have the whole picture here, but would it make the most sense to prioritize the kit activation instruments over the other dates, and additionally prioritize the survey instrument date before the custom fields within each instrument?

davereinhart:
Deployed.

PR #308, opened 2022-05-09
Grant permissions on deliverables log table to return-results-exporter.

davereinhart:
Insert permissions are required to write back to ID3C's deliverables_log table during return of results.


davereinhart:
Deployed via sqitch.

PR #313, opened 2022-05-17
Add shipping view for SCAN encounters with best available vaccination…

davereinhart:
… data.

Data quality issues were previously identified in the SCAN REDCap project vaccination
date and manufacturer fields, with participants sometimes entering invalid or conflicting
data during one or more encounters. To determine the most accurate vaccination status with
the available data, this shipping view is being added.

The logic being used here is first to identify the most-frequently reported valid vaccination date
for each dose. Valid dates are considered to be those between 12/01/2020 (based on the wording
of the questionnaire in REDCap) and the date of the encounter they were entered on (to exclude
future dates entered). The most frequently-reported (or only) valid date for each dose is then
applied to all encounters for that participant. The vaccination status for each encounter is then
calculated based on those applied dates and associated manufacturer values.

Most recent related Slack thread:
https://seattle-flu-study.slack.com/archives/CCAA9RBFS/p1652205460785849

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -2328,6 +2328,196 @@ grant select
   to "incidence-modeler";
 
 
+create or replace view shipping.scan_encounters_with_best_available_vaccination_data_v1 as
+
+    with vac_1_most_frequent_valid_date as (
+      select distinct on (individual)
+      individual,
+      vac_date,
+      vac_name_1
+      from shipping.scan_encounters_v1
+      where vac_date >= '2020-12-01'
+        and vac_date::date <= encountered::date
+      group by individual, vac_date, vac_name_1
+      having count(*) > 1
+      order by individual, count(*) desc
+    ),
+    vac_1_next_best_date as (
+      select distinct on (individual)
+      individual,
+      vac_date,
+      vac_name_1,
+      case
+        when vac_date < '2020-12-01' or vac_date::date > encountered::date then 1
+      end as vac_date_1_out_of_range
+      from shipping.scan_encounters_v1
+      where individual not in (select individual from vac_1_most_frequent_valid_date)
+        and vac_date is not null
+      order by individual, vac_date_1_out_of_range nulls first, encountered
+    ),
+    vac_2_most_frequent_valid_date as (
+      select distinct on (individual)
+      individual,
+      vac_date_2,
+      vac_name_2
+      from shipping.scan_encounters_v1
+      where vac_date_2 >= '2020-12-01'
+        and vac_date_2::date <= encountered::date
+      group by individual, vac_date_2, vac_name_2
+      having count(*) > 1
+      order by individual, count(*) desc
+    ),
+    vac_2_next_best_date as (
+      select distinct on (individual)
+      individual,
+      vac_date_2,
+      vac_name_2,
+      case
+        when vac_date_2 < '2020-12-01' or vac_date_2::date > encountered::date then 1
+      end as vac_date_2_out_of_range
+      from shipping.scan_encounters_v1
+      where individual not in (select individual from vac_2_most_frequent_valid_date)
+        and vac_date_2 is not null
+      order by individual, vac_date_2_out_of_range nulls first, encountered
+    ),
+    vac_3_most_frequent_valid_date as (
+      select distinct on (individual)
+      individual,
+      vac_date_3,
+      vac_name_3
+      from shipping.scan_encounters_v1
+      where vac_date_3 >= '2020-12-01'
+        and vac_date_3::date <= encountered::date
+      group by individual, vac_date_3, vac_name_3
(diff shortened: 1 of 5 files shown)


davereinhart:
[Create a shipping view for vaccination status per encounter](https://trello.com/c/tQO55o3U/1344-create-a-shipping-view-for-vaccination-status-per-encounter)

schema/deploy/shipping/views.sql
```diff
@@ -2328,6 +2328,196 @@ grant select
...
+      select distinct on (individual)
+      individual,
+      vac_date,
+      vac_name_1
+      from shipping.scan_encounters_v1
+      where vac_date >= '2020-12-01'
+        and vac_date::date <= encountered::date
+      group by individual, vac_date, vac_name_1
+      having count(*) > 1
+      order by individual, count(*) desc
```
kschwabefry, line 2343: Would adding another order by encountered make the ordering more deterministic? I can imagine someone entering, say, two different valid date twice.

kschwabefry, line 2343: Would adding a final `order by ... encountered` make the ordering more deterministic? I can imagine someone entering, say, two different valid date twice.

schema/deploy/shipping/views.sql
```diff
@@ -2328,6 +2328,196 @@ grant select
...
+      vac_date,
+      vac_name_1,
+      case
+        when vac_date < '2020-12-01' or vac_date::date > encountered::date then 1
+      end as vac_date_1_out_of_range
+      from shipping.scan_encounters_v1
+      where individual not in (select individual from vac_1_most_frequent_valid_date)
+        and vac_date is not null
+      order by individual, vac_date_1_out_of_range nulls first, encountered
+    ),
```
kschwabefry, line 2357: To document conversation, we discussed whether these CTEs could be combined to simplify the query, but because of requirements to prefer the most frequently entered valid date, but to prefer less frequently invalid dates if thats all thats available, it makes sense to keep these separate.

schema/deploy/shipping/views.sql
```diff
@@ -2328,6 +2328,196 @@ grant select
...
+          when vac_date_1_out_of_range is null and calculated_vac_name_1 = 'johnson' and date_part('day', encountered::timestamp - calculated_vac_date_1::timestamp) >= 14 then 'fully_vaccinated'
+          when vac_date_2_out_of_range is null and calculated_vac_name_1 in (values('pfizer'), ('moderna')) and calculated_vac_name_2 in (values('pfizer'), ('moderna')) and date_part('day', encountered::timestamp - calculated_vac_date_2::timestamp) >= 14 then 'fully_vaccinated'
+          when vac_date_1_out_of_range is null and calculated_vac_name_1 in (values('pfizer'), ('moderna'), ('johnson')) and date_part('day', encountered::timestamp - calculated_vac_date_1::timestamp) >= 1 then 'partially_vaccinated'
+          when vac_date_1_out_of_range is null and date_part('day', encountered::timestamp - calculated_vac_date_1::timestamp) = 0 then 'not_vaccinated'
+        when coalesce(vac_date_1_out_of_range, vac_date_2_out_of_range, vac_date_3_out_of_range) = 1 then 'invalid'
+          else 'unknown'
+      end as best_available_vac_status
+    from encounters_with_best_available_vac_applied_dose_count
```
kschwabefry, line 2517: Documenting conversation because this confused me at first- there was a request from the end users that `best_available_vac_status` be `boosted` if theres a valid `date3` to even if prior doses were invalid dates.

kschwabefry (approved):
Looks good Dave. Thanks for working with the end users to pull this together!

schema/deploy/shipping/views.sql
davereinhart: Good call! I don't think we had any cases of this happening, but it wouldn't hurt to add it.

davereinhart: Added 00e8e18e01b9d814ff5313500fc0fcc324b06f6e. This change does actually impact a handful of records, and in theory should be more accurate since we're selecting the encounter that would have been closest to the vaccination date when there is a tie in the count.

kschwabefry (approved):

schema/revert/shipping/views.sql
```diff
@@ -42,6 +42,7 @@ drop view if exists shipping.genome_submission_metadata_v1;
 drop view if exists shipping.flu_assembly_jobs_v1;
 
 drop view if exists shipping.scan_follow_up_encounters_v1;
+drop view if exists shipping.scan_encounters_with_best_available_vaccination_data_v1;
```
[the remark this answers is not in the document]
davereinhart: No, it shouldn't be necessary there because it should not exist yet if you haven't deployed that version.

davereinhart:
Deployed.

PR #314, opened 2022-05-18
Update ETL for SCH Year 4

seidelma:


lib/seattleflu/id3c/cli/command/etl/clinical.py
```diff
@@ -463,15 +463,19 @@ def encounter_details(document: dict) -> dict:
...
+    for k in ["CovidShot1", "CovidShot2", "CovidShot3"]:
         if k in document:
             details["responses"][k] = covid_shot(document[k])
 
     if "CovidShotManufacturer" in document:
         details["responses"]["CovidShotManufacturer"] = covid_shot_maunufacturer(document.get("CovidShotManufacturer"))
 
+    for k in ["CovidShot1Manu", "CovidShot2Manu", "CovidShot3Manu"]:
+        if document.get(k):
+            details["responses"][k] = [document[k]]
```
davereinhart, line 475: Could use the `covid_shot_manufacturer` function here to validate and standardize incoming values.

lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -223,6 +223,24 @@ def parse_sch(sch_filename, manifest_format, output):
...
+            "covid_vx_d2": "CovidShot2",
+            "cov_d2_date": "CovidShot2Date",
+            "covid_vx_manu2": "CovidShot2Manu",
+            "covid_vx_d3": "CovidShot3",
+            "cov_d3_date": "CovidShot3Date",
+            "covid_vx_manu3": "CovidShot3Manu",
+        })
+    else:
+        # Don't fall through silently
+        LOG.warning(f"Invalid manufest_format {manifest_format}")
```
davereinhart, line 243: minor typo

davereinhart (changes requested):
We'll need to add "year4" to the `--manifest-format` option here, and should probably become the new default:
https://github.com/seattleflu/id3c-customizations/blob/seidelma/feature/sch_year_4/lib/seattleflu/id3c/cli/command/clinical.py#L176-L178

davereinhart:
Picking this work up since @seidelma is out of office.

davereinhart:
To test, pull the latest SCH Excel file from S3 and run the following three commands:

`envdir ../env.d/deidentification pipenv run id3c clinical parse-sch -o [missing barcodes output file] --manifest-format year4 [path to XLSX file] > [parsed ndjson file]`

`PGSERVICE=seattleflu-local pipenv run id3c clinical upload [parsed ndjson file]`

`PGSERVICE=seattleflu-local LOG_LEVEL=debug pipenv run id3c clinical etl --commit`

kschwabefry:
[Update SCH variables for covid vaccines manufactures](https://trello.com/c/b20Nps57/1368-update-sch-variables-for-covid-vaccines-manufactures)

kschwabefry:
Looks good for the new data. My main thought here is that its confusing to call the new format year4. The SCH files labeled 'year4' but pre 4/27/22 would actually need to use manifest format 'year3', which seems counterintuitive given the current naming. I think that could be addressed by choosing a format name thats more descriptive than just year, since this was a "mid-year" format change.

davereinhart:
> Looks good for the new data. My main thought here is that its confusing to call the new format year4. The SCH files labeled 'year4' but pre 4/27/22 would actually need to use manifest format 'year3', which seems counterintuitive given the current naming. I think that could be addressed by choosing a format name thats more descriptive than just year, since this was a "mid-year" format change.

Good point! That would apply to the `year3` value too since it applies to some year4 files, so maybe all of them should be updated to a date range.  (e.g. `20210101-20220426`, `20220427-present`, etc.)

davereinhart:
Actually, it looks like the new file that was uploaded contains ALL records for year 4, so the previous year 4 files can be archived and the format naming convention can stay intact.

davereinhart:
Deployed, and moved older format year 4 files to the `archive` subfolder on S3.

kschwabefry:
Thanks Dave! Good call to archive the previous y4 files if the new 4/27 file includes them.

PR #317, opened 2022-05-25
Refactor EHS shipping view to improve performance

davereinhart:
The uw_reopening_ehs_reporting_v1 shipping view has had ongoing performance issues,
becoming unresponsive at times. This refactored version utilizing CTEs sucessfully ran on one
of those occasions, and has an improved (lower cost) query plan than the existing version.


davereinhart:
[Performance improvements for EHS shipping view](https://trello.com/c/RppiQu6I/1384-performance-improvements-for-ehs-shipping-view)

schema/deploy/shipping/views.sql
```diff
@@ -3697,44 +3699,57 @@ grant select
...
+      when on_campus_freq = 'one_or_less' then 'One day a week or less'
+      when on_campus_freq = 'two_or_more' then 'Two days a week or more'
+      when on_campus_freq = 'not_on_campus' then 'Do not come to campus'
+      else 'Unknown'
+  end as on_campus_frequency_description
+  , wfh_base as able_to_work_or_study_from_home_code
+  , case
+      when wfh_base = 'onsite' then 'No, I always have to be on-site for work or school'
+      when wfh_base = 'only_wfh' then 'Yes, I have only worked or studied from home'
+      when wfh_base = 'wfh_onsite' then 'I have worked or studied both from home and on-site'
```
kschwabefry, line 3729: Is there a reason to change the capitalization there? Otherwise, I would prefer to leave this capitalization consistent

kschwabefry (approved):
Thanks Dave, this looks good to me. I'm getting similar timings with the two currently, though slightly different query plans with the two. I'll be interested to see how this performs in prod.

At first glance I thought the CTE might cause some issues with the query planner, but it looks like that has changed with PG 12 (see, e.g. https://medium.com/alan/ctes-optimization-8c082efc47ef), and since we're on v13 now (hooray), this shouldn't be an issue 

I was able to confirm the results were the same with 

```
select * from (...) as query1
except
select * from (...) as query2 
```

and vice versa

schema/deploy/shipping/views.sql
davereinhart: Good catch! I Inadvertently changed this when I was lowercasing instances of "AND" that I had added.

davereinhart:
> Thanks Dave, this looks good to me. I'm getting similar timings with the two currently, though slightly different query plans with the two. I'll be interested to see how this performs in prod.
> 
> At first glance I thought the CTE might cause some issues with the query planner, but it looks like that has changed with PG 12 (see, e.g. https://medium.com/alan/ctes-optimization-8c082efc47ef), and since we're on v13 now (hooray), this shouldn't be an issue
> 
> I was able to confirm the results were the same with
> 
> ```
> select * from (...) as query1
> except
> select * from (...) as query2 
> ```
> 
> and vice versa

That's a handy query for comparing queries! Definitely going to start using that.

I'm seeing similar runtimes on prod with both versions, but the main difference I see in the query plan is it's using hash joins instead of nested loops at the beginning of the two main branches, and the second branch (results) is working on just 1/3 of the records from RoR shipping view.

seidelma (approved):
LGTM. Here's a visual analysis of how the query plans differ, as run on my local machine.

old: https://explain.dalibo.com/plan/0p0
new: https://explain.dalibo.com/plan/UZQ

davereinhart:
> LGTM. Here's a visual analysis of how the query plans differ, as run on my local machine.
> 
> old: https://explain.dalibo.com/plan/0p0 new: https://explain.dalibo.com/plan/UZQ

Thanks, @seidelma!  That looks like a useful tool. I'm still trying to understand how and why the query planner is handling these differently, so let me know if you have any thoughts there. I'm guessing there are other shipping views that could benefit from similar changes.

PR #319, opened 2022-06-01
Only Dump PHSKC NDJSON if Data Exists

bencap:
PHSKC data is uploaded to Sharepoint prior to the lab adding it to AQ sheets. When this data is parsed the outputted files are non-empty since they have a newline character, making it more difficult to verify if a file is truly non-empty. This change only outputs PHSKC data when the DF is non-empty, ensuring the file is empty if there is no aliquot data to join to and making it simpler to check files downstream for emptiness.


davereinhart (approved):

bencap:
Deployed.

PR #320, opened 2022-06-01
Update hcov19_presence_absence_result_v1 shipping view.

davereinhart:
New result timestamps are being received from Samplify, and stored in `presence_absence.details`
which provide a more accurate and static value than the `presence_absence.modified` timestamp
currently being used to determine the result date. Since historical data does not include this data,
we'll continue using `modified` when the preferred timestamp value is not present.


bencap (approved):
Looks good to me and seems to be coalescing properly locally.

bencap (commented):
Makes sense. If these records from end of May until 6-8 did ever get modified for whatever reason though then this `result_ts` for those records would change from their original `pa.modified` value (the one before 6-8) to the value they have in `resultTimestamp` correct?

davereinhart:
> Makes sense. If these records from end of May until 6-8 did ever get modified for whatever reason though then this `result_ts` for those records would change from their original `pa.modified` value (the one before 6-8) to the value they have in `resultTimestamp` correct?

Good point! I guess it would be better to apply the cutoff to the `result_timestamp` from Samplify since that should be more static.

bencap (approved):
Ah yeah that's smart and should be a more defined cutoff. Looks good!

davereinhart:
> Looks good to me. I had some concern that the `>` operator might miss some data from 6/8/2022, but that doesn't seem to be the case from a search of the warehouse.presence_absence table.

Thanks, @seidelma. It's an ISO-formatted string coming from Samplify, so any that start with that date are covered by > string comparison.

schema/deploy/shipping/views.sql
```diff
@@ -1476,7 +1476,16 @@ create or replace view shipping.hcov19_presence_absence_result_v1 as
         sample_id,
         presence_absence_id,
         present,
-        pa.modified::date as result_ts,
```
kschwabefry, line 1479: I think this is a good move overall, but the one thing that sticks out to me is that this changes that column from a date to a time stamp. I worry this could have downstream impacts where code expects a date and doesn't explicitly cast it. One example, EHS view uses this as  test_result_date which it will try to put into a redcap column thats defined as a date. I also don't know if we want that level of specificity for things for return of results?

This could be resolved by casting the result_timestamp to a date instead.

kschwabefry (changes requested):
See comment below on date vs date time. Lets talk some more before this gets merged.

schema/deploy/shipping/views.sql
davereinhart: > I think this is a good move overall, but the one thing that sticks out to me is that this changes that column from a date to a time stamp. I worry this could have downstream impacts where code expects a date and doesn't explicitly cast it. One example, EHS view uses this as test_result_date which it will try to put into a redcap column thats defined as a date. I also don't know if we want that level of specificity for things for return of results?
> 
> This could be resolved by casting the result_timestamp to a date instead.

Thanks for catching that! When I dropped coalesce in the last push, I accidentally removed the date casting. I'll get that fixed now and change the date cutoff to 2022-06-09.

kschwabefry (approved):

davereinhart:
Deployed via sqitch.

PR #321, opened 2022-06-06
Refactor Clinical Functions into New Shared Module

bencap:
As part of the PHSKC ETL work, several functions were imported from `redcap_det_uw_retrospectives.py` to `clinical.py` in order to expedite the work and avoid duplicating code. Other functions were copied and modified as needed. 

This work cleans up these imported/copied and modified functions, placing them in a new module `clinical_retrospectives.py`. Existing functionality was maintained in the new module and these generalized functions should be usable for future clinical ETLs including SCH and KP data as well as our current processes.


bencap:
[Refactoring clinical ETL functions](https://trello.com/c/1nZOdfnX/1360-refactoring-clinical-etl-functions)

bencap:
Still testing this but its up in case anyone has some initial feedback.

lib/seattleflu/id3c/cli/command/etl/clinical_retrospectives.py
```diff
@@ -275,6 +276,174 @@ def create_questionnaire_response(record: dict,
...
+        'ncvrt':         '871562009',
+        'result_value' : '871562009', # PHSKC UW Virology Test
+        'revfla':        '181000124108',
+        'fluapr':        '181000124108',
+        'revflb':        '441345003',
+        'flubpr':        '441345003',
+        'revrsv':        '441278007',
+        'revrhn':        '440925005',
+        'revadv':        '440930009',
+    }
```
davereinhart, line 361: Does changing these snomed codes require updating previously ingested data?

davereinhart (changes requested):
Still reviewing, but wanted to get some preliminary feedback to you. Also has some merge conflicts, so could use a rebase.

lib/seattleflu/id3c/cli/command/etl/clinical_retrospectives.py
bencap: Yeah this probably warrants some discussion. I could leave the associated `ncvrt` value unchanged, but when this was first implemented a UK version was used so I think this new code is more reflective of the correct snomed mapping. If we were to update it, future/updated ingestions would be the new value while previously ingested values would map to the old one. In that case it probably would make sense to update old data so that all the targets are the same. I think that the data could either be re-ingested or we could do some sort of direct SQL update where we change any records mapped to the old `target_id` (see below) to the new one.

```
select * from warehouse.presence_absence where target_id = 789
```

bencap: Ah nice find, I'll dig into if anything else needs updating.

bencap: Bumping revision number on UW Retros ETL to reingest records with the updated Snomed target, per discussion here: https://seattle-flu-study.slack.com/archives/C0173B4S1LY/p1655408195063209?thread_ts=1655402400.117989&cid=C0173B4S1LY

schema/deploy/warehouse/target/data.sql
```diff
@@ -125,7 +125,7 @@ with target_lineage (identifier, lineage) as (
 
         -- Snomed CT targets used by Ellume, Cepheid, UW Retros
         ('http://snomed.info/id/440930009',        'Adenovirus'),
-        ('http://snomed.info/id/1240581000000104', 'Human_coronavirus.2019'),
```
davereinhart, line 128: This sqitch change won't actually remove this entry from the table, it only adds new ones. That's actually preferable, because there are existing `presence_absence` records that reference that `target_id` (789), so deleting it would throw an error on the foreign key constraint. So I'd recommend not removing this line until all `presence_absence` records referencing `target_id` 789 have been updated to the new `target_id`, at which point it should be safe to remove this line and delete it manually from the table.

davereinhart: Also may be worthwhile to add a verify placeholder on warehouse/target/data (to prevent the sqitch message about it missing), similar to the one for warehouse/site/data: https://github.com/seattleflu/id3c-customizations/blob/master/schema/verify/warehouse/site/data.sql

bencap: Good point on the targets, I'll add that line back and put a Trello card or something up to remind me to do that after the retros are reprocessed. I'll add that verify script today.

bencap: We'll also need to clear out PA records with the old target once the new one goes in there. It will create new PA entries, so this should clear old records out that have a new record generated (replace second select with delete): 
```
with new_snomed as (
	select sample_id from warehouse.presence_absence where target_id = 952
) select * from warehouse.presence_absence join new_snomed USING (sample_id) where target_id = 789
```

davereinhart (approved):
Looks good to me! In addition to the follow-up deletion of presence_absence records with target_id 789, it would be worthwhile to do some QC on UW retros records in warehouse that are updated as a result of the revision number bump, as well as other clinical data that gets processed after this is deployed. Refreshing you local db first might be a good option, so you can compare any existing records that were updated.

bencap:
Good thought! I'll do a refresh before deploying.

bencap:
Deployed. Will monitor clinical documents updated due to revision number bump and other clinical data using these functions (such as phskc retros).

bencap:
Old PA records are cleaned up.

PR #322, opened 2022-06-07
Add AIRS swab-and-send site to warehouse.site

seidelma:
Database changes for the AIRS ETL. This doesn't need to go at the same time as the AIRS ETL at large, so I put it in a separate PR. Tested it on my local dev database using a snapshot from a month or so ago with no errors.


schema/deploy/warehouse/site/data.sql
```diff
@@ -5,6 +5,8 @@ begin;
 
 insert into warehouse.site(identifier, details)
     values
+        ('AIRSSwabNSend',                           '{"category": "community",  "type": "swab-n-send",
```
kschwabefry, line 8: Double checking here, is this the swab site being used by the lab for these samples? I spot checked an airs sample in the lims and it had sample_origin `airs` but I didnt see a swab site listed.

seidelma: I thought this was only used on the ID3C side--if it's coming from the lab without a swab site included then I can use Dave's suggestion and remove the swab site logic from the ETL entirely.

kschwabefry:
[Add AIRSSwabNSend site to warehouse.site table](https://trello.com/c/aEd5FD0U/1381-add-airsswabnsend-site-to-warehousesite-table)

seidelma:
I've rebased this from master, and fixed the `sample_origin` field.

kschwabefry (approved):

PR #323, opened 2022-06-10
Filter samples marked `never-tested` out of DOH linelist shipping view

davereinhart:
There are cases where records in REDCap are marked as not testable, but we receive
results from Samplify the referenced barcode. Adding an additional filter to ensure that
these samples (which have RoR result status of `never-tested`) are not included in DOH
linelist submissions.


davereinhart:
[Update RoR, EH&S Reporting, and WA DOH Linelist processes to utilize new timestamp values and automate linelist catch ups (3 of 3)](https://trello.com/c/fZBQKXd0/1380-update-ror-ehs-reporting-and-wa-doh-linelist-processes-to-utilize-new-timestamp-values-and-automate-linelist-catch-ups-3-of-3)

schema/deploy/shipping/views.sql
```diff
@@ -4206,6 +4207,7 @@ create or replace view shipping.linelist_data_for_wa_doh_v1 as (
       -- `shipping.return_results_v3` uses this same filter.
       where collected >= '2020-06-10 00:00:00 US/Pacific'
       and hcov19_presence_absence_result_v1.details @> '{"assay_type": "Clia"}'
+      and not sample.details @> '{"note": "never-tested"}'::jsonb
```
kschwabefry, line 4210: Looks good to me. With current prod data, I see this filtering out 1 result. Out of curiosity, why the explicit cast to jsonb?

schema/deploy/shipping/views.sql
```diff
@@ -4177,7 +4177,8 @@ create or replace view shipping.linelist_data_for_wa_doh_v1 as (
               when present = 'f' then 'negative'
               else 'inconclusive'
           end as test_result,
-          hcov19_result_release_date as date_tested,
+          result_ts as date_tested,
+          hcov19_result_release_date,
```
kschwabefry, line 4181: We can discuss over in https://github.com/seattleflu/backoffice/pull/312, but I wonder if we'd benefit from making a new _v2 of the shipping view.

schema/deploy/shipping/views.sql
davereinhart: I ended up deleting the P/A records for that one sample.

Casting to jsonb may not be necessary, but that's how it's shown in postgres docs: https://www.postgresql.org/docs/13/functions-json.html#FUNCTIONS-JSONB-OP-TABLE

davereinhart:
TODO prior to merging:

- [ ] Update cutoff date used in `date_tested` CASE statement to date of deployment.

kschwabefry (approved):

davereinhart:
Deployed.

PR #325, opened 2022-06-14
Add new encounter class mapping

seidelma:
Also make error messages more informative when we get an unknown
encounter class/status so it's easier to track down where they're
coming from.


kschwabefry:
Oh I just noticed the tests were failing. That would be good to look into before deploying. Thanks!

PR #326, opened 2022-06-16
Specify timezone when converting date to datetime

davereinhart:
When parsing clinical data, several date values are converted to datetime data type which
by default are interpretted as midnight UTC time. Indicating the timezone when these
conversions occur to ensure that date values are converted to midnight local time when a
time component is being added.


davereinhart:
[Fix encountered timestamps in clinical data parsing for SCH and KP ETLs](https://trello.com/c/OQncaUyR/1390-fix-encountered-timestamps-in-clinical-data-parsing-for-sch-and-kp-etls)

davereinhart:
Deployed.

PR #327, opened 2022-06-17
Add value to vaccine map

davereinhart:
A new value was received in UW Retros project with a slight variation from a
previously mapped value ("years" vs. "yrs"). Adding the new value to the map.


kschwabefry (approved):

davereinhart:
Deployed.

PR #330, opened 2022-06-23
Use timezone aware encountered value for age calc in PHSKC parsing

davereinhart:
The `encountered` and `birth_date` values were recently updated to be timezone-aware, so the age
calculation should use those two values rather than `birth_date` and `collect_ts` (which is timezone-naive).


kschwabefry (approved):

davereinhart:
Deployed.

PR #331, opened 2022-07-02
Add New Vaccine Mapping for UW Retrospectives

bencap:
This change adds a new mapping for the `covid-19 moderna mrna 12 yrs and older`
vaccine code so that the uw retrospectives det can process them successfully.


davereinhart (approved):

bencap:
Deployed.

PR #334, opened 2022-07-15
Use Local Time When Determining Result Date

bencap:
On 6/8/22 we started using a more accurate timestamp provided by Samplify to determine the result date of samples. However, this timestamp was reported in UTC which caused some samples sent in the evening to be resulted a day later than they actually were. This change uses local time when making this calculation so that late arriving samples are resulted on the correct day.


davereinhart (approved):

bencap:
Deployed.

PR #336, opened 2022-07-22
UW-retros ETL add sinovac vaccine and rework mapping

davereinhart:
Mapping a new value for Sinovac vaccine in the UW Retros ETL.

Also reworked the mapping to use two dictionaries to eliminate duplicate entries.


kschwabefry (approved):

davereinhart:
Deployed.

PR #337, opened 2022-07-29
Check Existence of Location Before Accessing its Identifier

bencap:
Previously the `encode_addresses` function attempted to directly access the location identifier, however this location has the potential of being a null value. We should first make sure the location exists before accessing its identifier.

Although I'm not sure if this fixes the root cause of the PHSKC parsing failure from 7/29, I think it is still an improvement to the code logic.


kschwabefry (approved):

bencap:
Deployed.

PR #339, opened 2022-08-10
Add novavax vaccine for uw retros

kschwabefry:
A new coding was added for a novavax vaccine, which
can be mapped to CVX 211.


kschwabefry:
Generated DETs and ran the ETL locally to confirm this works and was the only new value added in yesterdays data.

davereinhart (approved):

kschwabefry:
Deployed.

PR #340, opened 2022-08-19
Remove Old Snomed Code From Target Table Definition

bencap:
We are now using a new Snomed code for Covid, under target 952. This change
removes the outdated target from the definition of the table. PA results
with this old target (789) were transitioned to the updated target (952).
When this sqitch change is deployed, we can also remove the outdated target
from the warehouse table and continue using 952 for future snomed cov targets.


bencap:
Finishes QC cleanup needed from #321

kschwabefry:
[Update hCoV-19 snomed code in warehouse.target](https://trello.com/c/cEcDzkCr/443-update-hcov-19-snomed-code-in-warehousetarget)

kschwabefry (approved):

bencap:
Deployed via Sqitch.

bencap:
Old target removed from table.

PR #341, opened 2022-09-21
Handle New Vaccine Date Format in UW Retros

bencap:
We recently filled in some missing date in a handful of UW Retros records in REDCap, but some date fields were populated with a non ISO date format (MM/DD/YYYY) which causes issues with downstream fhir processes. This change adds a conditional that reformats this date string into ISO format for compatibility with downstream processes.


kschwabefry (approved):

bencap:
Deployed.

bencap:
Uploaded DETs to reprocess these records.

bencap:
[Address negative ages in UW retros](https://trello.com/c/mxX0OwpG/1429-address-negative-ages-in-uw-retros)

PR #342, opened 2022-09-27
Add New HCT Dropbox Sites

bencap:
This change:
- Adds 6 new HCT Dropbox locations to the site table
- Makes the relevant changes in the UW Reopening ETL to pull detailed dropbox locations into site references for records.

One thing that requires some discussion. If you look at the `dropbox_location` variable under the daily attestation instrument in the [REDCap HCT Codebook](https://hct.redcap.rit.uw.edu/redcap_v12.3.3/Design/data_dictionary_codebook.php?pid=148), you'll see that we have possible values for all of our dropbox locations but also have a `courier` value. I just map that value to the swab and send location, but maybe it is worth it to also check and make sure the collection method is what we expect? I don't have that logic in at the moment, but perhaps we would want to check that our collection method is `swab_and_send` and skip the record if they don't match? If whoever reviews this has some thoughts let me know.


bencap:
[Add HCT dropbox sites](https://trello.com/c/p47uMMaa/1435-add-hct-dropbox-sites)

schema/deploy/warehouse/site/data.sql
```diff
@@ -178,11 +178,23 @@ insert into warehouse.site(identifier, details)
         ('UWReopeningSwabNSend',                    '{"category": "community",  "type": "uw-reopening"}'),
         ('UWReopeningDropbox',                      '{"category": "community",  "type": "uw-reopening",
             "swab_site": "uw_box"}'),
+        ('UWReopeningDropboxTacoma',                '{"category": "community",  "type": "uw-reopening",
```
kschwabefry, line 181: I wonder whether these should have `type: collegeCampus` or `type: uw-reopening` - its a bit of a can of worms, maybe best discussed at standup tomorrow.

bencap: Thanks for the brief discussion at standup this morning about this. You think we should talk more about it along with these changes? I can add a Trello card for it if not so we don't forget to revisit.

kschwabefry: I think let's go ahead with the database portion of these changes (as-is is fine for now!!) and just drop the etl changes as you did already!

kschwabefry (approved):

PR #343, opened 2022-10-05
Process SCH covid dose 4 data

kschwabefry:
There are new variables in the data for Seattle
Childrens retros for covid vaccine dose 4
(covid_vx_d4, cov_d4_date and covid_vx_manu4)

These still have the Year4 prefix, which means we have two formats (with and without these new variables) under the same year4 format. Additional tweaks are made to accomodate year4 files both with and without these new variables.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,96 +18,112 @@
     "default": {
         "aiobotocore": {
             "hashes": [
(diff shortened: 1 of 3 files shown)


kschwabefry:
Tested out with `id3c clinical parse-sch` and `id3c clinical upload`, and after a fresh db dump will run `id3c etl clinical`

davereinhart (approved):
LGTM! Thanks for the walkthrough on the identifier updates too.

kschwabefry:
Deployed.

PR #344, opened 2022-10-07
Add current HCT project id to shipping views

davereinhart:
HCT shipping views are filtered on REDCap project id 45 (2021-2022). Adding project 148 (2022-2023) to include encounters from both projects.  Enrollments have been migrated from project 45 to 148, so we're using only project 148 enrollment data going forward.


davereinhart:
[Update HCT shipping views to include new project](https://trello.com/c/UByL2G0J/1438-update-hct-shipping-views-to-include-new-project)

bencap (approved):
Looks good. Out of curiosity why were all the enrollment identifiers updated but not the encounter identifiers? Is it because we continue to use the enrollment encounters even if they occurred in the prior project but non-enrollment encounters are finished?

davereinhart:
> Looks good. Out of curiosity why were all the enrollment identifiers updated but not the encounter identifiers? Is it because we continue to use the enrollment encounters even if they occurred in the prior project but non-enrollment encounters are finished?

Good question! The enrollments were migrated to the Y3 project, so we updated those identifiers to reference the current records. The Y2 encounters were not migrated, so those still reference the REDCap project where they exist. All encounters (Y1, Y2, Y3) should still be linked to the appropriate `individual_id`, it's just the enrollment records that have been migrated each year. (Note: between Y1 and Y2, some enrollments were intentionally excluded from migration)

davereinhart:
Deployed via sqitch.

PR #345, opened 2022-10-24
Add Shipping View for HCT Dashboard Data Pulls

bencap:
This change adds a shipping view for HCT dashboard data pulls, essentially a simplified version of the hct encounters v1 view but with less data and simplified joins. This shortens query time and provides only the necessary data for quicker lambda execution.

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -3694,6 +3694,112 @@ grant select
   on shipping.uw_reopening_encounters_v1
   to "incidence-modeler";
 
+
+create or replace view shipping.uw_reopening_results_hct_data_pulls as
+(
+  with encounters as
+      (
+      select encounter_id
+      , encountered
+      , age
+      , individual_id
+      from warehouse.encounter
+      where
+          encounter.identifier like 'https://hct.redcap.rit.uw.edu/45/%/encounter_arm_1/%' or
+          encounter.identifier like 'https://hct.redcap.rit.uw.edu/148/%/encounter_arm_1/%'
+      )
+
+  select
+      encounters.individual_id as individual_id,
+      sample.identifier as sfs_sample_identifier,
+      target.identifier as target_identifier,
+      presence_absence.present
+  from
+      warehouse.sample join
+      warehouse.presence_absence using (sample_id) join
+      warehouse.target using (target_id) join
+      encounters using (encounter_id)
+  where
+      encounter_id in (
+          select encounter_id from encounters
+      )
+)
+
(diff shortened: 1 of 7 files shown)


bencap:
I'll also throw up an entry in the data dictionary for this view.

schema/deploy/shipping/views.sql
```diff
@@ -3694,6 +3694,69 @@ grant select
...
+  , enroll_details.campus_location
+  , enroll_details.core_house_members
+
+  from encounters encounter
+  join enrollments enroll on encounter.individual_id = enroll.enrollment_individual_id
+  join warehouse.individual individual on individual.individual_id = encounter.individual_id
+  join shipping.uw_reopening_enrollment_fhir_encounter_details_v1 enroll_details on enroll_details.encounter_id = enroll.enrollment_encounter_id
+
+  -- Filter out follow up encounters
+  where not encounter.details @> '{"reason": [{"system": "http://snomed.info/sct", "code": "390906007"}]}'
```
kschwabefry, line 3743: Do you mind explaining a bit what youre intending here? In this view, are you trying to have one row per individual or per any encounter or certain types of encounters? Most of the variables seem to be coming from enrollments but since youre merging to encounters youre ending up with multiple encounter records per individual in the view.

schema/deploy/shipping/views.sql
```diff
@@ -3694,6 +3694,69 @@ grant select
...
+  (
+  select encounter_id
+  , encountered
+  , age
+  , individual_id
+  , site_id
+  , details
+  from warehouse.encounter
+  where
+      encounter.identifier like 'https://hct.redcap.rit.uw.edu/148/%/encounter_arm_1/%'
```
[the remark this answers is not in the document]
bencap: Ah right. Only enrollments were migrated right? So encounters would still need to get pulled in from past years?

schema/deploy/shipping/views.sql
bencap: I see your point now. I guess I want one row per individual for the enrollment data so I don't duplicate any, but then I do still want one row per encounter so that I can pull in all encounter data for aggregation.

davereinhart (changes requested):
The two new shipping views should be added to `schema/verify/shipping/views.sql` to verify they exist.

davereinhart (approved):
Looks good!  Thanks for additions to README file too.

bencap:
Database changes deployed via Sqitch.

bencap:
API changes deployed.

PR #346, opened 2022-10-26
AIRS ETL updates 2

davereinhart:
Minor changes based on QC of records after #338 was deployed.  Updated logic to check for completed PCDEQC _or_ swab results to determine if sample was received. Changed order of prioritized barcodes to use value from BEMS instrument (which is always required).


davereinhart:
Deployed.

PR #350, opened 2023-01-20
Add new vaccine value to map for UW retros ETL

davereinhart:
An unmapped vaccine value was received for the Pfizer bivalent booster, adding here as CVX code 300 based on codes available from CDC (https://www.cdc.gov/vaccines/programs/iis/COVID-19-related-codes.html).


bencap (approved):

davereinhart:
Deployed.

PR #352, opened 2023-01-23
Add another new value to race map

davereinhart:
Adding newly encountered value 'alaska native' from UW retrospective data to Redcap DET ETL race mapping function.


davereinhart:
Deployed.

PR #353, opened 2023-01-23
Force PHSKC Barcode and CID Comparisons to be Uppercase

bencap:
Original implementation of the PHSKC ETL assumed that barcodes and CID values would always be provided as uppercase. This turned out to not be the case, and this caused many samples from October 2022 to not be linked to the associated metadata. This change forces all columns which will be compared for metadata associations to be uppercase.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/clinical.py
@@ -612,6 +612,14 @@ def add_phskc_manifest_data(df: pd.DataFrame, manifest_filename: str) -> pd.Data
         LOG.warning(f'Dropping {duplicated_cids.sum()} rows with duplicated CID(s) from PHSKC manifest data')
         manifest_data = manifest_data[~duplicated_cids]
 
+    # ensure all of our comparison columns are uppercase so barcodes can be compared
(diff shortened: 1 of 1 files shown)


bencap:
[Investigate high proportion of October PHSKC samples with missing metadata](https://trello.com/c/HkzXR0Hm/1451-investigate-high-proportion-of-october-phskc-samples-with-missing-metadata)

davereinhart (commented):
Might also be worth removing whitespace from these columns before merging.

Also Is there any particular reason to use upper instead of lower here?  If values are only used for joining in pandas, it won't matter, but if these end up being stored in ID3C in this format, it might be worth using lowercase for consistency.

bencap:
@davereinhart Good point on the whitespace, I'll do that as well.

No particular reason for upper other than that these values usually come as uppercase values. None are pulled into ID3C though. The fact that the column is named `barcode` is also a bit confusing. It comes to us from PHSKC with that name but it is some internal identifier on their end for the samples and isn't the same as our barcodes.

davereinhart (approved):

bencap:
Deployed.

PR #354, opened 2023-01-26
Map new vaccine values to CVX codes

davereinhart:
Mapping two newly encountered vaccine values to the appropriate CVX codes in UW retros ETL.


bencap (approved):

bencap:
Deployed.

PR #355, opened 2023-02-06
Updates to SCH ETL to handle changes for Y5 and end of Y4

davereinhart:
Updates to handle `year4` and `year5` manifest formats for SCH. This includes new columns that were added at the end of Year 4 for a fifth covid vaccine dose status, date and  manufacturer.


davereinhart:
This is a quick fix to get the ETL back on track, but this could benefit from some changes in how the file formats are designated and handled. The Year 1-5 format versioning approach does not work well since we have formatting changes occur within the same year.  

A new file naming convention with format versioning independent of study year, or some logic that looks at the list of columns to determine the format would be a better long-term solution...

bencap (approved):
Looks good Dave, thanks for taking care of this. Agreed that this could use a bit of cleanup at some point.

PR #356, opened 2023-02-16
Treat Empty Strings as None in Clinical ETL

bencap:
Some PHSKC records were coming through with non breaking spaces as the data for certain clinical fields. This caused issues with the mapper, which assumed that any strings should have some valid data in them. This change removes that assumption, instead treating strings that have been stripped empty as `None` type mappings.


davereinhart (approved):

bencap:
Deployed.

PR #359, opened 2023-03-31
PHSKC ETL Bugfixes

bencap:
There was a bug affecting ingestion of 63 records for the PHSKC rework that dropped records if they had a null value for one of their identifiers. A single null identifier isn't an issue, and if all three of a records' identifiers are null it will still be dropped.

An additional bug casted the `census_tract` field to an integer instead of a string, which had unintended downstream ramifications during the FHIR ETL.


davereinhart (approved):
LGTM!

PR #360, opened 2023-06-27
Add `Native Hawaiian` Race to Clinical ETL

bencap:
Adds a new race mapping, `native hawaiian` to our clinical ETL. This race mapping was present before without a space, but a new value parsed out by our ETL includes a space.


bencap:
Deployed.

PR #361, opened 2023-07-12
Upgrade click to 8.x and temporarily pin to <8.1.4

sallybg:
Upgrade click for compatibility with M2-compatible Fiona, and temporarily pin click to <8.1.4 to avoid mypy type check failure (pallets/click#2558). This PR should be tested alongside https://github.com/seattleflu/id3c/pull/325.


lib/seattleflu/id3c/cli/command/__init__.py
```diff
@@ -80,6 +80,12 @@ def duplicated_barcode(df: pd.DataFrame) -> pd.DataFrame:
     and a description of the problem.
     """
     duplicates = pd.DataFrame(df.barcode.value_counts())
+    
+    # pandas 2.0.2 acts differently than earlier versions here, either during value_counts() or DataFrame()
+    # v2.0.2 names the column in duplicates 'count' instead of 'barcode'
+    if 'barcode' not in duplicates.columns:
+        duplicates = duplicates.rename(columns={"count":"barcode"})
+
```
davereinhart, line 88: This was actually new in Pandas 2.0: https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#value-counts-sets-the-resulting-name-to-count

So another way to deal with this would be to pin pandas to >=2.0, and then we can assume that the column will be always be "count", but we need to be careful with major version upgrades like this.  Our prod environment is currently pinned to version 1.5.3. (https://github.com/seattleflu/backoffice/commit/9ec869671f7a4dadc310de775927ba2e5ccd57d3)

davereinhart (approved):

PR #365, opened 2023-07-20
Map newly encountered vaccine description to CVX code

sallybg:


bencap (approved):

davereinhart:
@sallybg Before this gets deployed, we may want to think about bumping the ETL version to correct all the previously ingested "gray cap" values by re-processing them. That will reprocess all records though, so another option would to manually identify and update those records.

sallybg:
Keeping the version number as is, and will manually generate and upload DETs for all records with vaccine code 208 in order to avoid reprocessing of all records.

PR #368, opened 2023-07-31
Add security invoker to shipping views to enable row-level security

davereinhart:
Requires ID3C PR: https://github.com/seattleflu/id3c/pull/328

Adds `security_invoker = true` to shipping views to implement row-level security added in the above PR.  Adds `cascadia` and `reporter-cascadia` roles, and related constraints to enforce row-level security on appropriate tables.


bencap (commented):
Looks good. I think the tests are failing because of the lack of a tag but you may be waiting for the ID3C PR?

And then similar to the PR in ID3C, I think we need to update our `sqitch.conf` file here as well to be able to deploy.

bencap (approved):
Tested locally and roles + access control + rls checks looked good

davereinhart:
Deployed via sqitch.

PR #369, opened 2023-08-02
Define condition and observation FHIR resources as TypedDicts in ETLs

sallybg:


jstone-uw:
The new classes look good to me.

I only have one really minor comment: I think `typing.List` has been deprecated, along with `typing.Dict` and others, in favor of built-in types like `list` and `dict`, which are generic types as of Python 3.9. I noticed one place in [redcap_det_uw_retrospectives.py](https://github.com/seattleflu/id3c-customizations/pull/369#diff-4469ea3d10dd91bdd205e7459c0c5148d372567f93bfe406170525f6737b08b7) where `list` was replaced with `List[dict]`. Making this generic with the `dict` subscript is important, but maybe we should use `list[dict]` and start replacing other deprecated types with their built-in counterparts.

PR #370, opened 2023-08-16
Add KP2023 ETL routine

sallybg:
Add clinical commands parse-kp2023, match-kp2023.
Update etl clinical command to process output from clinical commands with the FHIR standard.
Fix drop_missing_rows in clinical commands, which was not returning result as intended.
Note that other steps in the ETL routine such as downloading metadata from s3, deciding which csv(s) to parse, and selecting only new or changed data for upload to id3c, will be handled in backoffice.


jstone-uw (approved):
The changes look good to me!

PR #371, opened 2023-09-08
Update sample table constraint to handle null values

davereinhart:
The constraint on warehouse.sample table to ensure cascadia row-level security is being used on relevant records was not handling null values correctly. Adding coalesce to ensure constraint logic evaluates to false, rather than null, so that constraint is applied correctly when `access_role` is null.


bencap (approved):

davereinhart:
Deployed via sqitch.

PR #372, opened 2023-10-31
KP2023 ETL input validation

sallybg:
Validate input metadata spreadsheet, strip aliquot number from collection id if present, and assign ambiguous flu vaccine names as unspecified when creating FHIR bundle.

### DIFF
--- a/lib/seattleflu/id3c/cli/command/clinical.py
+++ b/lib/seattleflu/id3c/cli/command/clinical.py
@@ -875,6 +875,10 @@ def parse_kp2023(kp2023_filename: str) -> None:
     clinical_records.columns = clinical_records.columns.str.lower()
     clinical_records = trim_whitespace(clinical_records)
     clinical_records = add_provenance(clinical_records, os.path.basename(kp2023_filename)) 
+    # since kp may provide updates as new files, and to avoid processing outdated records in id3c, add spreadsheet timestamp column
+    last_modified_time = os.path.getmtime(kp2023_filename)
+    clinical_records['spreadsheet_timestamp'] = last_modified_time
+    clinical_records['spreadsheet_timestamp'] = pd.to_datetime(clinical_records['spreadsheet_timestamp'], unit='s')
 
     # check that all expected columns are present (even if empty)
     # do this check before standardizing names, but allow fixes for known typos through
@@ -1043,7 +1047,8 @@ def parse_kp2023(kp2023_filename: str) -> None:
         'date_covid_5',
         'date_covid_6',
         'date_symptom_onset',
-        'patient_class'
+        'patient_class',
+        'spreadsheet_timestamp'
     ]
 
     # throw a warning if there are any columns not in columns_to_keep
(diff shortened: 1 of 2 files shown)


lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -874,35 +874,84 @@ def parse_kp2023(kp2023_filename: str) -> None:
...
+    # therefore, we will check that there is no aliquot number at the end of the collection id in the metadata spreadsheet,
+    # and strip the aliquot number if it is there, with a warning since we don't expect it
+    # in the LIMS, this id is called KaiserPermanenteSpecimenId and will also have its aliquot number stripped
+    collection_ids_with_aliquot = clinical_records.loc[clinical_records['collection_id'].apply(lambda x: True if re.search(r'-\d+$', x) else False)]['collection_id'].values
+    # if there are any collection ids with aliquot (not expected), give a warning and strip the aliquot number
+    if len(collection_ids_with_aliquot) > 0:
+        LOG.warning(f'Warning: One or more Marshfield lab IDs contain aliquot id, which is unexpected. Stripping aliquot id from these samples: {*collection_ids_with_aliquot,}')
+        for cid in collection_ids_with_aliquot:
+            # strip aliquot number ('-<number>') from collection id
+            clinical_records.loc[clinical_records['collection_id'] == cid, 'collection_id'] = re.sub(r'-\d+$','',cid)
```
bencap, line 954: You may find this less readable, so feel free to ignore, but when I am working with Pandas dataframes I will generally try to avoid for loops. You can do a lot of the same things with internal Pandas functions and indexing, which have some pretty significant performance benefits since it is all C compiled code. Obviously here those performance gains don't matter so much, but for transparency here is a way you can keep this logic within Pandas.

 ```suggestion
    collection_ids_with_aliquot = clinical_records['collection_id'].apply(lambda x: True if re.search(r'-\d+$', x) else False)
    # if there are any collection ids with aliquot (not expected), give a warning and strip the aliquot number
    if any(collection_ids_with_aliquot):
        LOG.warning(
            'Warning: One or more Marshfield lab IDs contain aliquot id, which is unexpected. Stripping aliquot id from' +
            f'these samples: {*clinical_records.loc[collection_ids_with_aliquot]["collection_id"].values,}'
        )
        clinical_records.loc[collection_ids_with_aliquot, 'collection_id'] = clinical_records.loc[
            collection_ids_with_aliquot, 'collection_id'
        ].apply(lambda cid: re.sub(r'-\d+$','', cid))
```

File: lib/seattleflu/id3c/cli/command/clinical.py [hunk 2]
```diff
@@ -996,6 +1045,14 @@ def parse_kp2023(kp2023_filename: str) -> None:
         'patient_class'
     ]
 
+    # throw a warning if there are any columns not in columns_to_keep
+    # except for known extra columns, like 'individual' which might not be included anymore but has been included in the past
+    extra_cols = list(set(clinical_records.columns).difference(columns_to_keep))
+    if 'individual' in extra_cols:
+        extra_cols.remove('individual')
+    LOG.warning(f'One or more unexpected columns present after parsing: {*extra_cols,}.\n\
+        Removing these columns before dumping records to stdout.')
```
bencap, line 1054: I think here you meant to only throw this warning if there were extra cols. At the moment, it will log even if the list is empty.
```suggestion
    if len(extra_cols) > 0:
        LOG.warning(f'One or more unexpected columns present after parsing: {*extra_cols,}.\n\
        Removing these columns before dumping records to stdout.')
```

bencap (commented):
Looks good overall, left a couple minor suggestions. We should make sure the warning is only logged when necessary before merging, but the other comment isn't critical and the way you implemented it is totally fine as well.

bencap (approved):
Thanks for the updates!

bencap (approved):
Looks great, thanks Sally.

PR #378, opened 2024-03-29
Update vaccine mapper in uw retros redcap det etl

sallybg:
Note that the newly added vaccines are actually older covid-19 vaccines. I spot checked in REDCap that the vaccine dates associated with these vaccines line up with the 'active' date ranges for these old vaccines (see https://www2a.cdc.gov/vaccines/iis/iisstandards/vaccines.asp?rpt=cvx).


bencap (approved):
Thanks for adding these @sallybg, looks good.

PR #379, opened 2024-04-09
Revive Kiosk ETL

bencap:
DRAFT


davereinhart:
I think the kiosk ETL will need changes similar to this commit to pass the mypy tests:
https://github.com/seattleflu/id3c-customizations/commit/66daa4fcad075d72002b76e2b7d55669fdd28c82

PR #383, opened 2024-04-19
KP2023 data dict update and ingest ICD10 codes

sallybg:
Update clinical parse-kp2023 to reflect data dictionary updates for symptom columns.
Update clinical ETL to add encounter reference to ICD-10 condition entries for KP2023 encounters. Previously, the encounter reference was missing from these entries, which caused them to be skipped by the FHIR ETL.
Update parse-kp to ingest comorbidity metadata for encounters that were part of the original KP sample stream (2018-2020).


lib/seattleflu/id3c/cli/command/clinical.py
```diff
@@ -1071,117 +1084,168 @@ def parse_kp2023(kp2023_filename: str) -> None:
...
+    a prefix denoting columns of interest, returns True if all columns
+    beginning with the provided prefix contain only 0/1/None values,
+    otherwise returns False.
+    """
+    cols = [c for c in df.columns if c.startswith(prefix)]
+    for c in cols:
+        if not np.isin(df[c].dropna().unique(), [0, 1]).all():
+            return False
+    
+    return True
```
bencap, line 1114: Could shorten this block to something like:
```suggestion
    return all([np.isin(df[c].dropna().unique(), [0, 1]).all() for c in cols])
```
but not a big deal (and is probably less readable).

bencap (approved):
Looks good, thanks for the changes.

PR #386, opened 2024-06-27
Kp2023 race and site updates

sallybg:
- Add "no_response" to etl race map
- Add "kp_2" to site map, mapping to KaiserPermanente

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,124 +18,139 @@
     "default": {
         "attrs": {
             "hashes": [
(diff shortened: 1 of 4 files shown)


sallybg:
It looks like we need to pin numpy at < 2.0 to avoid this error:
"ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject"
see https://stackoverflow.com/questions/78634235/numpy-dtype-size-changed-may-indicate-binary-incompatibility-expected-96-from

davereinhart (approved):

PR #388, opened 2024-09-05
Add Four new Covid-19 Vaccine Codes

bencap:
Vaccine source link has also moved here: https://www.cdc.gov/iis/code-sets/covid-19-related-codes.html. Seems like historical vax codes are embedded within an excel sheet now as well, although it seems reasonable we won't have any new vaccines coming in that map to retired vaccines.


davereinhart (approved):

bencap:
Deployed

